PR #2, opened 2020-01-29
Minor improvements to HTML/CSS generation

windymilla:
Edits to the default header to fix non-centred HRs and omitted borders in ADE ebook viewer
Edits to code generating anchors for h2 headings to avoid chapters appearing as hyperlinks in ADE
Also intended as a test of the Git process (new to me:)

### DIFF
--- a/lib/Guiguts/HTMLConvert.pm
+++ b/lib/Guiguts/HTMLConvert.pm
@@ -1031,7 +1031,7 @@ sub html_convert_body {
 			# insert chapter heading unless already a para or heading open
 			if ( not $selection =~ /<[ph]/ ) {
 				$textwindow->ntinsert( "$step.0",
-					"<h2><a id=\"" . $aname . "\" />" );
+					"<h2 id=\"" . $aname . "\">" );
 				my $linesinheader=1;
 				while (1) {
 					$step++;
--- a/lib/Guiguts/HTMLConvert.pm
+++ b/lib/Guiguts/HTMLConvert.pm
@@ -589,7 +589,7 @@ sub html_convert_body {
 #   print "int:$intitle:las:$last5[3]:sel:$selection\n";
 			if (
 				(
-					length($selection) 
+					length($selection)
 					&& (   ( !$last5[3] )
 						or ( $last5[3] =~ /<\/?h\d?|<br.*?>|<\/p>|<\/div>/ ) )
 					&& ( $selection !~ /<\/?h\d?|<br.*?>|<\/p>|<\/div>/ )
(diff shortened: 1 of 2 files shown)


lib/Guiguts/HTMLConvert.pm
```diff
@@ -1045,7 +1045,7 @@ sub html_convert_body {
 					} else {
 						# end of header reached
 						$step--;
-						$textwindow->ntinsert( "$step.end", '</a></h2>' );
```
cpeel: I would think that removing the `</a>` would generate invalid HTML - is that not the case?

windymilla: A few lines further back (1034), where the h2 element is started, I have changed what was an opening `<a name=...>` tag which I have made self-closing by adding the / at the end of the tag. With the old code, the anchor element surrounded the chapter title, which was fine in browsers, but in the ADE epub reader, it interpreted the `<a>...</a>` as marking up a hyperlink, rather than the target of a link, so displayed it blue and underlined.  With the new code, the self-closing tag doesn't enclose any text, so the problem is avoided. 

However, there is an alternative edit which would be to remove the `<a>` element completely, and just put the id attribute in the h2 tag. I was hesitant to do that when making the change initially to avoid altering Guiguts output too much, but looking online a bit more it looks as though that would be a more modern way (at least this century!) than using an anchor element as Guiguts has always done. If you agree and I wanted to do that, am I able to "uncommit" (I'm sure there's a proper word for that) my change, adjust it to what we agree, then re-commit? Or is there a better way?

windymilla: That line 1034 would then be `"<h2 id=\"" . $aname . "\">" );`

cpeel: Ah! I missed the addition of `/` to the `<a>` tag. Lets go with your addition of the `id` attribute to the `<h2>` tag instead as that would be the better way to do it.

There are 3 different ways you can do this in git:
1. Just add another commit with your change -- this is what I would recommend in this case.
2. Add your commit but before you push it up, use `git rebase -i` to fixup your two commits that change the `<a>` tag together so they are one. Then you will need to `git push --force` your changes up since you've rewritten history for the branch. This is a common git development approach but `git rebase` can be a bit to wrap ones head around.
3. I can squash your MR into one  big commit on merging.

windymilla: Thanks Casey - I've committed the change we discussed. Thanks also for the Git training!

PR #3, opened 2020-02-22
Improvements to handling of chapter headings in HTML version

windymilla:
Ebookmaker uses div with a class of "chapter" to flag the start of a new chunk of the epub file, and default PP practice is to use this to surround each chapter heading. This reduces the chance of a chunk becoming too full at an inconvenient moment in the epub generation, which would cause an unwanted pagebreak at a random point in the book. These edits surround each h2 element with a chapter class div, also including the preceding page number if appropriate.


windymilla:
Bug reported by tester - not ready for including yet.

cpeel:
In the future, you're also welcome to leave the MR open and just add a comment that it needs further testing and another one when you're ready for it to be reviewed. Closing it and opening another works as well, just listing some options.

windymilla:
Thanks Casey - don't hesitate to tell me if I do something annoying, or even if it's just something I can do to make your life easier. I would like to be as small a nuisance as possible :)

PR #4, opened 2020-02-25
Surround h2 headings with chapter divs

windymilla:
A div with class "chapter" causes ebookmaker to create a new chunk in the epub file. Using them around every h2 is best practice, and helps to avoid mid-chapter pagebreaks occurring due to epub chunk size overflow.

### DIFF
--- a/headerdefault.txt
+++ b/headerdefault.txt
@@ -44,6 +44,9 @@ hr.full {width: 95%; margin-left: 2.5%; margin-right: 2.5%;}
 
 hr.r5  {width: 5%; margin-top: 1em; margin-bottom: 1em; margin-left: 47.5%; margin-right: 47.5%;}
 hr.r65 {width: 65%; margin-top: 3em; margin-bottom: 3em; margin-left: 17.5%; margin-right: 17.5%;}
+ 
+div.chapter {page-break-before: always;}
+h2.nobreak  {page-break-before: avoid;}
 
 ul.index { list-style-type: none; }
 li.ifrst { margin-top: 1em; }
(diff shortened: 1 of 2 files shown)


cpeel:
Consider removing the commit that you reverted and the reversion so that there is just 
4f286e1 and 8c4cb38. To do that:
```
git rebase -i 4f286e1^
```
In the editor that comes up, delete entirely the lines for the two commits we want to remove (407f478 & 8c4cb38) then save and exit.

This will leave you with just the two commits you want. Because you've rewritten history, you'll need to use `--force` to push your branch back up:
```
git push --force origin windymilla:master
```

This is a bit of advanced git usage that comes in handy. Not required for this MR, but I thought you might be interested in a bit more git knowledge and how we would ideally handle this.

windymilla:
Commit/revert pair removed - should be ready for merging now - thanks

cpeel:
I'll be frank - the code looks good to me but I'm not familiar enough with guiguts to validate it functionally. I'm going to act mostly as a pass-through on these MRs rather than an in-depth reviewer and assume that you've done testing. I'm very comfortable with this approach -- some traction on this project is better than none -- but I wanted to be clear the value that I'm adding (or not adding) to the review process.

windymilla:
Thanks - that seems like a reasonable approach. Where it's non-trivial, I
will get others to test the functionality (as I did this time - hence the
bug being spotted). It's a shame Guiguts isn't well documented, as it means
it takes a long time to work out what's supposed to be going on, and anyone
trying to modify it is more likely to have inadvertent consequences.

srjfoo:
> It's a shame Guiguts isn't well documented, as it means it takes a long time to work out what's supposed to be going on, and anyone trying to modify it is more likely to have inadvertent consequences.

It looks like you're trying to add some comments to document, though, so that's good. Thanks for working on this, Nigel. It's very much appreciated.

PR #5, opened 2020-02-28
Improve HTML code for pagenums & footnotes

windymilla:
Minor CSS improvement to avoid page numbers being affected if a page break occurs in the middle of  bold, italic or smallcap markup.
Also simplifed code output for pagenums and footnotes, removing the deprecated "name" attribute and combining <a> elements where possible.

### DIFF
--- a/headerdefault.txt
+++ b/headerdefault.txt
@@ -70,6 +70,9 @@ table {
     left: 92%;
     font-size: smaller;
     text-align: right;
+    font-style:normal;
+    font-weight: normal;
+    font-variant: normal;
 } /* page numbers */
 
 .linenum {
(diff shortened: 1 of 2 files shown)


windymilla:
Developed in a branch as suggested - thanks

headerdefault.txt
```diff
@@ -70,6 +70,9 @@ table {
     left: 92%;
     font-size: smaller;
     text-align: right;
+    font-style:normal;
```
cpeel: nit: missing whitespace (and if you don't care that's fine with me but I wanted to point it out)

windymilla:
Oops. For something like that, do I fix it or will you as you merge?

cpeel:
> Oops. For something like that, do I fix it or will you as you merge?

If we want to fix it, you should fix it before I merge it.

windymilla:
Thanks

PR #6, opened 2020-02-29
Regex & Greek changes by Tony Browne

windymilla:
Files from Tony's NEW1028.zip file. "These fix mostly errors relating to Greek processing plus correcting the treatment of some active regular expression strings when referenced as ordinary text plus handling of y-uml." and "There is a recently discovered bug/function deficit in Guiguts concerning the handling of the Latin-1 characters i-uml (ï) & u-uml (ü) when used in association with Greek processing and iota-dieresis (ϊ) & upsilon-dieresis (ϋ)."


windymilla:
The only file of Tony's which I needed to edit was Utilities.pm due to Tony's version pre-dating the move to GitHub & consequent URL changes in the code.

windymilla:
Tony has tested this work carefully, and his patch has been used by others without any issues. I have run Guiguts with these files included, but not tested the effects of the patch on Greek or regexp handling.

windymilla:
Forgot to mention, Tony's files have different spacing/wrapping - looks like some sort of beautifying.

cpeel:
Ouch. This won't work as-is. Tony's files use spaces instead of tabs -- in addition to some other (minor?) formatting changes -- which is why it looks like everything is different. We can have a philosophical discussion if the files should be using spaces or tabs, but they need to be consistent and if we are changing them we need to change all of them at once in a single commit with that being the only change.

I suspect if you run the regex `s/        /\t/g` against all of his files before committing them there would be *a lot* fewer differences.

rtonsing:
I'm not fluent with GuiGuts code, but it has been my experience as a SW engineer that it is always recommended to use all spaces.

For example, according to the Python PEP8 standard:
"Spaces are the preferred indentation method.
Tabs should be used solely to remain consistent with code that is already indented with tabs.
Python 3 disallows mixing the use of tabs and spaces for indentation.
Python 2 code indented with a mixture of tabs and spaces should be converted to using spaces exclusively."

My 2 cents worth.
I can help with conversion if needed.

windymilla:
Thanks Rick. I fully agree with Casey though that re-formatting shouldn't happen at the same time as functional changes as it makes it very hard to spot the significant edits. It's not just tabs/spaces, the positioning of curly brackets has also been changed. I'm either going to remove this commit and try again after un-reformatting the files in Tony's zip, or it maybe easier to just start a new branch. Unfortunately un-reformatting is a bit more complicated than putting in a tab for every 8 spaces, because there are places in the existing code where spaces are currently used for alignment, etc., normally assuming 4 spaces per tab.

rtonsing:
Robert, not Rick. Guess I should use a better user name.

I agree with you, each kind of change should be separate.

cpeel:
> I'm either going to remove this commit and try again after un-reformatting the files in Tony's zip, or it maybe easier to just start a new branch.

If you mean having a new branch to maintain to support just his changes that's not a good idea, but perhaps I'm misunderstanding you.

> Unfortunately un-reformatting is a bit more complicated than putting in a tab for every 8 spaces, because there are places in the existing code where spaces are currently used for alignment, etc., normally assuming 4 spaces per tab.

Well, you could start by doing:
```
s/^        /\t/
s/^\t        /\t\t/
s/^\t\t        /\t\t\t/
```
so it's anchored at the start of the line. Then doing a diff against the current code and manually adjusting things. It's going to be a PITA regardless.

> I'm not fluent with GuiGuts code, but it has been my experience as a SW engineer that it is always recommended to use all spaces.

Well, this is perl, not python, and really tabs vs spaces is on a project-by-project basis. We can all rally around the bikeshed if you want. Consistency within a project is the most important, but I think we're all on the same page with that. I, personally, am a fan of spaces just for the record.

One downside to switching GuiGuts to spaces is that if other people have code like Tony's they want to include it's going to be challenging to get their code in without reformatting because of the change.

windymilla:
So sorry for the name error, Robert. Thanks anyway for the offer of help.
Nigel

windymilla:
Deleting tony branch and creating new branch without tab/space and other code formatting changes

PR #7, opened 2020-03-01
Regex & Greek changes by Tony Browne

windymilla:
Files from Tony's NEW1028.zip file, without space/tab and other minor code formatting changes. Tony wrote "These fix mostly errors relating to Greek processing plus correcting the treatment of some active regular expression strings when referenced as ordinary text plus handling of y-uml." and "There is a recently discovered bug/function deficit in Guiguts concerning the handling of the Latin-1 characters i-uml (ï) & u-uml (ü) when used in association with Greek processing and iota-dieresis (ϊ) & upsilon-dieresis (ϋ)."


windymilla:
Spacing/tab/bracket position changes reduced to a minimum. Minor alignment of columns within some files, e.g. Greek.pm, retained as it makes the code more readable around the edits made.

lib/Guiguts/Greek.pm
```diff
@@ -143,59 +157,57 @@ sub greekpopup {
...
-					  [ '6',   'cdigamma', 'ldigamma', '',        '' ],
-					  [ '90',  'ckoppa',   'lkoppa',   '',        '' ],
-					  [ '900', 'csampi',   'lsampi',   '',        '' ]
+					  ['(yu)','cupsilon', 'lupsilon', 'chupsilon', 'hupsilon' ],
+					  [ 'ph', 'cphi',     'lphi',     '',          '' ],
+					  [ 'ch', 'cchi',     'lchi',     'nch',       '' ],
+					  [ 'ps', 'cpsi',     'lpsi',     '',          '' ],
+					  [ '�',  'comega',   'lomega',   'chomega',   'homega' ],
+					  [ 'st', 'cstigma',  'lstigma',  '',          '' ],
+					  [ 'w',  'cdigamma', 'ldigamma', '',          '' ],      #ATB
```
cpeel: What are all the `#ATB`s for? Are these Tony's initials? I'm uncertain if these add value and I suspect will just confuse future developers wondering what these are.

windymilla:
I think they are Tony's initials, yes. I suppose they made more sense to be
there when Tony's edits were not under any type of source control system.
Since there are other reordering changes too, perhaps it could be argued to
keep them as a record of what Tony considered were changes worth flagging
in the absence of other comments. In the slightly longer term, they could
be removed as part of a tidying and/or documenting task? If you're strongly
against them, I'll remove them now of course.

cpeel:
I agree these made sense before source control. My suggestion would be to remove them for this commit since everything in this commit is coming in under the "from Tony" umbrella -- and not everything has those initials with it. Then it's unambitious and clear that everything in this commit came from Tony, not that some of them are from him and some of them aren't.

windymilla:
Ok. I'll tackle that, probably tomorrow as its getting later here. Would I
submit those as another commit, or revert my commit and submit a fresh one,
or start a 3rd branch? I feel I'm getting the hang of using GitHub, but not
sure of most convenient way for you. Thanks

rtonsing:
You should be able to commit and push the changes to the same branch, and the pull request will automatically update.

cpeel:
Agreed, just adding that commit to this branch and pushing it up would be all that you need. I can then squash the two together when I merge it in so they come in as one commit.

windymilla:
Comment removals committed - thanks

windymilla:
Thanks, Casey

PR #8, opened 2020-03-04
All-small-caps, default flag and title settings changed for HTML generation

windymilla:
Implemented all-small-caps when text in small caps has no lowercase letters.
Keep utf8 characters in HTML by default
Use div with CSS rather than HTML blockquote elements

### DIFF
--- a/headerdefault.txt
+++ b/headerdefault.txt
@@ -70,7 +70,7 @@ table {
     left: 92%;
     font-size: smaller;
     text-align: right;
-    font-style:normal;
+    font-style: normal;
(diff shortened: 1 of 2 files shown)


windymilla:
Sorry - hold off for a few minutes. One further minor change to include in this MR if that's OK.

windymilla:
OK to merge now, thanks. Additional commit was a switch in the order of info in the HTML title field to help distinguish books in multiple browser tabs.

windymilla:
Thanks, Nigel

PR #11, opened 2020-03-13
Add an id to illustrations added using Auto Illus Search

windymilla:
Create an id based on the illo filename (processed to adjust/remove illegal characters) and add to the div surrounding the illustration.

### DIFF
--- a/lib/Guiguts/HTMLConvert.pm
+++ b/lib/Guiguts/HTMLConvert.pm
@@ -1730,7 +1730,6 @@ sub htmlimage {
 	$selection =~ s/\[Illustration:?\s*(\.*)/$1/;
 	$selection =~ s/\]<\/p>$/]/;
 	$selection =~ s/(\.*)\]$/$1/;
-	my ( $fname, $extension );
 	my $xpad = 0;
 	$::globalimagepath = $::globallastpath
 	  unless $::globalimagepath;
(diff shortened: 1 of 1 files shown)


lib/Guiguts/HTMLConvert.pm
```diff
@@ -1865,36 +1865,35 @@ sub htmlimage {
...
 					$alt = " alt=\"$alt\"";
-					$selection = "<div class=\"caption\">$selection</div>\n"
+					$selection = "  <div class=\"caption\">$selection</div>\n"
 					  if $selection;
 					$preservep = '' unless $selection;
 					my $title = $::lglobal{titltext}->get || '';
 					$title =~ s/"/&quot;/g;
 					$title = " title=\"$title\"" if $title;
+
+					# Use filename as basis for an id for the div
```
cpeel: You might consider making this a function as it's a very contained, logical piece of code that might be useful elsewhere. These are usually called "slugs" so something like `make_slug()` perhaps.

lib/Guiguts/HTMLConvert.pm
```diff
@@ -1865,36 +1865,35 @@ sub htmlimage {
...
+					$idname =~ s/\x00/-/g;
+					# Only keep valid id characters
+					$idname =~ s/[^A-Za-z0-9_-]//g;
+
+					# Replace [Illustration] with div, img and caption
 					$textwindow->addGlobStart;
-					my $closeimg =
-"px;\">\n<img src=\"$name\" $sizexy$alt$title />\n$selection</div>$preservep";
-
-					if ( $alignment eq 'center' ) {
```
cpeel: Nice consolidation of this code!

PR #12, opened 2020-03-15
Use CSS class instead of style for illo widths

windymilla:
When created using Auto Illus Search or HTML Markup, Image, the width of the div surrounding an img was specified using the style attribute on the div element. It is now specified using a class, with sorted class definitions written into the CSS block at the top of the file


cpeel:
I'm going to trust that this does what you say it does (I have no reason to doubt that it does).

windymilla:
Thanks Casey

PR #14, opened 2020-03-17
Fix minor perl warnings in pphtml

windymilla:
Extra arg in printf due to copy from another line.
Open braces should be escaped with backslash.


windymilla:
Thanks for those bug fix merges, Casey

PR #15, opened 2020-03-22
Add bookloupe as replacement for gutcheck

windymilla:
Bookloupe is a utf-8 compatible replacement for gutcheck. Add it as a tool, and make it the default checking tool. Users can still select gutcheck within the program if required. May need additional edits when testing release process.


windymilla:
This should get the bookloupe source and windows exe into GitHub, parallel to gutcheck. Although folder structure of the Guiguts tools is unusual/inconsistent, that could be tackled separately if required. I propose leaving any such work to immediately before release when testing of making & releasing for different platforms will be needed anyway.

cpeel:
We should not be adding the bookloupe source code into the guiguts repo. If we want to include bookloupe binaries into the guiguts releases, it would be far better to have release-time tooling that fetches bookloupe, builds it, and bundles the binary into the release.

lib/Guiguts/Utilities.pm
```diff
@@ -823,7 +823,7 @@ sub initialize {
 			$::gutcommand,
 			::catfile(
 				$::lglobal{guigutsdirectory}, 'tools',
-				'gutcheck',                   'gutcheck.exe'
+				'bookloupe',                   'bookloupe.exe'
```
cpeel: nit: remove extra space so the `'tools'` and `'bookloupe.exe'` line up like before. Ditto line 874.

windymilla:
Fixes made - there were quite a few Gutcheck/Bookloupe labels to reverse

PR #16, opened 2020-03-30
Improve Search/Replace handling of page markers

windymilla:
Previously, all page markers would be moved to the start of the search/replace string. This version splits the new text around page markers in same proportion as old text. Particularly needed for changing straight quotes to curly quotes.


cpeel:
The code looks good to me, although I freely admit I don't grok everything it's doing (my perl is a decade old at this point).

Thanks for your patience on me getting to this - life has been crazy lately.

windymilla:
No problem. Thanks a lot for merging. I have another couple of branches
almost ready. I have a bit more time on my hands at the moment
unfortunately.

PR #17, opened 2020-04-05
fix for '.' not in path

t0mm0:
Settings were not persisted and other files not found because `'.'` is apparently no longer in the `@INC` path in more recent versions of perl.

I know nothing about perl so not sure if this fix is 'correct', but it seems to make things work again :wink:

### DIFF
--- a/guiguts.pl
+++ b/guiguts.pl
@@ -15,6 +15,8 @@
 use strict;
 use warnings;
 
+use lib '.';
+
 #use criticism 'gentle';
 our $VERSION = '1.0.25';
 
@@ -343,7 +345,7 @@ unless ( -e 'header.txt' ) {
 ::checkforupdatesmonthly();
 
 sub dofile {
-	my $filename = "./" . shift;
+	my $filename = shift;
 	return do $filename;
 }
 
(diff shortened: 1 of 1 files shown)


t0mm0:
hi @windymilla 

sorry about that, i'm afraid i don't have access to anything other than linux to test.

i don't use the app myself and am just trying to get it working for my dad who is stuck with me during the lockdown so can't access his usual setup :wink: and so i am not sure how to test everything. it turned out as he continued using it that more things were broken that don't use the `dofile` sub and after a bunch more googling this fix came up and seems to work for everything he has tried, and i think will stop the issues you mentioned in your testing on windows.

might be worth a try anyway!

windymilla:
Should have asked, what is your dad's name on Distributed Proofreaders? Thanks for getting him set up to continue working!

t0mm0:
I'm guessing the line you mention would include `./lib` rather than just `.`? but as I say I am not a perl expert so might be misreading that. 

It was mainly causing issues with various settings files which without this line were getting written to but not read from (`settings.rc` for example, as well as wherever page markers and labels get stored I think).

His dp username is martindp, I will tell him to get back to work now :wink:

windymilla:
Many thanks anyway. The main thing is you've got it working for him, but hopefully also a useful edit for other linux users in the future with more recent versions of perl.

cpeel:
Thanks for the fix @t0mm0!

PR #18, opened 2020-04-06
Allow checkfile from external tool to be viewed

windymilla:
When tools like gutcheck or pptxt are run, the results are displayed and the user can click the error to jump to the correct line. Add the option to "Load Checkfile" that has been created externally, in particular using the ppwb suite of tools. Also add feature (same as in gutcheck window) so that when user right-clicks to hide an error, the next error location is automatically displayed - this will now apply to other checks such as HTML validate, Tidy, etc.


cpeel:
Very nice!

windymilla:
Thanks Casey

PR #20, opened 2020-04-08
Markup index during HTML conversion (using /I...I/)

windymilla:
First commit is from previous pull request. Should be no code overlap.

Previously, had to mark index with /X...X/, convert to HTML, remove <pre>...</pre>, select index section and use Auto-Index.
Now, just mark index with /I...I/ (or /i...i/) before HTML conversion.


windymilla:
Many thanks, Nigel

PR #22, opened 2020-04-10
Disable autotoc

t0mm0:
another patch requested by martindp. This allows disabling of the auto TOC generation when converting to HTML. I hope i caught everything!


windymilla:
1) I think there is an undesirable side-effect which is that if you use this feature to disable autotoc, the h2 chapter headings don't get given their id. This means that if the PPer wants to link to them from their own ToC, they will have to manually go and put an id on each chapter heading. I don't know if that was intentional or not.
2) I also spotted a bug at line 1012, where by adding "and $autotoc" to the if, the else gets triggered when it wouldn't previously if autotoc is false. That means the second & subsequent lines of multiline chapter headings now fail to be enclosed in the <h2> markup.
3) Can you tell me more why this feature is wanted? I'm concerned that adding to the complexity of the dialog and this part of the code outweighs the overhead of just deleting the autogenerated ToC if it isn't required.
Thanks, Nigel

t0mm0:
1. This was intentional, and actually the main reason for the patch. Perhaps a separate setting?

2. Ah, I'll fix that if you do want the patch

3. martindp is extremely fussy about having classes and IDs in the generated html that he does not explicitly reference elsewhere (I've explained they don't hurt, but he insists!). This means that either he runs the generator and then manually removes all the classes/IDs he doesn't want (tedious and seems silly to generate code "wrongly" then spend ages hand editing when the computer should do what you actually want!) or maintain a separate branch with all the changes to the html generator, which is a lot of effort now the codebase is moving again ;) 

It would be great if all that stuff could be templateable rather than hard coded (for example he goes through and removes all classes from several tags like hr if I remember correctly). 

From my point of view it would be better to add configuration for as much as possible, as I'd rather write patches once instead of keep fixing lots of merge conflicts, but that is probably just me being lazy and I understand if that is not the way the project wants to go.

windymilla:
Thanks for explaining more of the background. My feeling is that unless there is a wider call for such a feature, it should not go in the main release. As you know, the code is already quite interwoven with lots of flags and dependencies. If martindp is really keen to remove all unused ids, names and classes, perhaps some regexps, or a standalone program to tidy up the file would be the best solution?

t0mm0:
ok, i will withdraw this PR and look at handling this stuff outside of guiguts.

thanks @windymilla!

PR #24, opened 2020-04-11
Character Counts search fails on backslash

windymilla:
Word Frequency, Character Counts finds all characters used in the file. If a backslash is present, then clicking on it in the list to find where it is in the file failed with an Invalid Regex popup.  Fixed by doubling backslashes :)


windymilla:
Thanks Casey

PR #26, opened 2020-04-12
Fix rewrap bug losing y-umlaut character (Tony Browne)

windymilla:
To avoid spaced ellipses being split, rewrap code converted the spaces to y-umlaut character, then converted them back to spaces at the end, hence losing any actual y-umlauts from text. Now using \x9f instead, which is not a legal character in the text, since not valid in ISO8859-1 nor utf-8.
Tony Browne made the fix but accidentally omitted it from his latest patch file which was merged previously.


windymilla:
Sorry, Casey. Hold off on this one please. I think I've found another place in the code where the same bug would happen that wasn't trapped in Tony's edit.

windymilla:
It's OK - false alarm. Similar use of a byte to mark positions, but this byte was \x7f, which couldn't be a genuine character in the text. OK to merge now, thanks.

PR #27, opened 2020-04-13
Improvements to /i...i/ index conversion

windymilla:
Improve placement of pagenum spans within lists, e.g. index, during HTML conversion. Should now be valid HTML.
Add blockquote-style rewrap margins feature to /i...i/ markup, e.g. /i[8.2.60] means index is indented by 2, continuation lines indented by 8, right margin is 60. Later numbers can be omitted.
Commonised list of block markup types - some types were omitted from some places.


cpeel:
Nice job pulling `$allblocktypes`, `$blockwraptypes` out into variables instead of having them littered throughout the code!

PR #28, opened 2020-04-16
Improve chapter hr positioning

windymilla:
Chapter hr's or chapter marks were placed before chapter headings, then
optionally moved earlier to attempt to place them before page number.
Incorrect logic meant they could move a long way back.
New code builds on chapter div work which already deals with page
numbers, and inserts chapter hr directly before chapter div.


windymilla:
Thanks for that
Nigel

PR #29, opened 2020-04-26
Improve dialog position persistence

windymilla:
Some dialogs' positions were persistent, i.e. if the user moved them,
then next time they popup in the new position, even on a new run.
All specific dialogs now have persistent positions.

If a dialog was positioned with negative coordinates, it could
cause the dialog height to be broken, only fixable by editing the
setting.rc file. Negative position coordinates, e.g. "+-123+-37"
now handled correctly.

Two pieces of history relating to dialog positions (geometry2
and ordpop) have been removed. No expected effects.

### DIFF
--- a/lib/Guiguts/Utilities.pm
+++ b/lib/Guiguts/Utilities.pm
@@ -669,7 +669,7 @@ sub initialize {
 		$::positionhash{fontpop}       = '+10+10';
 		$::geometryhash{footcheckpop}  = '+22+12';
 		$::positionhash{footpop}       = '+255+157';
-		$::geometryhash{gcpop}         = '1000x800+224+72';
+		$::geometryhash{gcpop}         = '800x600+224+72';
 		$::geometryhash{gcrunoptspop}  = '+244+72';
 		$::geometryhash{gcviewoptspop} = '+264+72';
 		$::positionhash{grpop}         = '+144+153';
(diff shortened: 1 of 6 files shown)


lib/Guiguts/Utilities.pm
```diff
@@ -653,49 +653,57 @@ sub initialize {
...
 		$::geometryhash{errorcheckpop} = '+484+72';
+		$::geometryhash{extpop}        = '+120+38';
 		$::positionhash{filepathspop}  = '+55+7';
 		$::positionhash{fixpop}        = '+34+22';
+		$::positionhash{floodpop}      = '+150+150';
 		$::positionhash{fontpop}       = '+10+10';
 		$::geometryhash{footcheckpop}  = '+22+12';
 		$::positionhash{footpop}       = '+255+157';
-		$::geometryhash{gcpop}         = '+224+72';
+		$::geometryhash{gcpop}         = '1000x800+224+72';
```
cpeel, line 672: Is 1000x800 the window size and 224/72 the position? If so, I'm somewhat wary of an initial window size that starts at 224/77 and is 1000 pixels wide and 872 high. That makes the right hand side at 1224 and the bottom at 872. Note that some laptops have a maximum screen size of 1366x768 which would place the bottom of the window well off-screen.

Basically: is this window size too big for a default / initial value?

windymilla:
Good point - I've reduced it to 800x600. It will only be used the very first time a user runs GG - once they have moved/resized the window it will store their own setting. But we don't want to confuse a newbie, so good catch. Thanks - should be ready to merge now.

windymilla:
Many thanks
Nigel

PR #30, opened 2020-05-07
Allow user to change multisearch size

windymilla:
The number of replacement terms/fields in the S&R box is a
variable stored in the resource.rc file, but there was no (friendly)
way for the user to change this. There are now plus/minus buttons
next to the single/multi switch which increase/decrease the
number of fields (range 2 to 10).


windymilla:
Thanks Casey

PR #32, opened 2020-05-11
Make HTML conversion defaults consistent

windymilla:
Some HTML conversion defaults, e.g. creating page anchors and
handling of i, b, g & f markup,  were reset each time the HTML
Generator dialog was popped. Others were initialised at start of
program & retained until the program exited.
All checkbutton and radiobutton settings now initialised once at
startup, and changes are retained until exit.

### DIFF
--- a/lib/Guiguts/Utilities.pm
+++ b/lib/Guiguts/Utilities.pm
@@ -782,24 +782,35 @@ sub initialize {
 	$::lglobal{alignstring}       = '.';
 	$::lglobal{asciijustify}      = 'center';
 	$::lglobal{asciiwidth}        = 64;
+	$::lglobal{autofraction}      = 0;				# HTML convert - 1/2, 1/4, 3/4 to named entities
 	$::lglobal{codewarn}          = 1;
-	$::lglobal{cssblockmarkup}    = 0;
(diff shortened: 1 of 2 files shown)


lib/Guiguts/Utilities.pm
```diff
@@ -782,24 +782,35 @@ sub initialize {
 	$::lglobal{alignstring}       = '.';
 	$::lglobal{asciijustify}      = 'center';
 	$::lglobal{asciiwidth}        = 64;
+	$::lglobal{autofraction}      = 0;				# HTML convert - 1/2, 1/4, 3/4 to named entities
 	$::lglobal{codewarn}          = 1;
-	$::lglobal{cssblockmarkup}    = 0;
+	$::lglobal{cssblockmarkup}    = 1;				# HTML convert - Use <div>/CSS rather than <blockquote>
```
cpeel, line 787: Was changing the default here intentional?

windymilla:
Yes. Thanks. I had previously made it set the default to 1 every time the
dialog popped, as happened with some settings previously. This change sets
it to 1 at program start, but if the user changes it, it remains until the
program exits.

windymilla:
Thanks for merging

PR #33, opened 2020-05-12
Fix external program directory name issue

windymilla:
When running an external program, $d can be used to pass the
directory name of the file being edited. This includes a trailing
slash (backslash on Windows). If the name contains spaces, then
it is surrounded by double quotes. However, a trailing backslash
followed by a double quote escapes the quote. This meant a
program such as ebookmaker which requires an output-dir
argument could not be run from within Guiguts on a file with
spaces in its pathname.
This edit provides an alternative $D which is identical to $d
but with the trailing slash removed.

### DIFF
--- a/lib/Guiguts/Utilities.pm
+++ b/lib/Guiguts/Utilities.pm
@@ -178,14 +178,19 @@ sub cmdinterp {
 			$arg = ::encode( "utf-8", $arg );
 		}
 
-# Pass file to default file handler, $f $d $e give the fully specified path/filename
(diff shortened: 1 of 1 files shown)


windymilla:
Discovered an alternative that does not require a code change. When outputting just a directory which might be quoted and has no filename after it, append . after $d, e.g. for ebookmaker change
`output-dir=$d`
to
`output-dir=$d.`
This will result in
`output-dir="C:\DP\Path Name With Spaces\."`
so the close double quote is protected from the backslash.

PR #34, opened 2020-05-22
Use em measure for image div widths

windymilla:
Because eBookMaker removes CSS declarations that end in px,
change to use em for the autogenerated width classes used by
Auto-Illus. Assumes 1em=16px, but if it isn't nothing too bad
will happen, just the caption wrapping a bit wider/narrower.
Also swapped from width to max-width as more flexible for
different screen widths.

### DIFF
--- a/headerdefault.txt
+++ b/headerdefault.txt
@@ -142,7 +142,13 @@ em.gesperrt
 .caption  {font-weight: bold;}
 
 /* Images */
-.figcenter   {
+
+img {
+    max-width: 100%;
+    height: auto;
+}
+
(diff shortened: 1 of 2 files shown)


windymilla:
No rush of course with everything else you have on

headerdefault.txt
```diff
@@ -142,7 +142,13 @@ em.gesperrt
 .caption  {font-weight: bold;}
 
 /* Images */
-.figcenter   {
+
+img {
+    max-width: 100%;
```
cpeel, line 147: Nice!

PR #35, opened 2020-05-22
Fix column copy/cut inconsistency

windymilla:
Since it's inception, there has been an mixup between column
copy and column cut. The two functions did the wrong operation,
and the accelerator text on the menu buttons was swapped. So
clicking "Column Cut (F1)" had a different effect to pressing the
F1 button, also not matching the hotkeys help text nor the
Guiguts manual. Similar swap for "Column Copy (F2)"


windymilla:
Again, no rush of course - thanks

PR #36, opened 2020-05-23
Increase image size limits in ppvimage tool

windymilla:
The size limits on images included with the HTML version are
being increased. Change the limits within the ppvimage checking
tool, and make the magic numbers constants rather than
hard-coded.

### DIFF
--- a/tools/ppvimage/ppvimage.pl
+++ b/tools/ppvimage/ppvimage.pl
@@ -887,7 +887,7 @@ sub checkepubcover {
   else {
     print LOGFILE ("\n\n");
     logprint("NOTE: epub cover will be $imgcover",$imgcoverline,"INFO","\n"," (line $imgcoverline)");
-    # warn if not jpg or smaller than 625x1000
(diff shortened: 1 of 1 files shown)


tools/ppvimage/ppvimage.pl
```diff
@@ -876,7 +887,7 @@ sub checkepubcover {
   else {
     print LOGFILE ("\n\n");
     logprint("NOTE: epub cover will be $imgcover",$imgcoverline,"INFO","\n"," (line $imgcoverline)");
-    # warn if not jpg or bigger than 600x800
+    # warn if not jpg or smaller than 625x1000
```
cpeel, line 890: Consider using your new constants here instead of these numbers since the numbers are no longer used below.

windymilla: Oops - thanks for spotting that. Done.

cpeel: No problem - that's why we do code reviews!

PR #37, opened 2020-05-25
Improve operation of hotkeys

windymilla:
If caps lock on, hotkeys failed, e.g. pressing Ctrl-x  did not do
"cut", because program was receiving Ctrl-X. Now all hotkeys
work as expected, depending on keys pressed, rather than
state of caps lock. Note that Ctrl-Shift-w can still be bound to
a different action from Ctrl-w. Attempt also made to enable
more hotkeys on Macs.

Goto Line, Goto Page and Goto Page Label dialog code has
been commonised. They can now also be dismissed with
Escape key or Window Manager close button.

Several new hotkeys added, accelerator text added to menus,
and hotkeys.txt documentation updated.

Shortcut keys for setting bookmarks was a "beta" feature from
many years ago, with a setting to enable/disable it. Now made
a permanently enabled feature.

### DIFF
--- a/lib/Guiguts/KeyBindings.pm
+++ b/lib/Guiguts/KeyBindings.pm
@@ -166,6 +166,7 @@ sub keybindings {
 				  or warn "Could not create file $!";
 			}
 			::runner('start scratchpad.txt') if $::OS_WIN;
+			::runner('open scratchpad.txt') if $::OS_MAC;
 		} );
 # Help
 	keybind( '<Control-Alt-r>',     sub { ::regexref(); } );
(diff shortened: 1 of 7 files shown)


lib/Guiguts/KeyBindings.pm
```diff
@@ -230,24 +110,132 @@ sub keybindings {
...
+	keybind( '<Alt-Right>',         sub { ::indent($textwindow, 'in'); } );
+	keybind( '<Alt-Up>',            sub { ::indent($textwindow, 'up'); } );
+	keybind( '<Alt-Down>',          sub { ::indent($textwindow, 'dn'); } );
+# Scratchpad
+	keybind( '<Control-Alt-s>',     sub {
+			unless ( -e 'scratchpad.txt' ) {
+				open my $fh, '>', 'scratchpad.txt'
+				  or warn "Could not create file $!";
+			}
+			::runner('start scratchpad.txt') if $::OS_WIN;
```
cpeel, line 168: FWIW, `open scratchpad.txt` will open the text file in the default editor on MacOS.

lib/Guiguts/KeyBindings.pm
```diff
@@ -230,24 +110,132 @@ sub keybindings {
...
+sub keybind {
+	my $textwindow = $::textwindow;
+	my $lkey = shift;				# Key-combination (lower-case letter)
+	my $subr = shift;				# Subroutine to bind to key/event
+	my $event = shift;				# Optional event argument
+	
+	
+	my $mkey = my $ukey = $lkey;
+	$ukey =~ s/-([a-z])>/-\u$1>/;			# Create uppercase version
+	$mkey =~ s/Control/Meta/ if $::OS_MAC;	# Create Mac alternative to Control key
```
cpeel, line 225: This gives me pause for a couple of reasons.
1. As someone who regularly switches between Linux and Mac I'm used to one or the other button working -- not both.
2. More importantly: what happens on Macs when the new shortcut keys override system defaults (Meta+W on Macs close the window -- will it rewrap now, close the window, or do both?). I like that this would align some of the common meta keys to their Mac equiv (Meta+C for copy, for instance) but worry this will break other expected uses of those keys.

My US$0.02 -- I defer to you but ask that we confirm what this actually does on Macs for all of the expected key combinations.

windymilla:
I see - that's clarified my understanding of the Mac situation. I thought I
was doing Mac users a favour, but I can see now it's probably not. I'll
remove the automatic Control/Meta addition and just put back the previously
defined Mac Meta keys - I removed all but Meta-q because they were covered
by the automatic code. It will probably be tomorrow before I get onto that.
Thanks - I'll let  you know when it's done.

windymilla:
OK - restored previous Meta shortcuts only. New ones only work with Control as for Windows. If Mac users in the future want to request specific shortcuts, they can be added. Thanks.

PR #39, opened 2020-05-28
Make highlight quotes function include curly quotes

windymilla:
Include curly quotes as well as straight ones when highlighting
single or double quotes. Also created common routines for
operation through default/old menu entries, key bindings
and dialog.

### DIFF
--- a/lib/Guiguts/Highlight.pm
+++ b/lib/Guiguts/Highlight.pm
@@ -6,7 +6,7 @@ BEGIN {
 	use Exporter();
 	our ( @ISA, @EXPORT );
 	@ISA    = qw(Exporter);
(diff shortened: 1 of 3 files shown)


windymilla:
Because I based this branch on my accelerators branch, it has those two commits in as well. I think you gave me a way before to "remove" those commits using git rebase -i. However, when I tried it, I hit a snag regarding merging changes, perhaps because KeyBindings.pm was also edited in the accelerators commits which was perhaps a different circumstance to the previous time. If there's a way to do it, I'm very happy to learn it. Or if it is easier, I can scrap this branch and create a new one from master after the accelerator code has been merged.

cpeel:
I rebased this branch off of master after merging in the accelerators branch so it looks good now.

PR #40, opened 2020-05-31
Display gutcheck/bookloupe errors correctly

windymilla:
When user right-clicks on a gutcheck/bookloupe error, the line
is removed from the list of errors. However, if lines were
removed from the header of the GC/BL output, then when the
error list was refreshed (due to change of GC options), some
error lines jumped up into the header section. Also, deleting
a blank line caused all blank lines to vanish on refresh.
Both issues now fixed.


windymilla:
Thanks

PR #41, opened 2020-05-31
Add Count button to S&R dialog

windymilla:
Add new button to count how many times current search would
find a match (including regex, whole word & case flags). Uses
same label as current part-solution which relies on a list of
"seen words" and only works for whole word searches. Also
added shortcut key Ctrl-n to both main window and S&R
dialog. Fixed Ctrl-g/Ctrl-Shift-g issue on dialog to match fix
made to main window hotkeys (MR #37)

### DIFF
--- a/lib/Guiguts/SearchReplaceMenu.pm
+++ b/lib/Guiguts/SearchReplaceMenu.pm
@@ -329,7 +329,7 @@ sub countmatches {
 	$::lglobal{selectionsearch} = $saveselectionsearch;
 }
 
-{
+BEGIN { # restrict scope of $countlastterm
 	# remember last term counted / searched
 	my $countlastterm = '';
 	# only need to clear counted label if current term is different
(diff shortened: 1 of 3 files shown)


lib/Guiguts/SearchReplaceMenu.pm
```diff
@@ -301,11 +289,59 @@ sub searchtext {
...
+	
+	# restore saved globals
+	$::searchstartindex = $savesearchstartindex;
+	$::searchendindex = $savesearchendindex;
+	$::sopt[2] = $savesopt2;
+	$::sopt[4] = $savesopt4;
+	$::lglobal{selectionsearch} = $saveselectionsearch;
+}
+
+{
```
cpeel, line 332: Is this suppose to be a part of the prior function? Is it a function itself? It seems lost.

windymilla: No, the purpose of the block is to enclose the variable with the function so that it is persistent but not global. I've added a comment to that effect. Also learnt it's best practice to make it a BEGIN block to ensure correct initialisation (though I don't think it would matter in this case).

windymilla:
Thanks

PR #42, opened 2020-06-07
Remove unused files & add MacOS instructions

cpeel:
This MR does two notable things:
1. Removes unused perl modules
2. Adds instructions for how to make Guiguts work on MacOS


cpeel:
@srjfoo - I would appreciate validation of the MacOS instructions. These instructions seemed to work for me.

srjfoo:
I'll be happy to!

cpeel:
Unless you also installed Tk::Toolbar from CPAN into your perl installation, this is expected. Guiguts still needs that and Tk::LineNumberText -- we just want it to get and use the CPAN version instead of bundling it in.

windymilla:
Yep. That makes sense

cpeel:
I removed the commit that removed the CPAN modules. I want to push that in with another commit that address some of the larger Windows/perl question.

windymilla:
Not important, but maybe of interest: my GG is running happily without lib/Tk/LineNumberText, though it does have lib/Guiguts/LineNumberText (looks like an edited/stripped down version)

cpeel:
> Not important, but maybe of interest: my GG is running happily without lib/Tk/LineNumberText, though it does have lib/Guiguts/LineNumberText (looks like an edited/stripped down version)

Indeed - it works for me without that too. I've removed it from the list of needed modules.

This branch is now rebased off of HEAD.

cpeel:
After more testing (and with help from @srjfoo) I've flushed out the [MacOS instructions](https://github.com/cpeel/guiguts/blob/remove-unused-files/INSTALL.md#macos) more. I've also added a helper script to install the perl modules which we should be able to reuse on Windows as well.

INSTALL.md
```diff
@@ -0,0 +1,140 @@
...
+perl guiguts.pl
+```
+
+## Perl Modules
+
+Guiguts requires the following Perl modules to be installed via CPAN:
+
+* LWP::UserAgent
+* Tk
+* Tk::ToolBar
```
srjfoo, line 140: `HTML::TokeParser` isn't actually needed, after all? (That seems to have disappeared form the instructions.)

INSTALL.md
```diff
@@ -0,0 +1,86 @@
+# Installation
+
+Upgrading requires a complete reinstall (save `header.txt`, `setting.rc`,
+and any files you wish to keep in the `data/` directory, although in
+theory these should not be overwritten).
+
+Please direct any help requests to the
+[DP forum Guiguts thread](http://www.pgdp.net/phpBB2/viewtopic.php?t=3075).
```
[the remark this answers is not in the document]
cpeel: Yes, I'll get that updated.

INSTALL.md
```diff
@@ -0,0 +1,86 @@
...
+using `cpanm`.
+
+For example:
+```
+cpanm --notest --install Tk
+```
+
+Now unzip `guiguts-n.n.n.zip` and start Guiguts with:
+```
+perl guiguts.pl
```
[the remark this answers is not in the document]
cpeel: Also a good idea!

INSTALL.md
cpeel: It is needed but as several people have pointed out, it is included with the base perl install for both Windows and Mac so I opted not to call it out (like we don't call out all the other modules needed but included with perl).

srjfoo: Makes sense -- I hadn't realized it was part of the base perl install, just noticed the difference between these instructions and the earlier ones.

cpeel:
We've had some good reports on the Mac instructions, so I think this is good for merging pending OK by @windymilla.

windymilla:
Yes, happy with this

PR #44, opened 2020-06-09
Assume we have a Unicode-capable perl install

cpeel:
Rather than attempting to build the Unicode data on-the-fly, simply return an error if the necessary Unicode files are not available. `unicore` is apart of all modern Perl distributions.


windymilla:
Looks fine

PR #45, opened 2020-06-09
Convert THANKS.txt to Markdown

cpeel:
Tweak the contributor listing to be alphabetical and add windymilla!

I also adjusted the help text a bit to deemphasize gutcheck since we will be replacing that with bookloupe in the next release.


cpeel:
> I see hanne is not in the list - is that deliberate?

No, that was an oversight and has been fixed - thanks for catching that.

cpeel:
Any additional thoughts on this @windymilla?

PR #46, opened 2020-06-09
Restructure directory layout in preparation for release tooling

cpeel:
Most tools have their source code in a `src/` directory within their repo. This moves to that structure. The big upside is that we then have a clear separation of:
* Guiguts source code
* project metadata & code the supports building the release

This is a stepping stone for updates to the release process that pulls in updated tools as part of building the release. This also nukes the huge `other.zip` file that includes things we no longer want to bundle with the release as-is.

Note that the `src/` directory does not have a `tools/` directory now. This is intentional as it is created as part of the release. A `src/tools/` directory is ignored by git, however, so developers can create it and populate it with tools and still use Guiguts from the `src/` directory.


windymilla:
Oh, one more. After running, there were 3 files identified as changed by Git: header.txt, setting.rc in the top folder and data/labels_en.rc

cpeel:
`run_guiguts.bat` should be run from `src/` now.

Those three files will now be created by GG under `src/` but since they aren't managed by git they remain where they were from other branches. If you move them to `src/` when using my branch git won't show them as changed.

Did you copy `tools/` into `src/`? That should allow GG to find Jeebies, etc.

windymilla:
You're right of course. The 3 files were left behind from my previous run on another branch. I removed them and they didn't get recreated there. GG found the tools fine when I copied them as you said.
So all fine, thanks.

cpeel:
I want to get the Strawberry Perl branch merged before this one. I also want to make sure we get any of your outstanding branches merged before this one goes in as otherwise it will just create a bunch of merge conflicts for you.

I'm not a in a big hurry for this one as I can continue to work on the tool bundling at build directly off of this branch.

PR #47, opened 2020-06-09
Don't start full-screen on first run

cpeel:
Dual monitors, at least on Macs, can result in the initial window opening so big that you can't actually resize it.


windymilla:
Also may have fixed issue I recently noticed with the new method of running from within src. I often got a very small startup window on first run when there was no setting.rc. Now a (presumably) 800x600 window appears :)

PR #49, opened 2020-06-09
Fix pgdp.net URLs

cpeel:
Use HTTPS URLs for all pgpd.net links and fix broken links.

I took my best guess at what URL `$url_no_proofer` pointed to in phpBB2. That URL has been wrong for years now.


srjfoo:
I have no idea what `$url_no_proofer` is, either, but the original looks like it points to composing a private message. Everything else looks good to me.

PR #50, opened 2020-06-09
Use Strawberry Perl on Windows

cpeel:
This updates the installation documentation to use Strawberry Perl and tweaks the batch file to prefix the Strawberry Perl directory to the path so it's the interpreter used to start Guiguts.

I've included a section on how to allow multiple interpreters to coexist.


INSTALL.md
```diff
@@ -11,25 +11,78 @@ See also https://www.pgdp.net/wiki/PPTools/Guiguts/Install
...
+path.
 
-Other distributions of versions of Perl, such as
-[Strawberry Perl](http://strawberryperl.com/) or
-[ActiveState Perl](https://www.activestate.com/products/perl/) may be used
-after installing additional [Perl modules](#perl-modules). Strawberry Perl 5.12
-or earlier and ActiveState Perl 5.10 or earlier, are both known to work. Later
-versions of ActiveState Perl will not successfully install Tk and cannot be used
-with Guiguts.
+If you have multiple Perl distributions installed you should edit
```
srjfoo, line 63: insert "the" between "edit" and "`run_guiguts.bat`"? (or remove "file")

INSTALL.md
```diff
@@ -11,25 +11,78 @@ See also https://www.pgdp.net/wiki/PPTools/Guiguts/Install
...
+used in 1.0.25 which may not be the full set needed by later versions.
+
+### Starting Guiguts
+
+Start Guiguts with:
+```
+run_guiguts.bat
+```
+
+### Helper applications
```
srjfoo, line 85: Would it be good to list the most common helper applications?

INSTALL.md
```diff
@@ -11,25 +11,78 @@ See also https://www.pgdp.net/wiki/PPTools/Guiguts/Install
...
+
+You can install them manually with `cpanm` however if you want, for example:
+```
+cpanm --notest --install LWP::UserAgent
+```
+
+#### Other Perl distributions
 
-It does not include an image viewer or a spell checker.
+Strawberry Perl is the recommended Perl interpreter as that is what the
```
windymilla, line 52: Maybe this paragraph should be first?

INSTALL.md
```diff
@@ -11,25 +11,78 @@ See also https://www.pgdp.net/wiki/PPTools/Guiguts/Install
...
+developers have tested with and supports the latest version of Perl and
+all necessary Perl modules. It can coexist along side other interpreters
+assuming they have the necessary Perl modules available.
 
-To install, unzip the the file `guiguts-win-n.n.n.zip` into a directory. This
-extracts Guiguts as well as a bundled version of Perl 5.8. To run Guiguts, change
-into the directory you extracted the files into and run `run_guiguts.bat`. This
-batch file prepends the bundled version of Perl onto your path and starts the
-program.
+When installing the Perl modules, either with the helper script or manually
```
windymilla, line 57: Is it worth explaining how to do this? I have already received messages of nervousness regarding returning to an unbundled perl, so it's important for us to provide less technical users with as simple a guide as possible. Also, should this paragraph go at line 35, i.e. straight after installing Perl and before running the cpan install. I think avoiding mention of cpan except in optional or "for experts" docs would also help the less-technical user.

run_guiguts.bat
```diff
@@ -1,10 +1,14 @@
 @echo off
 
-cd %~dp0
+REM Prefix Strawberry Perl to the path
+set PATH=C:\Strawberry\c\bin;C:\Strawberry\perl\site\bin;C:\Strawberry\perl\bin;%PATH%
```
windymilla, line 4: I have had issues with this method of adding to the path. If run_guiguts is run from a command window, then after it finishes, the extra folders are still in your path, so if you run it again, it prepends or appends again. After a few runs, it dies presumably because the command line length exceeds some limit. I tried removing all the path setting and things worked fine - we don't need python27 for anything, and kindlegen & tidy are only for ebookmaker not for GG. Adding them to the system path is done during ebookmaker installation. So we could maybe remove all the PATH setting?

PR #51, opened 2020-06-12
Update tool packaging

cpeel:
This creates a packaging framework that pulls down the tools upon release build. Each tool gets its own `README.md` that we include with the release that gives a brief overview of what it is and where it came from. Tools also have a `package.sh` that is responsible for packaging that tool upon release for the specified platform. To add a new tool to Guiguts is a simple case of creating a new directory under `tools/` with those two files.

As part of this, I've added tweaks to the `setdefaultpath()` function so it will find tools on the user's path (most useful for Macs) in addition to in the tools directory but without extensions (also useful for Macs).

Example builds for this code (with hacked-in version) are:
* [guiguts-win-1.1.0-alpha2.zip](http://www.pgdp.org/~cpeel/forumPosts/guiguts-win-1.1.0-alpha2.zip) - 21MB
* [guiguts-mac-1.1.0-alpha2.zip](http://www.pgdp.org/~cpeel/forumPosts/guiguts-mac-1.1.0-alpha2.zip) - 16MB


cpeel:
> HTML Validation failed - [...]

I've added the xhtml* files back and only extracted the single .exe (and the .dll which I think it also requires).

> CSS Validation ran successfully using CSS3+SVG. It did output a single unnecessary line to the command window stating which options it had used. We could probably use 2>&1 when we call it from within GG to capture that line to the GG CSS results window rather than the command window - users are asked to report messages that appear there.

That sounds like a good idea. I'd like to do that separately from the packaging MR (or you can throw up an MR with it too :).

> I couldn't find a need for the other .exe files in tools/W3C nor the .jar files in tools/W3C/lib. I could do both HTML and CSS validation without them.

The .jar files in tools/W3C/lib are interesting, because I _swear_ I couldn't get the validator working without them before ... but it does now for me too. <shrug> I'll pull the code to import them out and save it later in case we decide they are needed after all.

cpeel:
CSS and HTML packaging has been updated. I've spun a new [guiguts-win-1.1.0-alpha2.zip](http://www.pgdp.org/~cpeel/forumPosts/guiguts-win-1.1.0-alpha2.zip) if you want to validate that I have all the right CSS and HTML bits in place.

windymilla:
Both HTML and CSS validation work correctly, thanks. You were right about the dll being needed, the .exe files (apart from onsgmls.exe) are not needed and can be removed from the release. I'm sure the jobs they do are useful for some people, but not us :)

cpeel:
> You were right about the dll being needed, the .exe files (apart from onsgmls.exe) are not needed and can be removed from the release.

Bugger, I updated the `W3C/package.sh` to only extract the exe and the dll but clearly it didn't do that correctly or I messed up the packaging itself. Thanks for pointing that out -- I will get that fixed/verified.

cpeel:
This version has the `tools/W3C/` content that I am expecting [guiguts-win-1.1.0-alpha3.zip](http://www.pgdp.org/~cpeel/forumPosts/guiguts-win-1.1.0-alpha3.zip) -- including no `lib/` and only the single EXE & DLL.

windymilla:
That looks great now - thanks

cpeel:
After testing some ebookmaker runs on MacOS I discovered that the version of tidy included with MacOS is too old. I've updated the `INSTALL.md` to have it installed via Homebrew.

Also the version of KindleGen Amazon provides for MacOS is 32-bit only, which won't work on Catalina. There's a manual work-around which I've included in the `kindlegen/README.md` but until Amazon fixes their packaging I don't see a way around it. I've touched base with several of my friends who work at Amazon to see if we can get that fixed.

With that and @windymilla's validation of the CSS/HTML bits on Windows I think this is good to merge assuming he and @srjfoo agree.

windymilla:
Yes. Agreed. Well done!

PR #54, opened 2020-06-15
Allow run_guiguts to be run from other folders

windymilla:
Historically, run_guiguts.bat had a "cd" to move to the
folder where guiguts.pl resides before running it.
Recent edit removed the cd, which meant batch file would
only execute correctly if run from the correct folder. This
caused Windows' "Open with" right-click functionality to
break, since this always runs from a system folder, like C:\Windows\system32.

Use %~dp0 to get name of folder containing
run_guiguts.bat and %+f1 for full pathname of file to
be edited.

Also use %~dp0 instead of %cd% (current folder) when
adding tools folders to PATH.

This solution is better than historical use of cd to change
folder, since when run from a command window, that
caused the current folder in the shell to be changed
once the batch file had exited.

### DIFF
--- a/src/run_guiguts.bat
+++ b/src/run_guiguts.bat
@@ -1,5 +1,10 @@
 @echo off
 
+REM Get the Guiguts install directory as the directory containing this batch file
+set GG_DIR=%~dp0
+REM Get full pathname of edit file from first argument to batch file
+set ED_FILE=%~f1
+
 REM Save original path to restore later. This prevents the path from being
 REM updated at every run.
 set OLDPATH=%PATH%
@@ -9,14 +14,14 @@ set PATH=C:\Strawberry\c\bin;C:\Strawberry\perl\site\bin;C:\Strawberry\perl\bin;
 
 REM Add older bundled perl directory to the path at the end as a fall-back
 REM for users who want to try and use it instead of Strawberry Perl.
-set PATH=%PATH%;%~dp0perl;%~dp0perl\lib
+set PATH=%PATH%;%GG_DIR%perl;%GG_DIR%perl\lib
 
 REM Prefix tooling directories
-set PATH=%~dp0tools\kindlegen;%~dp0tools\tidy;%PATH%
+set PATH=%GG_DIR%tools\kindlegen;%GG_DIR%tools\tidy;%PATH%
 
 REM Find guiguts.pl in same folder as this batch file
 REM Use full pathname of file to be edited
-perl %~dp0guiguts.pl %~f1
+perl %GG_DIR%guiguts.pl %ED_FILE%
 
 REM Restore original path
 set PATH=%OLDPATH%
--- a/src/run_guiguts.bat
+++ b/src/run_guiguts.bat
@@ -1,6 +1,7 @@
 @echo off
 
 REM Get the Guiguts install directory as the directory containing this batch file
+REM using arcane Windows batch file magic.
 set GG_DIR=%~dp0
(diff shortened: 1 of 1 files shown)


cpeel:
What is `%~dp0` suppose to do, because your batch file doesn't work on my Windows system.

cpeel:
> What is `%~dp0` suppose to do, because your batch file doesn't work on my Windows system.

Nevermind, helps if I Google that. Wow that's unintuitive.

src/run_guiguts.bat
```diff
@@ -9,14 +9,14 @@ set PATH=C:\Strawberry\c\bin;C:\Strawberry\perl\site\bin;C:\Strawberry\perl\bin;
 
 REM Add older bundled perl directory to the path at the end as a fall-back
 REM for users who want to try and use it instead of Strawberry Perl.
-set PATH=%PATH%;%cd%\perl;%cd%\perl\lib
+set PATH=%PATH%;%~dp0perl;%~dp0perl\lib
```
cpeel, line 12: It would be less arcane if we simply set this into a variable and used that variable everywhere. For instance:
```
REM Get the Guiguts install directory as the directory containing this batch file
REM using arcane Windows batch file magic.
set GG_DIR=%~dp0
...
set PATH=%PATH%;%GG_DIR%perl;%GG_DIR%perl\lib
```

windymilla:
Is that better? Does it work OK on your Windows system now?

cpeel:
That's much easier to understand, thanks (I was trying to type the commands in from the command line and they weren't working -- obvious to me now since they're only for batch files). Although honestly I would leave in a comment about `%~dp0` being an arcane batch file thing, otherwise future developers might, like me, wonder if this is some hard-coded DP-related thing.

Sadly, no. Before it worked fine in the `C:\Users\Casey Peel\Downloads\guiguts` directory, but now it fails because there is a space. This works, however:
```
perl "%GG_DIR%\guiguts.pl" "%ED_FILE%"
```

windymilla:
Good catch. I'll look at it tomorrow

cpeel:
@srjfoo mentioned that we probably need to do some other GG tweaks to make paths-with-spaces work/work better. I'm hoping it's mostly a case of applying judicious quotes to the command runner.

srjfoo:
> @srjfoo mentioned that we probably need to do some other GG tweaks to make paths-with-spaces work/work better. I'm hoping it's mostly a case of applying judicious quotes to the command runner.

It would be good to, if we can. I know I've run into the problem before, and modern OSs (and users) think nothing of including spaces in both filenames and directories.

windymilla:
The spaces issue in run_guiguts.bat should be fixed now. I have used GG a lot to edit files in folders that have spaces in their name, so the issues @srjfoo mentioned may be historical. I also tested putting bookloupe in a folder named "book loupe" and changing it via Set File Paths - it works OK. If we find specific areas still sensitive to spaces, we can fix them later. I think it should be ready to merge now. Thanks

srjfoo:
> The spaces issue in run_guiguts.bat should be fixed now. I have used GG a lot to edit files in folders that have spaces in their name, so the issues @srjfoo mentioned may be historical. I also tested putting bookloupe in a folder named "book loupe" and changing it via Set File Paths - it works OK. If we find specific areas still sensitive to spaces, we can fix them later. I think it should be ready to merge now. Thanks

It may just be a Mac/*nix problem, and since this MR is just for Windows, I think we can defer dealing with it.

PR #55, opened 2020-06-18
Add .bak files to global .gitignore

windymilla:
WinMerge (& maybe other tools) create backup files with
the extension .bak


windymilla:
Thanks

PR #57, opened 2020-06-20
Update extops list and allow an arbitrary number of them

cpeel:
This removes deprecated extops entries and adds EBookMaker and Google Ngram (with the latter using the selected words as the search entry).

This also removes the limit on the number of entries in the extops list entirely.

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -127,9 +127,6 @@ our $globalbrowserstart = $ENV{BROWSER};
 if ( !$globalbrowserstart ) { $globalbrowserstart = 'xdg-open'; }
 if ($OS_WIN)                { $globalbrowserstart = 'start'; }
 if ($OS_MAC)                { $globalbrowserstart = 'open'; }
-our $globalfirefoxstart = 'firefox';
-if ($OS_WIN) { $globalfirefoxstart = 'start firefox'; }
-if ($OS_MAC) { $globalfirefoxstart = 'open -a firefox'; }
 our $globalimagepath        = q{};
 our $globallastpath         = q{};
 our $globalspelldictopt     = q{};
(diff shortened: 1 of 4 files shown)


cpeel:
> 1. Would it be more logical to re-order the defaults slightly: browser, Onelook, Ngram, Shape catcher, Markup, CSS, Ebookmaker. After browser come check word/character options, then come file checking tools.
>    Minor edit suggestions:
>    Rename Ebookmaker to Ebookmaker Online (note capitalisation of Ebookmaker)
>    Rename Markup option to "W3C HTML Validation Service"
>    Change urls for HTML & CSS validators to
>    https://jigsaw.w3.org/css-validator/#validate_by_upload+with_options
>    https://validator.w3.org/#validate_by_upload+with_options

Happy to reorder them and make these updates. re: ebookmaker - they don't know how the hell they want to capitalize the stupid thing. The program is called `ebookmaker` and yet the usage statement of the program and the [docs for the program](https://github.com/gutenbergtools/ebookmaker) say EBookMaker. The website says EBookmaker. I'm happy to rename it to whatever, but whatever we choose is going to be both right and wrong.

> 2. Is deleting blank entries most helpful for the user, or leaving them to shuffle entries around? Alternative would be to retain all entries until the last non-blank plus one.

Can you elaborate on the use-case you are envisioning? If you want to always include a blank one in the file that should be pretty easy. Having multiple blank ones is going to be more challenging.

> 4. Dialog size is an issue. When new entries are added, the "Close" button gets pushed off bottom of dialog. At first I didn't realise and so was crossing off the dialog which doesn't save changes.

Good point - I will see what I can do about this.

> 5. Would it be better to change "Close" to "OK" to emphasise that it saves your changes?

Yes - I will make that change. Good catch.

windymilla:
1. Yes, I can see the docs aren't consistent on capitalisation. I'll go with your capitalisation in the new EBookMaker menu option then.
2. I was thinking of a user wanting to insert an option near the top of the list, maybe a website they often use, and so they moved some other options down to make a space in the middle of the list, which could then get closed up again. On reflection, it's not an everyday activity, so keep it as you have it - thanks.

So, with Close changed to OK, and the minor re-ordering/renaming, I'm happy for it to be merged.

cpeel:
Updated extops order (and names & URLs) are in-place. I have also disabled persisting the dialog geometry for the External Programs dialog which seems like the best way to make sure that it is always big enough to show all of the contents.

I did not change the Close button. After looking we are _wildly_ inconsistent on our dialog labels across the codebase. I will put up a separate MR that updates this and other dialogs to be consistent.

windymilla:
There are two hashes used for persistent dialog position/size. The geometryhash stores size and position, but positionhash just stores position. Rather than introducing a bypass mechanism, would it work to just move extpop from geometryhash to positionhash? The dialog might need a new name (extpopv for variable?) to cope with loading of old setting.rc files. A future review of which dialogs are in which hash might be in order. I think geometryhash should be reserved for those that the user would want to resize, e.g. error output from tools, but it's possible over the years that some dialogs have ended up in the wrong list.

cpeel:
Yes, that is a much better solution. I didn't understand those two hashes until you explained it to me, so thank you.

windymilla:
Thanks - that looks fine.

PR #58, opened 2020-06-21
Standardize dialog buttons & titles

cpeel:
This standardizes when we use `OK` vs when we use `Close` in dialogs. I've added a style guide as part of this.

I've also massaged dialog titles to match the menu item they came from and other small spacing and capitalization changes to make things more consistent.

The only functional change is adding an OK button to the Set Margins dialog.

### DIFF
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -1,0 +1,28 @@
+# Style Guide
+
+When adding or changing user interface elements in Guiguts, use this style
+guide as a reference. Guiguts was developed by numerous people over many years
+and this style guide is very recent, so not all parts adhere to this (yet).
+
+## Dialog buttons
+
+Most Guiguts users are on Windows, so we use the
+[Windows UX guide](https://docs.microsoft.com/en-us/windows/win32/uxguide/win-dialog-box)
+for dialog button styles.
+
+### Ordering
+
+Dialog buttons are in left-to-right order following the Windows convention:
+
+* OK / Cancel
+* Yes / No / Cancel
+
+Note that "OK" should be capitalized.
+
+### OK vs Close
+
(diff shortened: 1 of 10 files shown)


windymilla:
When I cross off the Margins dialog with the Window Manager cross, it does disappear, but I don't think it's getting destroyed properly now. When I go to pop it up again, I get nasty error messages and it doesn't appear. I think it still needs WM_DELETE_WINDOW to destroy and undef it.

To possibly consider for a future release: should we have OK & Cancel buttons for this type of dialog? Let's not pull too hard on the thread of making GG consistent at this stage in case we unravel too much :)

cpeel:
> When I cross off the Margins dialog with the Window Manager cross, it does disappear, but I don't think it's getting destroyed properly now. When I go to pop it up again, I get nasty error messages and it doesn't appear. I think it still needs WM_DELETE_WINDOW to destroy and undef it.

Good catch! I updated the dialog to open with destroy which seems to fix it.

> To possibly consider for a future release: should we have OK & Cancel buttons for this type of dialog? Let's not pull too hard on the thread of making GG consistent at this stage in case we unravel too much :)

I looked into that some but my initial attempt didn't work because for most things we bind the global variables into the dialog so the values are persisted even if we hit Cancel. I could have easily been doing it wrong however. Agreed we should consider doing this, but not now :)

windymilla:
Ah yes - I guess a sprinkling of entry fields through the code will be linked to global variables, rather than locals, with OK assigning them to the globals. As you say, something for later (or never!)

windymilla:
Yes, works fine now, thanks.

cpeel:
A force-push to resolve a merge conflict. Now I'll merge this in.

PR #59, opened 2020-06-21
Add EbookMaker to run from HTML menu

windymilla:
- Add EBookMaker to Set File Paths dialog.
- Initialise location of EBookMaker.
- Commonise executable location code.
- Add menu entry for EBookMaker.

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -17,6 +17,7 @@ BEGIN {
 	  &xtops &toolbar_toggle &toggle_autosave &killpopup &expandselection &currentfileisunicode &currentfileislatin1
 	  &getprojectid &setprojectid &viewprojectcomments &viewprojectdiscussion &viewprojectpage
 	  &scrolldismiss);
+
 }
 
 sub get_image_file {
@@ -1772,9 +1773,20 @@ sub ebookmaker {
 
 	print "\nBeginning ebookmaker\n";
 	print "Files will appear in the directory $::globallastpath.\n";
-	
+
+	# EBookMaker needs to use Tidy and Kindlegen, so temporarily append to the path
+	my $tidypath = ::catdir( $::lglobal{guigutsdirectory}, 'tools', 'tidy' );
+	my $kindlegenpath = ::catdir( $::lglobal{guigutsdirectory}, 'tools', 'kindlegen' );
+	# local variable means global value will be restored at end of sub
+	local $ENV{PATH} = pathcatdir( pathcatdir( $ENV{PATH}, $tidypath ), $kindlegenpath );
+
 	::runner( $::ebookmakercommand, "--verbose", "--max-depth=3", "--make=epub.images", "--make=kindle.images",
 	          "--output-dir=$outputdir.", "--title=$fname", "$filepath" );
+}
+
+# Concatenate given directories using OS-specific path separator
+sub pathcatdir {
+	return $_[0] . $Config::Config{path_sep} . $_[1];
 }
 
 ## Sidenote Fixup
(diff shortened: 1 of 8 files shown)


windymilla:
@srjfoo wanted to check the running of EBM from within GG on a Mac

windymilla:
New commit adds tidy & kindlegen to path. Local variable means that once sub ebookmaker exits, path will restore to previous value. Tested on Windows by removing path change from run_guiguts.bat, then realised there was no point in putting it back in.
Hopefully ready to merge.

src/lib/Guiguts/Utilities.pm
```diff
@@ -1747,6 +1755,40 @@ sub seeindex {
...
+	my $kindlegenpath = ::catdir( $::lglobal{guigutsdirectory}, 'tools', 'kindlegen' );
+	# local variable means global value will be restored at end of sub
+	local $ENV{PATH} = pathcatdir( pathcatdir( $ENV{PATH}, $tidypath ), $kindlegenpath );
+
+	::runner( $::ebookmakercommand, "--verbose", "--max-depth=3", "--make=epub.images", "--make=kindle.images",
+	          "--output-dir=$outputdir.", "--title=$fname", "$filepath" );
+}
+
+# Concatenate given directories using OS-specific path separator
+sub pathcatdir {
```
cpeel, line 1788: Can you please add this to the export list so I can use it in `Preferences.pm` as well?

cpeel:
Functionally this worked great for me on Mac. A small request and I think this is ready to merge.

src/lib/Guiguts/Utilities.pm
cpeel: Actually, strike that - the code I have in place for `setdefaultpath` doesn't need updating so this is good as-is.

srjfoo:
I realize that I'm late with responding, but thought I'd mention it, anyway -- finally got the right code. ebookmaker worked fine for me, creating epubs, though kindlegen did not. The error message I got for kindlegen:

```
ERROR:root:KindleWriter: kindlegen [Errno 86] Bad CPU type in executable: 'kindlegen'
WARNING:root:
INFO:root: kindle.images made in 0:00:00.005271
```

(And of course, in spite of what it said, it did not make the kindle.images file)

windymilla:
I think I may have seen a comment somewhere that Kindlegen as released
won't run on 64 bit Macs.

Yes. Googling comes up with some hits. One said that it's included in
Kindle Previewer 3, and said the file is here:

/Applications/Kindle\ Previewer\ 3.app/Contents/MacOS/lib/fc/bin/kindlegen

Not sure what we do about that, if that is the problem you have?

srjfoo:
Casey walked me through pulling the 64 bit version kindlegen out of Kindle Previewer 3, so that's what should be in my tools directory. I can check tomorrow to see if that's the case, though -- it's probably time for me to stop for the night.

I know that it works on my system, though, because the last itme I ran ebookmaker from the command line, the mobi was generated quite nicely.

cpeel:
Nigel is correct, the version of kindlegen it is running is the i386 version which won't run on Catalina.

Expecting this to come up in general, the [tools/kindlegen/README.md](https://github.com/DistributedProofreaders/guiguts/tree/master/tools/kindlegen) that we include with the releases talks about this.

srjfoo:
I probably just missed something, then -- I had already extracted the 64 bit kindlegen from KP3. I'll poke some more and make sure it wound up where it was suppsed to be.

windymilla:
Ah. The readme must be where I recalled seeing it. For it to work from the menu entry, kindlegen would have to be in tools/kindlegen below your main GG folder

srjfoo:
Yes, that was it -- I hadn't copied the correct kindlegen into the alpha4 tools folder, so when I copied the alpha4 tools to the working draft, I had the wrong kindlegen. No problem creating the mobi this time. Thanks!

PR #60, opened 2020-06-23
Move GG 1.0 changelog from DP wiki to CHANGELOG

cpeel:
This is a straight lift of https://www.pgdp.net/wiki/Guiguts_new_features_and_bug_fixes to the `CHANGELOG.md` file, after which I want to delete the former wiki page to avoid confusing GG users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -258,5 +258,100 @@ and jeebies directories without the .exe files to the guiguts-n.n.n.zip
 file.
 
 ## Version 1.0.0
-See https://www.pgdp.net/wiki/Guiguts_new_features_and_bug_fixes for a
-description of new features and bug fixes.
+Relative to version 0.2.10, the main changes in version 1.0 are
+
+1. One click installation on Windows and Macintosh/OSX computers with no need
+   to install perl separately (see guiguts-win and guiguts-mac zip files)
+2. Several major new features, including running all HTML checks with one
+   button, side by side viewing of text and images, and support for RST
+   and PGTEI
+3. Fixes to many long-standing bugs
+
+
+### Major New Features
+
+All the HTML checks can be run with a  single click, and the output is
+clickable in most cases. Second, HTML and CSS  validation can now be done on
+your own computer (and PGTEI as well) and there are checks for unused CSS and
+image issues (using the pphtml and ppimage scripts). Third, there is now an
+option to view text and images side-by-side without having to click on "See
+Image" for each page. For instance you move forward or back one page for both
+text and image with the "<" and ">" buttons on the status bar. Also, the Auto
+Show Images option lets you see the image for instance for the page you are
+spellchecking or for each search hit.
+
+### Other New Features
+
+A "View in Browser" and Hyperlink page numbers buttons on the HTML palette,
+tearoff of the Unicode menu, listing small caps in the Word Frequency popup,
+automatic checking for updates (which can be turned off), horizontal rules as
+css, an option if nothing is found to return to the starting point, better
+ability to find executables automatically, GutWrench scanno files are
+included, a warning to use human readable filenames, option to include
+goodwords in spellcheck project dictionary, a text processing menu to ease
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -258,5 +258,100 @@ and jeebies directories without the .exe files to the guiguts-n.n.n.zip
...
+if longer than 8 characters. Reversed order of "Title" and "Caption" in HTML
+image popup. Word frequency count is run before any spell check. Toolbar font
+is no longer italic for readability. Default poetry left rewrap margin set to
+4. Fixed case sensitivity of searches from Word Frequency Popup. Made "Stay on
+Top"  preference apply to most popups except Word Frequency. Fixed double
+click search on Word Frequency popup to work for strings with nonalphanumeric
+characters (',-,--) while searching from Character Cnts does not do a whole
+word search. Word search from Word Frequency popup works if the word contains
+an apostrophe. Allow search from Word Frequency popup for expressions with
+regex metacharacters such as "\". Made default sort order for the word
```
srjfoo, line 340: I think the backslash needs to be escaped. It doesn't show as  [displayed](https://github.com/DistributedProofreaders/guiguts/blob/f15eacaf8d191591ad84959968564e1d7d4484b8/CHANGELOG.md).

CHANGELOG.md
```diff
@@ -258,5 +258,100 @@ and jeebies directories without the .exe files to the guiguts-n.n.n.zip
 file.
 
 ## Version 1.0.0
```
srjfoo, line 260: I realize that none of the other versions are dated in the changelog, but should we do that from now on? The wiki page for 1.0.0 was created 2011-12-27, and unless we include that here, that information will be lost.

CHANGELOG.md
```diff
@@ -258,5 +258,100 @@ and jeebies directories without the .exe files to the guiguts-n.n.n.zip
 file.
 
 ## Version 1.0.0
-See https://www.pgdp.net/wiki/Guiguts_new_features_and_bug_fixes for a
-description of new features and bug fixes.
\ No newline at end of file
+Relative to version 0.2.10, the main changes in version 1.0 are
+
+1. One click installation on Windows and Macintosh/OSX computers with no need
+   to install perl separately (see guiguts-win and guiguts-mac zip files)
```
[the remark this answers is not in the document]
cpeel: I left it out because it's not relevant to 1.0 and I have no idea where in the 1.x sequence it matters -- partially because it is MacOSX version dependent. I'm not worried about losing this text.

CHANGELOG.md
cpeel: Thanks for catching this. I will fix this and might tweak a few other places where the quotes would be better as Markdown instead.

CHANGELOG.md
cpeel: No. Release dates are displayed on the https://github.com/DistributedProofreaders/guiguts/releases page instead and I'm not worried about losing when 1.0.0 was released.

cpeel:
> Probably a separate thing, in the previous release, the Windows users got a .txt file in place of .md files because there's no built-in app for reading .md files. Would that be possible/desirable for the various .md files we have?

Windows users can open .md files in any text editor -- including Guiguts. As part of the Windows build we could rename the `.md` files to `.txt` files but the contents of the files would still point to `.md` files which would likely generate confusion. I think people will just need to adapt.

CHANGELOG.md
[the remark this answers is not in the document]
srjfoo: That's fair.

CHANGELOG.md
srjfoo: Also fair. :)

windymilla:
> Windows users can open .md files in any text editor -- including Guiguts. As part of the Windows build we could rename the `.md` files to `.txt` files but the contents of the files would still point to `.md` files which would likely generate confusion. I think people will just need to adapt.
OK - happy to merge then

PR #62, opened 2020-06-24
Add "Always treat as UTF-8" option

windymilla:
Previously, GG searched the file to see if there were
characters above \xFF in order to decide whether to
save, etc., as utf8.
Flag set via preferences now forces treating as utf8.
Default is on.

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -183,6 +183,7 @@ our $twowordsinhyphencheck = 0;
 our $unicodemenusplit   = 2; # 2 or 3
 our $utffontname        = 'Courier New';
 our $utffontsize        = 14;
+our $utf8save           = 1;	# True = always save utf8, false = only if unicode characters in file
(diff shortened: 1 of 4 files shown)


cpeel:
`currentfileisunicode()` is used for more than file saving though, it appears to be used for some file checks. What happens if the file loaded is Latin-1 and we treat it as UTF-8 for some of these checks?

windymilla:
It is used for some file checks, but GG writes a temporary file to be used for the check. The output encoding of that tempfile depends on `currentfileisunicode()` too (Errorcheck.pm line 392) so will be UTF-8 when written out. Then when the results of the check are read in, they will also be treated as UTF-8. I think the logic works, and the tools seem to work.
However, the good news is that while re-confirming that this morning, I realised I hadn't saved the flag in setting.rc and spotted a Capitalisation typo in the menu entry which are fixed in the new commit. Thanks.

PR #63, opened 2020-06-25
Preserve file permissions on saving

windymilla:
When saving an existing file (or bin file) remember its
file permissions before deleting it, so they can be copied
to the new version of the file. To fix reported bug that GG
was removing group/other access under Linux every save.

Also fixed incorrect use of decimal constant intended
to be octal constant.

Tested as best I could, given limited ability to set permissions on my Windows machine - "readonly"ness was preserved.


cpeel:
Confirmed this works as-designed on my Linux VM.

PR #64, opened 2020-06-25
Require "[Footnote" to mark start of footnote

windymilla:
At just one point, "[foot" was accepted to start a
footnote, which messed up footnote processing.
Require "[Footnote" (case insensitive) instead.

Bug reported as footnotes going wrong when
text contained "at the [foot of the] south-east hills"


windymilla:
Thanks :)

PR #66, opened 2020-06-26
Use em or % for Auto-Illustration widths

windymilla:
User can choose to specify width in em or %
When user loads image or changes between em & %
GG calculates and displays best guess at suitable width.
If ems, width is pixels/16
If %, width is small enough % that whole image should
fit horizontally & vertically on a 4:3 screen.

Minor fixes added after testing

### DIFF
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -1796,104 +1796,102 @@ sub htmlimage {
 sub htmlimageok {
 	my $textwindow = shift;
 	my $name = $::lglobal{imgname}->get;
-	if ($name) {
-		my $widthcn = my $width = $::lglobal{widthent}->get;
-		$widthcn =~ s/\./_/; # Convert decimal point to underscore for classname
-		return unless $name;
-		my ( $fname, $extension );
-		( $fname, $::globalimagepath, $extension ) =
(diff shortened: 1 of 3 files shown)


src/lib/Guiguts/HTMLConvert.pm
```diff
@@ -1901,6 +1793,152 @@ sub htmlimage {
 	&thumbnailbrowse();
 }
 
+sub htmlimageok {
+	my $textwindow = shift;
+	my $name = $::lglobal{imgname}->get;
+	if ($name) {
+		my $widthcn = my $width = $::lglobal{widthent}->get;
+		$widthcn =~ s/\./_/; # Convert decimal point to underscore for classname
+		return unless $name;
```
cpeel, line 1802: 1. Thanks for making this a real function.
2. I know this is carry-over, but I don't understand why this line is here. We wouldn't be in this function unless `$name` was something. Come to think of if, instead of having this entire function's contents inside an if block, perhaps we just need to move this `return unless $name` up a few lines, remove the if block, and outdenting everything.

windymilla:
Suggested edit made, outdenting most of sub.
Also realised it was a bit intrusive to have width:100% apply to all img elements, so extracted it to CSS class, and added the class when writing img elements from HTML generation code.
Should be ready for merging now.

PR #67, opened 2020-06-27
Upgrade and usage from git repo instructions

cpeel:
This MR has two commits:
* Upgrade instructions
* A section in the install file for how to use GG from a git checkout

I am very open to any suggestions or improvements to these!


windymilla:
In UPGRADE.md, I don't think we want the user to copy across setting.rc. It contains things like urls that have been updated for phpBB3, which they would then override with the old version. Also, I think it could mess up the automated finding of tools and changes to external operations.

I also think it would be better to expand on step 5, and suggest a slightly different compare/edit process.

Here's a possible draft

4. The files in the `data` directory contain words used in HTML generation like Footnote
   and Page, so they can be customised for different languages. If you have changed these in
   the old release, copy the relevant ones to the new release. This will not apply to most users.
5. The file `headerdefault.txt` has had important changes made to it in this release.
   When you first run Guiguts, it copies this file to one named `header.txt`. If you did not
   make any manual edits to `header.txt` in the old directory, proceed to step 6. If you are
   unsure whether you did make changes or what they were, compare `header.txt` with
   `headerdefault.txt` in the old directory, making note of any differences. Run the newly
   installed Guiguts program then immediately exit. Edit `header.txt` in the new directory and
   make the changes you want to carry over from the old release.
6. You can now run Guiguts from the new installation directory.

Similarly in the instructions on usage from a git repo, I'm slightly concerned that "advanced" users might just copy their header.txt and setting.rc, negating some benefits of the new release. Perhaps something like

3. (optional) Copy any manual edits you had made to Guiguts data files (`header.txt`, etc) into
   the new versions of the files in `src/` if you want to retain them.

cpeel:
All good points - and thank you for the suggested updates. I've lifted most of your text with a few other changes to talk about `settings.rc` a bit. Additional suggestions and changes welcome!

UPGRADE.md
```diff
@@ -0,0 +1,40 @@
...
+   (eg: `c:\guiguts-old`).
+3. Follow the instructions in [INSTALL.md](INSTALL.md) to install Guiguts
+   for your operating system, including any helper applications.
+4. Review and possibly copy over changes made to Guiguts files.
+   * The `settings.rc` file stores Guiguts configurations. This file is
+     automatically generated by Guiguts if it doesn't exist. Newer versions of
+     Guiguts will include updated defaults, such as DP URLs and bundled tool
+     locations. It is recommended that you **do not copy** this file from your
+     old version but let the new Guiguts create the file afresh. This may
+     require you to copy over updates for External Programs, etc.
```
windymilla, line 24: Instead of "This may require you to copy over updates for External Programs, etc.", how about, "You may then have some Preferences you want to change within Guiguts which will be saved in this file." My reasoning is that we don't want non-technical users trying to edit settings.rc - they should use the Preferences menu. Experts will already know they can copy/paste chunks from their old settings.rc to their new.

windymilla: Then I'm happy with all this - thanks.

INSTALL.md
```diff
@@ -195,6 +194,28 @@ For other platforms, you will need to install Perl and the necessary
...
 ```
 
+## Using Guiguts from a Git checkout
+
+_This section is for advanced users who want to run the latest in-development
+version of Guiguts and are comfortable with git._
+
+You can run Guiguts directly from the git repo with a few small changes.
+
+1. Clone the repo somewhere. You'll find the main Guiguts files (`guigut.pl`,
```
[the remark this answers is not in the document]
cpeel: It can't hurt! Added.

UPGRADE.md
cpeel: Good suggestion - fixed!

cpeel:
Sounds like Nigel is good with this. What are your thoughts, Sharon?

srjfoo:
> Sounds like Nigel is good with this. What are your thoughts, Sharon?

Looks good to me!

PR #69, opened 2020-06-30
Ensure HTML Internal Link finds all ids

windymilla:
When user creates an internal link from the HTML markup
dialog, only link targets defined with <a> elements
were listed. New code lists all link targets defined with "name"
or "id".

Also, order of 4 buttons (anchor, image, internal link,
external link) was "random" due to being extracted from
hash by keys. This is documented as depending on Perl
version and other unpredictables. Buttons were ordered
differently under Strawberry Perl compared to bundled
Perl version. Buttons now stored in ordered array.


cpeel:
Nice find & fix for the button ordering!

PR #70, opened 2020-07-02
Use Ctrl-b in place of Ctrl-n for search count

windymilla:
Since Ctrl-n might be wanted later for "New", use Ctrl-b
instead to count number of search matches.

Ctrl-b executes a default action of "move back one space"
in Text widgets. It was therefore necessary to disable that
default binding and bind the new action.

Ctrl-f was previously executing its default action (forward
one space) if pressed in the search/replace fields, as well as
its search function. Now fixed by new code in binding sub.
Also, no longer need for deleting Ctrl-g from text field when
that is pressed, nor removing second line of text if Enter key
(to search) was pressed in middle of text string.

Handling of multiline search string, e.g. by selecting 2 lines
of text in main window and pressing Ctrl-f was inconsistent.
Some paths through code attempted to remove all but the
first line, sometimes successfully. Some paths did not remove
additional lines. For now, only the first line selected is
included in the search box, though (as previously) user can
copy and paste multiple lines of text into the search or
replace boxes deliberately, but maybe with unexpected
results.

Consider replacing the Text boxes which allow multilines
with Entry boxes which are single line, for future release.

Notes also made in KeyBindings on order of callbacks in
Perl/Tk and how to override behaviour for a specific
widget if required in future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 - HTML title wording puts book title first
 
 ### Improved Search & Replace functionality
-- a Count button (Ctrl+N) counts how many times the current search
+- a Count button (`ctrl+b`) counts how many times the current search
   settings would find a match
 - number of replacement terms can be changed by the user
 - search & replace preserves the position of page markers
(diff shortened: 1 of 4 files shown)


cpeel:
Don't forget to update the changelog for the Ctrl+N -> Ctrl+B.

windymilla:
Thanks for the reminder - should be OK now

windymilla:
Thanks

PR #74, opened 2020-07-05
Remove requirement for unicore/Blocks.pl

windymilla:
Some versions of Perl had list of Unicode blocks in
Blocks.pl. Not available in all versions so the
Unicode Character Search feature was failing.

Switch to existing list of blocks already used to
generate the Unicode menu. This does not contain
every block, just those likely to be needed in GG.

Also, get list of all Unicode character names when
dialog is popped rather than for every search. Add
a few comments to code logic.


cpeel:
This works great on Mac too. 👍

windymilla:
Thanks

PR #75, opened 2020-07-05
Fix reverse search start point error

windymilla:
Reverse searches with a new string always started from
the end of the file, rather than the current insert point,
unlike forward searches.

Fixed, simplified logic slightly and commented.


cpeel:
Very nice!

windymilla:
Thanks

PR #76, opened 2020-07-05
Document using other Perls on Windows

windymilla:
Although we do not want to encourage use of old versions
of Perl, I think these comments are worth including in
the installation instructions.

After experience installing under Windows with other
Perl distributions and attempting to use the old
bundled Perl, add further suggestions on how to get
GG working in these circumstances

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -77,8 +77,9 @@ The bundled perl interpreter included with Guiguts 1.0.25 may also work but
 is no longer maintained. The bundled perl includes the required modules
 used in 1.0.25 which may not be the full set needed by later versions.
 
-To temporarily use the old perl, copy the `perl` directory from the top
-level of your 1.0.25 installation to the top level of your new installation,
+While not officially supported, if you wish to use the perl included with
+Guiguts 1.0.25, copy the `perl` directory from the top level directory of
+your 1.0.25 installation to the top level of your new installation,
 e.g. copy the `perl` directory from `C:\dp\guiguts\guiguts-win-1.0.25` to
 `C:\dp\guiguts\guiguts-win-1.1.0`. You also need to copy the `Tk` directory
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -72,6 +77,13 @@ The bundled perl interpreter included with Guiguts 1.0.25 may also work but
 is no longer maintained. The bundled perl includes the required modules
 used in 1.0.25 which may not be the full set needed by later versions.
 
+To temporarily use the old perl, copy the `perl` directory from the top
```
cpeel, line 80: I'm wary of including this paragraph. I worry it implies that running with this perl is supported which would obligate us to test it.

windymilla:
I wondered if the word "temporarily" covered that. I would like to mention
what can be done, but do you have an idea of how to strengthen the wording,
making it clearer it's unsupported, and just a temporary situation until
they get a modern Perl installed.
Or are you saying "definitely not"?

cpeel:
Why don't we just be explicit then? Perhaps?
```
While not officially unsupported, if you wish to use the perl included
with Guiguts 1.0.25, copy the `perl` directory from the top
...
```

windymilla:
Ok. Will edit that tomorrow. Thanks

srjfoo:
> Why don't we just be explicit then? Perhaps?
> 
> ```
> While not officially unsupported, if you wish to use the perl included
> with Guiguts 1.0.25, copy the `perl` directory from the top
> ...
> ```

Did you mean "While not officially supported"?

Should it also include `and cannot be guaranteed to work as expected` following that statement?

windymilla:
Edit made - does that look OK? If so, I'm happy for it to be merged. Thanks

cpeel:
Thanks, that sounds great. Merging it in now.

Issue #77, opened 2020-07-05
Tk::Error: Failed to AUTOLOAD 'Tk::Frame::yview' when using the mouse wheel in Unicode Character Search


cpeel:
I am getting the following error when I use the mouse wheel to scroll in the Unicode Character Search dialog:
```
Tk::Error: Failed to AUTOLOAD 'Tk::Frame::yview' at /Users/cpeel/git/guiguts/src/lib/Guiguts/Utilities.pm line 577.
 Carp::croak at /usr/local/Cellar/perl/5.30.3/lib/perl5/5.30.3/Carp.pm line 291
 Tk::Widget::__ANON__ at /usr/local/Cellar/perl/5.30.3/lib/perl5/site_perl/5.30.3/darwin-thread-multi-2level/Tk/Widget.pm line 347
 Guiguts::Utilities::__ANON__ at /Users/cpeel/git/guiguts/src/lib/Guiguts/Utilities.pm line 577
 <Button-5>
 (command bound to event)
```

To reproduce this:
1. Open Guiguts
2. Open the Unicode Character Search dialog (Tools -> Character Tools -> UTF Character Search)
3. Search for a character (eg: `latin b`)
4. Scroll the dialog using the touchpad

Env:
* Platform: macOS 10.15.5
* Perl: v5.30.3 via homebrew
* Guiguts: pre-release 1.1.0 (commit f3f37c3f)

I'm able to scroll with the mouse wheel in the main window just fine. It doesn't appear(?) to use the BindMouseWheel function so perhaps this function is no longer needed?

Aside: The "UTF Character Search" menu item should probably be renamed "Unicode Character Search". The "UTF Character Entry" menu doesn't match the dialog name that opens up -- both of which should probably be "Unicode Character Entry".

windymilla:
I can't see the bug as it is occurring in platform-specific (non-Windows) code. However, I do see the scroll wheel doesn't work like it does in the main window (and in the gutcheck window, which seems to also call BindMouseWheel, but doesn't actually need it to work).

windymilla:
I see the same behaviour (scroll wheel not working) in 1.0.25. It would be interesting to know if the Tk error was also happening in 1.0.25. I suspect so, but maybe not - maybe due to a change of Perl interpreter/version. This dialog is probably not (heavily) used by (m)any users.

cpeel:
[...]

PR #78, opened 2020-07-05
Add Basic Latin and Latin-1 Supp. to UCS

windymilla:
I haven't edited Menu entries or dialog titles to make them match (as described in https://github.com/DistributedProofreaders/guiguts/issues/77). That seems like a separate issue to this bug fix.

Previous change from using Blocks.pl to using
Unicode menu hash meant that these two blocks
were omitted. Add them manually at start of sub.


windymilla:
Thanks

PR #88, opened 2020-07-08
Cause `make` or `make all` to build all versions

windymilla:
Each OS target zip file needs to have a clean folder
and copies of generic files before the OS-specific
make. To force this, use recursive make to make each
OS's zip independently.

Only other expected difference is that the generic zip
will be called guiguts-generic-$(VERSION).zip rather
than guiguts-$(VERSION).zip

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -116,8 +116,8 @@ Guiguts requires.
 
 ### Extracting Guiguts
 
-Unzip `guiguts-n.n.n.zip` to some location on your computer (double click the
-zip file in Finder or run `unzip guiguts-n.n.n.zip` on the command line). You
+Unzip `guiguts-mac-n.n.n.zip` to some location on your computer (double click the
+zip file in Finder or run `unzip guiguts-mac-n.n.n.zip` on the command line). You
 can move the `guiguts` directory it creates to anywhere you want. A common place
 for this is your home directory.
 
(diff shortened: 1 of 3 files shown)


cpeel:
This LGTM, but if we are changing to `guiguts-generic-$(VERSION).zip` from `guiguts-$(VERSION).zip` we need to update `INSTALL.md`. Please also update `RELEASING.md` with the instructions that we can just `make all`.

windymilla:
Thanks
Is generic the best name for this version? Rather than linux or something
similar?

windymilla:
Looking at RELEASING.md it says
   Be certain to keep the format
   `guiguts-1.0.0.zip` (version number will vary) in order for
   the update mechanism to work correctly.
Is changing the name to guiguts-generic-1.0.0.zip (or similar) a problem for that?

cpeel:
> Thanks Is generic the best name for this version? Rather than linux or something similar?

That's a fair question as we inherited "generic" from the prior code. For us it is mostly about Linux, but there isn't anything linux-specific about it. I think generic is as good as anything unless you have something better.

windymilla:
I'm OK with that, so if you think the changes I've made are OK, I'm happy
for it to be merged.
Thanks
Nigel

cpeel:
> Looking at RELEASING.md it says
> Be certain to keep the format
> `guiguts-1.0.0.zip` (version number will vary) in order for
> the update mechanism to work correctly.
> Is changing the name to guiguts-generic-1.0.0.zip (or similar) a problem for that?

Sorry, missed this question earlier. The way that the online version check happens as long as it can find `(\d+)\.(\d+)\.(\d+)\.zip` in the releases page it should be fine. Granted, that's never been actually tested with more than one version but...

PR #103, opened 2020-07-11
Fix online (remote) HTML validation

windymilla:
Code was printing returned results from remote
HTML validation with "W3C" at the start, but then
removing all "W3C" lines. Now prints matching
format of local validator so lines are not removed.

Also fixed to include warnings as well as errors as
these should be corrected by PPer.

If validation does not occur, report the actual
error rather than 'Could not contact remote
validator'.

Online validation requires two perl modules in
order to work, so added these to the
optional/recommended list during installation.

Finally, switching between local and remote
validation in the Prefs menu did not rebuild the
menus, so the label for the previous validation
type still showed in the HTML menu.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -239,12 +239,15 @@ Guiguts requires the following Perl modules to be installed via CPAN:
 * Tk
 * Tk::ToolBar
 
-The following modules are optional but recommended:
+The following modules are optional but recommended in order to provide
+all available functionality:
 
-* Text::LevenshteinXS
-* File::Which
-* Image::Size
-* LWP::Protocol::https
+* Text::LevenshteinXS               # Improved harmonics code
+* File::Which                       # Find tools on the system path
+* Image::Size                       # Report size of HTML images
+* LWP::Protocol::https              # Check for updates
+* WebService::Validator::HTML::W3C  # Online HTML validation
+* XML::XPath                        # Online HTML validation
 
 The required Perl modules can be installed with the included helper script:
 ```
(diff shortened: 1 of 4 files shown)


INSTALL.md
```diff
@@ -239,12 +239,15 @@ Guiguts requires the following Perl modules to be installed via CPAN:
 * Tk
 * Tk::ToolBar
 
-The following modules are optional but recommended:
-
-* Text::LevenshteinXS
-* File::Which
-* Image::Size
-* LWP::Protocol::https
+The following modules are optional but recommended in order to provide
```
cpeel, line 242: This list will not look good rendered: https://github.com/windymilla/guiguts/blob/validate/INSTALL.md#perl-modules

src/install_cpan_modules.pl
```diff
@@ -10,6 +10,9 @@
     "File::Which",
     "Image::Size",
     "LWP::Protocol::https",  # needed for update checking
+	# Required for remote HTML validation
```
cpeel, line 13: The other lines are using spaces and this is using tabs. Yes, this is inconsistent with the other perl files but let's at least be consistent within this one.

src/install_cpan_modules.pl
```diff
@@ -10,6 +10,9 @@
     "File::Which",
     "Image::Size",
     "LWP::Protocol::https",  # needed for update checking
```
cpeel, line 12: Let's move this comment to the line above it so it matches the lines you just added.

src/install_cpan_modules.pl
windymilla: Once this release is out of the way, we'll convert all tabs to spaces in the main source, and I can set NP++ to use spaces instead of tabs when it auto-indents to the same as the line before.

windymilla:
Hopefully that's all OK to merge now - thanks

PR #107, opened 2020-07-13
Replace DPCustomMono2 with DP Sans Mono

cpeel:
Replace DPCustomMono2 with information on DP Sans Mono. Because the
font is a living font that will be updated over time, point users
to the forum topic they can subscribe to and gives information on
where to download the latest version from.


tools/DPSansMono/README.md
```diff
@@ -0,0 +1,10 @@
+# DP Sans Mono
+
+[DP Sans Mono](https://www.pgdp.net/wiki/DP_Sans_Mono) is a custom proofreading
+font designed to distinguish between common-looking characters. It is the
```
srjfoo, line 4: `common-looking` doesn't sound quite right. Perhaps `common look-alike`? I don't know if it's useful to give an example like `Il1`, or `O0`. (And yes, I realize that the wording comes from the DPCM readme.)

I don't know if it's appropriate to mention or not, but another big advantage is that it also exaggerates punctuation and diacriticals, which is actually more useful when it comes to non-Latin scripts like Greek.

Also, it doesn't distinguish between look-alike characters that are in different scripts, like ABP in Basic Latin, and the upper case alpha, beta and rho in Greek.

cpeel: I've updated the wording a bit. 👍

PR #124, opened 2020-07-17
Fix pngs dir error after import prep files

windymilla:
Behaviour of File::Spec::Functions::catfile has
changed between Perl bundled with 1.0.25 and
latest Strawberry Perl. It now strips a trailing
slash rather than ensuring one is there.

This is the only place catfile is called without a
filename, so the only place where this issue can
arise. Append OS-specific slash to path.

### DIFF
--- a/src/lib/Guiguts/FileMenu.pm
+++ b/src/lib/Guiguts/FileMenu.pm
@@ -199,6 +199,9 @@ sub file_import_preptext {
 	my $tmppath = $::globallastpath;
 	$tmppath =~ s|[^/\\]*[/\\]$||; # go one dir level up
 	$tmppath = ::catfile( $tmppath, $::defaultpngspath, '');
+	# ensure trailing slash - recent versions of catfile remove it
+	my $slash = $::OS_WIN ? "\\" : "/";
+	$tmppath .= $slash unless substr( $tmppath, -1 ) eq $slash;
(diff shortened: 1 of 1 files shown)


src/lib/Guiguts/FileMenu.pm
```diff
@@ -199,6 +199,9 @@ sub file_import_preptext {
 	my $tmppath = $::globallastpath;
 	$tmppath =~ s|[^/\\]*[/\\]$||; # go one dir level up
 	$tmppath = ::catfile( $tmppath, $::defaultpngspath, '');
+	# ensure trailing slash - recent versions of catfile remove it
```
cpeel, line 202: Note that the line above tries to ask catfile to append a trailing slash with the use of `''` at the end there. It sounds like catfile changed and this no longer enforces the slash. I would remove the `, ''` on the line above and the ` - recent versions of catfile remove it` here (because that comment isn't going to be useful to future developers who won't know what catfile use to do and only care that the slash is appended now).

windymilla: That's not strictly true. `defaultpngspath` has a trailing slash already, and `catfile` is removing it - it's not being asked to append one. But anyway, I agree the comment is no use so have removed it, and also switched to using `catdir` which I discovered while investigating how `catfile` has changed. Should be ready to merge, thanks

Issue #86, opened 2020-07-07
Add a CSS3 validation check to the HTML Menu


windymilla:
Currently `HTML | CSS Validator` performs a CSS3 validation check.
Rename this to `CSS 3 Validator` and add `CSS 2.1 Validator`

PPers should validate at 2.1 and only validate at level 3 if the file contains specific permitted CSS3.

windymilla:
Decided to make CSS3 the default, but provide a toggle to allow validation at level 2.1 in the preferences menu. Should be addressed by #125

PR #125, opened 2020-07-18
Make CSS validation default to level 3

windymilla:
Local CSS validation previously defaulted to Level
3+SVG.
Now defaults to Level 3, though user can select
Level 2.1 via Prefs menu for this run of GG.

Also dealt with single line output by validator to
stderr which showed the options used for this
validation. When redirecting stdout to a file in
order to capture output from an external
program, you can now also redirect stderr, either
to the same file or a different one. Only used
for CSS validation at the moment - errors from
other tools that output to stderr will appear in
the command window.


windymilla:
This should address #86

PR #135, opened 2020-07-19
Restrict Count search function to selection

### LINKED ISSUE #108: "Count" on Search & Replace dialog should only count what's in a selection, if there is one
"Count" always counts the entire file, even if there's a selection. When there is a selection, it should count only what's within it; when there's no selection, it should continue to count the entire file.

windymilla:
If just some text is selected, then only count
occurrences in the selection, rather than the whole
file.

Fixes #108.


windymilla:
Should be OK to merge now.

PR #138, opened 2020-07-21
Adjust autogenerated image width handling

windymilla:
1. If user specifies percentage width, don't allow
image to exceed its natural width, i.e. make the
maximum width the width dimension of the image
file.
2. Allow user to specify that the set percentage
width is to be overridden to 100% on handheld
devices (i.e. through ebookmaker generation)
Note that all images with that percentage width
either will or will not have override. Class CSS is
rewritten to ensure user's latest choice is used.


cpeel:
I'm not gonna like - I'm lost in this change but trust that you've tested this directly.

PR #155, opened 2020-07-28
Update Windows installation instructions

windymilla:
If user has not ended up with `.pl` files
associated with their installed Perl, they can't
double-click `install_cpan_modules.pl` to run it.

Instead, tell them how to right click and set up
the file extension association.


windymilla:
Thanks for the quick response

PR #176, opened 2020-08-05
Standardize and enforce a perl coding style

cpeel:
This standardizes on a single perl coding style using perltidy. It contains a handy function to format and test the files (`format_files.sh`) and uses TravisCI to confirm the formatting as part of MRs.

The initial change is a beast since it touches virtually every line as we move from tabs to spaces -- and since we do that anyway we move from Windows to Unix file endings. The best way to review the file formatting changes is to [browse the branch](https://github.com/cpeel/guiguts/tree/perltidy) and see how the perl files are formatted. I'm very amenable to `.perltidyrc` adjustments and now is the best time to make them so we don't have to touch everything yet again.

@windymilla, please test out the tooling and make sure it's easy to run on Windows as well. I'm hoping it's easy to run `./format_files.sh` in GitBash and have it format files before you commit them. The last thing I want to do is make development difficult for the head developer!


windymilla:
I can't get perltidy to work inside the Git Bash. 
In a command window, I can type `perltidy --version`, with the expected results.
However in a Git Bash window, I get
```
Can't locate Perl/Tidy.pm in @INC (you may need to install the Perl::Tidy module) (@INC contains: /usr/lib/perl5/site_perl /usr/share/perl5/site_perl /usr/lib/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib/perl5/core_perl /usr/share/perl5/core_perl) at /c/Strawberry/perl/site/bin/perltidy line 4.
BEGIN failed--compilation aborted at /c/Strawberry/perl/site/bin/perltidy line 4.
```
Seems like it's using the perltidy from Strawberry perl (unsurprising, since I used Strawberry Perl cpanm to install it). 
In my Git Bash, `type perl` gives `/usr/bin/perl`, but `type cpanm` gives `/c/Strawberry/perl/bin/cpanm`.

cpeel:
> Seems like it's using the perltidy from Strawberry perl (unsurprising, since I used Strawberry Perl cpanm to install it).
> In my Git Bash, `type perl` gives `/usr/bin/perl`, but `type cpanm` gives `/c/Strawberry/perl/bin/cpanm`.

You might try updating your GitBash path so that Strawberry Perl is at the front instead of the end. At a GitBash prompt:
```
export PATH=/c/Strawberry/perl/bin:$PATH
type perl
```
If running perltidy with that works, you can [add the export to a `.bashrc`](https://stackoverflow.com/questions/10217154/bash-startup-script-in-windows).

windymilla:
Changing the path to put Strawberry perl earlier worked fine to get format_files.sh working under Git Bash, thanks

windymilla:
However, there is a line endings issue. When I get files from Github, they end up with Windows endings, because the files in your branch that I cloned all have Windows endings. Of course, then when I perltidy a file, it converts all to Unix endings which means it finds diffs for every file. Do we need a '.gitattributes' file to force unix line endings on non-binaries?

windymilla:
If you run format_files.sh from a directory other than the top level guiguts directory, it doesn't pick up the .perltidyrc
I discovered this when I went to src/lib/Guiguts and used `../../../format_files.sh ASCIITables.pm` to try tidying one specific file.
Changing the perltidy line in format_files.sh to 
`    perltidy --profile=.../.perltidyrc $file`
would fix this. Beginning the .perltidyrc path with three dots apparently means search from the current directory upwards to find a .perltidyrc file, intended to support projects having ".perltidyrc in their root directories"

windymilla:
I'm happy with all the formatting, i.e. settings in `.perltidyrc`.

Also, I can just double click `format_files.sh` in Windows and it runs it in a shell - another reason to add the --profile argument mentioned previously. 

It takes under 40 seconds on my computer, so will be very manageable to run prior to committing - now all I need to do is to remember to do it!

cpeel:
> Do we need a '.gitattributes' file to force unix line endings on non-binaries?

I futzed with this for a while and it became a real mess. I've changed perltidy to use Windows line endings again -- is that better?

> Changing the perltidy line in format_files.sh to
> ` perltidy --profile=.../.perltidyrc $file`
> would fix this.

Ohhh, thanks for finding this! I've fixed `format_files.sh`.

windymilla:
> > Do we need a '.gitattributes' file to force unix line endings on non-binaries?
> 
> I futzed with this for a while and it became a real mess. I've changed perltidy to use Windows line endings again -- is that better?

Yes - that does work for me, but won't be so good for a Linux user. I've tried `--preserve-line-endings` instead of `--output-line-ending=win` and that works fine for me too, and could be better for other users. 

> > Changing the perltidy line in format_files.sh to
> > ` perltidy --profile=.../.perltidyrc $file`
> > would fix this.
> 
> Ohhh, thanks for finding this! I've fixed `format_files.sh`.

I can't see that fix in the version I've got.

Also, after trying it out a bit more across the whole tree, I have a couple more suggestions for format_files.sh.
1. I would prefer if it didn't update the file even when no tidying was done. It means the dates for all Perl files get set to when you last ran the script. I sometimes find it useful to have my directory view sorted by date order and the files I've edited recently are at the top.
2. I would like a message when it does some tidying. At the moment, it outputs the filename before processing it, but you can't see from that output whether any tidying was done. I think that could be useful as it could alert you if some tidying happened when you didn't expect it.

In place of [line 27](https://github.com/cpeel/guiguts/blob/perltidy/format_files.sh#L27)
`        mv $file.tdy $file`
I put
```
        cmp -s $file $file.tdy
        CMP_RESULT=$?
        if [[ $CMP_RESULT -ne 0 ]]; then
            echo $file tidied
            mv $file.tdy $file
        else
            rm $file.tdy
        fi
```
which compares the original and tidied files. If they're different it outputs a message before moving the tidied version into place, but if not it silently deletes the tidied file. I know there is some overlap with the code for the --check case, but restructuring it didn't reduce that didn't seem to make it simpler as the diff output would need capturing, the messages are different, etc.

cpeel:
Not sure what happened to the `perltidy` command updates - maybe I rebased them out? I'll get that fixed. Your other updates sound good and I'm happy to get them in.

A question: would you rather the format_files.sh only output filenames if they are either tidied or don't match?

windymilla:
No, I'm happy for it to list them as it does them - as I said, it takes about 40 seconds, so it gives some feedback that something is happening. But if you think it would be better not to echo the names for the good files, I wouldn't mind. Thanks - happy for it to merge then

PR #177, opened 2020-08-06
fix #175 - CSS validator line numbers wrong

### LINKED ISSUE #175: Local CSS Validator error messages give line numbers that are 11 too low
Example: When there's a CSS error in line 70, the error report says it's in line 59.  Apparently, it isn't counting the heading lines above <style type="text/css">. 

I can supply a test file if needed, but think you can reproduce it yourself by introducing any error in the css of an existing file you already have.

windymilla:
New version of validator reports line numbers
starting at beginning of CSS style block rather
than start of file.

Also remove historical adjustment of line number which causes double-click on error to go to the
line following the error.

Fixes #175

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -257,10 +257,9 @@ sub errorcheckpop_up {
 						  or ( $thiserrorchecktype eq "Link Check" )
 						  or ( $thiserrorchecktype eq "pptxt" ) )
 				{
-					# Format line number, adjusting for start line of style block if CSS check
+					# Format line number, adjusting for tool's idea of start line
 					if ( $line =~ /Line : (\d+)/ ) {
-						my $lineno = $1;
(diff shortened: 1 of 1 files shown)


src/lib/Guiguts/ErrorCheck.pm
```diff
@@ -251,13 +257,17 @@ sub errorcheckpop_up {
 						  or ( $thiserrorchecktype eq "Link Check" )
 						  or ( $thiserrorchecktype eq "pptxt" ) )
 				{
-					$line =~ s/Line : (\d+)/line $1:1/;
+					# Format line number, adjusting for start line of style block if CSS check
+					if ( $line =~ /Line : (\d+)/ ) {
+						my $lineno = $1;
+						$lineno += $lineadjust if $thiserrorchecktype eq "W3C Validate CSS";
```
cpeel, line 263: Do you need the `$thiserrorchecktype` here? If it's not "W3C Validate CSS" $lineadjust will be zero so adding it to $lineno is just fine. Having this check in just one place paves the way for other adjustments to be added later without changing two places.

PR #188, opened 2020-08-19
Fix single-clicking in error windows

### LINKED ISSUE #123: PPHTML rings the bell; it shouldn't
When the PPHTML report is already on the screen and loses the focus (when you make a correction in the GG text window), then, when another item in the PPHTML list is clicked, it rings the bell.  It shouldn't, as the sound is both unnecessary and annoying.

Turning off the "bell" option in "Preferences" prevents this from happening, but there are other situations in GG where the bell is helpful, either because an unexpected error occurred or to call your attention to something else that's happened.

This is not new behavior: GG 1.0.25 did it, too.

Other utility lists that report errors, such as the Validator, do not ring the bell when an item in their lists is clicked.

windymilla:
In the bookloupe, jeebies and "others" error
windows, single-clicking on an error was
supposed to take you to the error in the main
window. It actually took you to the previously
selected error.

This was due to the event being triggered on
the mouse button being pressed, rather than
released, so the list of errors had not actually
selected the one clicked on yet.

This should also fix #123, that clicking on errors
after running pphtml sometimes caused the bell
(system beep). When the error does not have a
line number, GG searches for the text at the start
of the line (between plus and colon symbol). The
error can be clicked multiple times to search
through the occurrences, and then the bell is
sounded when there are no more search
matches, just as with a normal search. Because
of the "previous error" bug, if the user removed
an unused definition, e.g. "h6", then clicked on
their next error (unused h5), it would first do a
search for h6 (causing a bell) before the user 
clicked again on the h5 error to jump to that.

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -78,7 +78,7 @@ sub errorcheckpop_up {
 
     ::drag( $::lglobal{errorchecklistbox} );
     ::BindMouseWheel( $::lglobal{errorchecklistbox} );
-    $::lglobal{errorchecklistbox}->eventAdd( '<<view>>' => '<Button-1>', '<Return>' );
+    $::lglobal{errorchecklistbox}->eventAdd( '<<view>>' => '<ButtonRelease-1>', '<Return>' );
     $::lglobal{errorchecklistbox}->bind( '<<view>>', sub { errorcheckview(); } );
 
     # buttons 2 & 3 delete the clicked error and select the next error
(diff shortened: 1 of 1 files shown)


cpeel:
Note that for git commit messages while we want to keep the first line limited to 55-ish characters, the body of the commit message has 72 characters of space to play with 😁

windymilla:
I think since changing computers, the screen is narrower, so the field in
GitHub Desktop that I type the message into is narrower. On my old
computer, I had it sized so I knew if I filled the line it was OK. But on
this one, 70 characters is almost half the screen width. Since the other
half shows the diffs for the file, I want that side to be as wide as
possible of course! I'll resize the screen split when I want to type into
it and try to get a reasonable width for you :)

PR #204, opened 2020-08-23
Improve the Check for Updates dialog

### LINKED ISSUE #182: Improvements to the Check for Update dialog
**Describe the bug**
Since we're going to be updating GG more frequently (yay!) there are a few improvements we can make to the Check for Updates dialog.
* At least on macOS, the initial dialog is not sized big enough to show all of the contents of the screen (indeed, it's about 75 x 50 pixels smooshed together).
* The selection of radio buttons:
  * don't have a lot of context (perhaps a "Update frequency check" label?)
  * have no default value for my system with a fresh GG-created `settings.rc`
  * might be better aligned vertically instead of horizontally
* The "Update" button should take the user to https://github.com/DistributedProofreaders/guiguts/releases
* Dialog could use an OK button

**Environment**
* Guiguts version: 1.1.0
* OS and version: macOS 10.15.5
* Perl version: 5.30.3
* Perl distribution: homebrew

windymilla:
1. Store dialog position as with other dialogs
2. Redesign dialog, grouping information with frames and adding OK button
3. If version up to date, show message rather than Update buttons
4. Correct faulty default value for update type stored in setting.rc
5. Update button now goes to releases page
6. Correct faulty logic which wouldn't detect a major version change if minor
version remained the same and user was ignoring revisions.
7. Make it possible for user to disable monthly update checks (bug caused by
checking the wrong string return from the monthly question dialog)

Fixes #182

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -141,8 +141,8 @@ our $htmldiventry           = ' class="i2"';
 our $htmlspanentry          = ' class="i2"';
 our $highlightcolor         = '#a08dfc';
 our $history_size           = 20;
-our $ignoreversions         = "revision";      #ignore revisions by default but not major or minor versions
-our $ignoreversionnumber    = "";              #ignore a specific version
+our $ignoreversions         = "none";          # Don't ignore any updates by default
+our $ignoreversionnumber    = "";              # Ignore a specific version
 our $jeebiesmode            = 'p';
 our $lastversioncheck       = time();
 our $lastversionrun         = $VERSION;
--- a/src/lib/Guiguts/FileMenu.pm
+++ b/src/lib/Guiguts/FileMenu.pm
@@ -922,6 +922,9 @@ sub readsettings {
             }
         );
     }
+
+    # Correct previous bug which stored bad value in $::ignoreversions
+    $::ignoreversions = "revisions" if $::ignoreversions eq "revision";
 }
 
 ## Save setting.rc file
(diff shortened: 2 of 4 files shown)


cpeel:
The initial size of the update dialog is still problematic for me I'm afraid:
update-dialog-screenshot

windymilla:
The only thing I could see odd in the way this dialog had been constructed was that there was an initial Frame filling the whole dialog that had all the other widgets as children, rather than them being direct children of the dialog.
I removed that Frame, and for good measure made the dialog non-resizable. It shouldn't be resizable anyway.

If that still doesn't work, could you try this please (as a test, not a permanent fix):
Insert the following line just *before* the dialog gets created (at [Line 1629](https://github.com/DistributedProofreaders/guiguts/blob/890fb25525bb3f0b5f8e372952767a4954de95da/src/lib/Guiguts/Utilities.pm#L1629))
`$::geometryhash{versionbox} = '300x250+80+80';`
You may need to delete your setting.rc before/after trying that test fix.

Thanks

windymilla:
Once we have a solution, I'm happy to close this MR and do a fresh one with a single commit

cpeel:
Alas, your changes didn't alter the dialog behavior on the Mac (except make it not sizeable).

> If that still doesn't work, could you try this please (as a test, not a permanent fix):
> Insert the following line just _before_ the dialog gets created (at [Line 1629](https://github.com/DistributedProofreaders/guiguts/blob/890fb25525bb3f0b5f8e372952767a4954de95da/src/lib/Guiguts/Utilities.pm#L1629))
> `$::geometryhash{versionbox} = '300x250+80+80';`

This worked nicely and I consider a totally acceptable hack for Macs if you wanted to use it directly.

cpeel:
This update worked perfectly - thanks!

PR #207, opened 2020-08-27
Pull kindlegen binaries from GG 1.1.0

cpeel:
Amazon removed kindlegen downloads, pointing people to KindlePreviewer instead. To ensure that 1.1.x releases still include the kindlegen binary, pull form the GG 1.1.0 release.

This isn't ideal, but it's probably better than removing kindlegen binaries from a 1.1.x release.

### DIFF
--- a/tools/kindlegen/README.md
+++ b/tools/kindlegen/README.md
@@ -6,17 +6,21 @@ or EPUB. KindleGen converts this source content to a single file which supports
 both KF8 and Mobi formats enabling publishers to create great-looking books
 that work on all Kindle devices and apps.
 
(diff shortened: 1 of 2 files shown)


windymilla:
Yes, that works fine. I created a .mobi file which uses ebookmaker & kindlegen.

Are you bothered about updating the README?
It points to the page formerly known as kindlegen download
It also says we include the full kindlegen package, when with this change we are only including the exectuable.

cpeel:
Oh good catch. I've pushed a new commit that updates the `README.md`.

windymilla (approved):
Good to merge, thanks

PR #214, opened 2020-09-05
Tidy up use of \x7f characters

### LINKED ISSUE #90: \x7f characters need tidying up
Are \x7f characters needed in HTML convert or SearchReplaceMenu?
They are used in SelectionMenu to temporarily mark page boundaries
Should be a constant variable, not x7f everywhere.

windymilla:
Fixes #90
Replace \x7f with a named constant in the wrapping code.
Remove reference to \x7f in other files where it was not required.

### DIFF
--- a/src/lib/Guiguts/SelectionMenu.pm
+++ b/src/lib/Guiguts/SelectionMenu.pm
@@ -12,12 +12,15 @@ BEGIN {
 
 my $blockwraptypes = quotemeta '#iI';    # Blockquotes & index allow user to specify left/first/right margins
 
+# "Delete" character is inserted to flag mark positions during wrapping
+my $TEMPPAGEMARK = "\x7f";
+
 sub wrapper {
     my ( $leftmargin, $firstmargin, $rightmargin, $paragraph, $rwhyphenspace ) = @_;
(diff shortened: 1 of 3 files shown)


src/lib/Guiguts/SelectionMenu.pm
```diff
@@ -49,16 +52,16 @@ sub greedy_wrapper {
         if ( $word =~ /\/[$blockwraptypes]/ ) {    # Get margins for blockwrap types
             $firstmargin = $leftmargin = $::blocklmargin if $word =~ /\/#/;    # blockquote has its own default
                                                                                # Check if there are any parameters following blockwrap markup
-            if ( $word =~ /^\x7f*\/[$blockwraptypes]\x8A(\d+)/ ) {
+            if ( $word =~ /^$TEMPPAGEMARK*\/[$blockwraptypes]\x8A(\d+)/ ) {
```
cpeel, line 55: Should we do something similar with `\x8A` too?

windymilla:
No. The next step is to remove the old wrap algorithm. There are several
special characters to replace different things, but all are either old wrap
or are not needed, I believe.

Issue #213, opened 2020-09-05
Extra blank line added after block rewrap


windymilla:
**Describe the bug**
With "Greedy rewrap algorithm" only, when block-rewrapping, an additional blank line appears after rewrapping

**To Reproduce**
1. Open Guiguts
2. Load [rewrap.txt](https://github.com/DistributedProofreaders/guiguts/files/5177948/rewrap.txt)
3. Ensure Preferences-->Processing-->Use Greedy Traditional Rewrap Algorithm is checked
4. Note two blank lines after block
5. Use Tools-->Rewrap All
6. Note three blank lines after block

**Expected behavior**
At step 6, should be 2 blank lines after block

**Environment**
* Guiguts version: Version 1.1.0 onwards - caused by #38 

**Additional context**
Caused by #38.
Will be irrelevant if/after #212 is merged

windymilla:
Fixed by #215

PR #215, opened 2020-09-06
Remove old rewrap algorithm

### LINKED ISSUE #212: Remove old rewrap algorithm
New rewrap algorithm has been the default for many years. 

Remove the old one (called "greedy traditional" in the menus) from within the code, as well as the Prefs option to switch between algorithms.

windymilla:
New "Knuth" algorithm has been the default for years and gives more consistent
results, so no need to keep old one.

Removed replacement of brackets with special characters, since only used for old
rewrap style. Also removed special treatment of italic markup, since other HTML
markups not done, and anyway markup is removed from the text file well before
rewrapping.

Fixes #212 & #213

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -162,7 +162,6 @@ our $projectid             = q{};
 our $projectfileslocation  = '';
 our $recentfile_size       = 9;
 our $regexpentry           = q();
-our $rewrapalgo            = 2;
 our $rmargin               = 72;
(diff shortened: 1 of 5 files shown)


windymilla:
Embarrassing failure to test thoroughly enough - caught by the inbuilt test suite which I was just casually investigating to see if it was worth keeping. Apparently it is!
These 2 should stay as separate commits - if that's a problem I can create a new branch/MR
Thanks

PR #220, opened 2020-09-06
Get the test suite working again

### LINKED ISSUE #216: Get the test suite working again
Typing "guiguts.pl runtests" executes a series of tests and checks that the results match the expected ones.
Some tests now fail due to changes in the program.

windymilla:
1. Tidy up the code a bit, correct a few messages, etc.
2. Hide the users setting.rc while tests take place
3. Update tests based on recent code changes
4. Update baseline files based on recent code changes
5. Add HTML tests 3, 4 & 5 which were in the test folder, but not actually run.
6. Fix pphtml to output "possibly not defined" CSS classes in alphabetical order
rather than random hash order which changes between runs.

Fixes #216


cpeel:
Assuming runtest returns a non-zero exit code upon failure, we should add this to the TravisCI so that it is run on every MR to prevent regressions. If it doesn't return a non-zero exit code upon failure it should :D

windymilla:
Yes, it returns the number of tests that failed, i.e. zero if they all succeed.
It works by running the whole Guiguts program with the argument `runtests`, not just a small script, i.e. `perl guiguts.pl runtests`. I don't know if that means it is possible/practical to add to the TravisCI. Is it OK that it creates windows? Would it need to run install_cpan_modules beforehand?
The other question is whether the tests all run OK on Mac & Linux systems.

cpeel:
Installing modules isn't a problem but needing a UI is going to be the harder part and something which pushes it way outside this MR for scope of effort.

PR #222, opened 2020-09-09
Stop autosave clashing with save

### LINKED ISSUE #196: Tidy up Autosave feature and make it safer
1. `toggle_autosave` and `set_autosave` should be in the same file (Utilities.pm?)
2. Add `cancel_autosave` to cancel the repeat timers
3. Add `reset_autosave` to cancel and set the timers if autosave is enabled (to replace `::set_autosave() if $::autosave;`)
4. Make sure `Save` and `Save As` cancel autosave at the start of their subs, then reset at the end. At the moment, an autosave can be triggered while you're in the middle of a save, which might be unfortunate.

windymilla:
Fixes #196

It was previously possible for an autosave to happen during a Save, Save As, or
Save Copy As. Since some of these do some manipulating of the filenames, this was
potentially a dangerous operation, and because some save operations pop a dialog
for the user to respond to, they could take a long time, increasing the likelihood of
a clash.

This edit turns off autosave during the manual save processes. It also tidies the code,
putting all the main autosave code into the same module, and simplifies the calls to
the main subroutine.

Finally, part of the Save As and Save Copy As code that needed editing (to avoid a
premature return) turned out to be unnecessary, so was removed. For the record,
there was code to check for a Bin File Collision. Historically, filename.txt had a bin
file called filename.bin. Hence, if the HTML file was saved as filename.html, it would
have had the same bin file name. More recently (1.0.25 or before) the naming of the
bin file was changed to be filename.txt.bin and filename.html.bin, so there is no
possibility of a collision.

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -324,7 +324,7 @@ if (@ARGV) {
 } else {
     $lglobal{global_filename} = 'No File Loaded';
 }
-set_autosave() if $autosave;
+::reset_autosave();
 $textwindow->CallNextGUICallback;
 $top->repeat( 200, sub { _updatesel($textwindow) } );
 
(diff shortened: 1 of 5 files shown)


cpeel:
Nicely done - I appreciate the comments on why you use `{}`s and a local variable.

windymilla:
Thanks. It seems the safest way.

PR #228, opened 2020-09-15
Fix #227 - unknown color name `SystemButtonFace`

### LINKED ISSUE #227: Error starting GG: unknown color name "SystemButtonFace"
**Describe the bug**
When trying to start guiguts I'm getting the following error:
```
unknown color name "SystemButtonFace" at /Users/cpeel/git/guiguts/src/lib/Guiguts/Preferences.pm line 368.
```
and Guiguts does not start. This line was changed in commit `3068561c` so presumably a regression, possibly due to being on a Mac?

**To Reproduce**
1. Open Guiguts
2. See error in console

**Environment**
* Guiguts version: git hash `a8b4a2c5d`
* OS and version: MacOS 10.15.5
* Perl version: 5.30.3
* Perl distribution: homebrew

windymilla:
It turns out that color name is not available on all platforms.
Safer solution is to store the button's default color before it ever gets changed.

Fixes #227

### DIFF
--- a/src/lib/Guiguts/Preferences.pm
+++ b/src/lib/Guiguts/Preferences.pm
@@ -363,10 +363,14 @@ sub reset_autosave {
         );
     }
 
+    # Save icon's default background color so it can be restored when turning off autosave
+    $::lglobal{savetoolcolor} = $::lglobal{savetool}->cget('-background')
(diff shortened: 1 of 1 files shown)


cpeel:
Confirmed that this fixes my problem on MacOS - thanks!

Issue #178, opened 2020-08-06
Consolidate/simplify line number processing in ErrorCheck.pm


windymilla:
[Large if/else block](https://github.com/DistributedProofreaders/guiguts/blob/master/src/lib/Guiguts/ErrorCheck.pm#L224) in ErrorCheck.pm handles slightly different line/column formats as output from tools. 
Simplify to make all line/column lines have same format, e.g. `line 123:45` or just `123:45`?
Use common code after if block to allow for line adjustment (used by CSS validate), set mark for later searching, and push line onto `@errorchecklines`

PR #232, opened 2020-09-20
Make bl/gc and jeebies use the common error dialog

### LINKED ISSUE #157: Consolidate the several tool error windows
There are several different error windows, e.g. bookloupe/gutcheck, jeebies, etc. which have their own subroutines, dialogs and slightly different behaviour (? not necessarily designed as such). Each one also has its own geometry stored
Combine into using one window with common code where possible

Leaving well alone :)

windymilla:
Bookloupe/gutcheck and jeebies each had their own error dialog, with duplicated
code, etc. Now all use the same dialog.
Highlighting code from bl/gc now highlights errors from other tools.

Tests added to test suite for bl/gc and jeebies.

Fixes #157

A separate issue #178 relates to making all line formats the same to give further
consistency and use of common code.

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -6,24 +6,27 @@ BEGIN {
     use Exporter();
     our ( @ISA, @EXPORT );
     @ISA    = qw(Exporter);
-    @EXPORT = qw(&errorcheckpop_up &gcheckpop_up &gutcheck &jeebiespop_up);
+    @EXPORT = qw(&errorcheckpop_up);
 }
 
 # General error check window
-# Handles HTML & CSS Validate, Tidy, Link Check
+# Handles Gutcheck, Jeebies, HTML & CSS Validate, Tidy, Link Check
 # pphtml, pptxt and Load External Checkfile,
-# TODO: Incorporate Gutcheck & Jeebies as well to avoid code duplication and divergence?
 sub errorcheckpop_up {
     my ( $textwindow, $top, $errorchecktype ) = @_;
     my ( %errors,     @errorchecklines );
     my ( $line,       $lincol );
     ::hidepagenums();
+
+    # Destroy and start afresh if already popped
     if ( $::lglobal{errorcheckpop} ) {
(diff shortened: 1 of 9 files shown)


src/lib/Guiguts/ErrorCheck.pm
```diff
@@ -153,153 +198,204 @@ sub errorcheckpop_up {
...
+            if (
+                    ( not $::verboseerrorchecks )
+                and ( $thiserrorchecktype eq 'W3C Validate CSS' )
+                and (  ( $line =~ /^To show your readers/i )
+                    or ( $line =~ /^Valid CSS Information/i ) )
+            ) {
+                last;
+            }
+            if (
+                ( $line    =~ /^\s*$/i )                                    # skip some unnecessary lines from W3C Validate CSS
```
cpeel, line 231: While it was like this before, would this comment perhaps be better before the `if` block? As-is it seems to imply it only applies to the current line which is incorrect.

src/lib/Guiguts/ErrorCheck.pm
```diff
@@ -1308,45 +1107,26 @@ sub gutcheck {
...
-    if ( $::gcopt[5] ) { $gutcheckoptions .= 'l' }
-    ;                                    # Line end checking - defaults on
-    if ( $::gcopt[6] ) { $gutcheckoptions .= 'v' }
-    ;                                    # Verbose - list EVERYTHING!
-    if ( $::gcopt[7] ) { $gutcheckoptions .= 'u' }
-    ;                                    # Use file of User-defined Typos
-    if ( $::gcopt[8] ) { $gutcheckoptions .= 'd' }
-    ;                                    # Ignore DP style page separators
+    my $gutcheckoptions = '-ey';    # e - echo queried line. y - puts errors to stdout instead of stderr.
+    $gutcheckoptions .= 't' if $::gcopt[0];    # Check common typos
```
cpeel, line 1115: This block is *much* more readable now - nicely done.

cpeel (commented):
One minor comment, but this looks great.

windymilla:
Thanks - slightly tidied - will do more tidying of same area in other MRs

PR #238, opened 2020-09-24
Consolidate error line processing code in ErrorCheck.pm

### LINKED ISSUE #178: Consolidate/simplify line number processing in ErrorCheck.pm

windymilla:
Duplicated and similar code extracted from large if-elsif block and placed at end.
Varying output format of lines/columns from different tools rationalised to the form:
`line:column error message` with lines starting at 1, and columns at 0, as per Guiguts
notation.

Baseline files updated for test suite.

Fixes #178


cpeel:
Love the standardization! Nothing jumps out to me, but I'm not intimately familiar with the tool outputs and such.

windymilla:
Thanks. The test suite has been really useful to check that I'm only
changing output in the way I intended.

PR #251, opened 2020-09-27
Remove the Old menu layout

### LINKED ISSUE #85: Remove the Old menu layout
`Preferences | Menu Structure` has two options: `Default Menu Structure` and `Old Menu Structure`.

Remove the `Menu Structure` cascade and the code which handles displaying the Old menu layout.

[It was announced](https://www.pgdp.net/phpBB3/viewtopic.php?p=1209549#p1209549) that 1.1.0 would be the last release to support both menu structures.

windymilla:
As well as removing old menu layout, two separate commits to break up the Check All test into smaller tests and to remove Check All from ErrorCheck.pm, since this is no longer accessible to the user.

Fixes #85


src/lib/Guiguts/ErrorCheck.pm
```diff
@@ -176,203 +168,200 @@ sub errorcheckpop_up {
...
+    my $size = @errorchecklines;
+    if ( ( $errorchecktype eq "W3C Validate CSS" ) and ( $size <= 1 ) ) {    # handle errors.err file with zero lines
+        push @errorchecklines,
+          "Could not perform validation: install java or use W3C CSS Validation web site.";
+        next;
     }
+    push @errorchecklines, "Check is complete: " . $errorchecktype
+      unless $errorchecktype eq 'Load Checkfile';
+    if ( $errorchecktype eq "W3C Validate" ) {
+        push @errorchecklines, "Don't forget to do the final validation at http://validator.w3.org";
```
cpeel, line 358: FYI, this is now https://validator.w3.org/ (note the HTTPS).

src/lib/Guiguts/ErrorCheck.pm
```diff
@@ -176,203 +168,200 @@ sub errorcheckpop_up {
...
+    }
+    if ( $errorchecktype eq "W3C Validate CSS" ) {
+        push @errorchecklines,
+          "Don't forget to do the final validation at http://jigsaw.w3.org/css-validator/";
```
cpeel, line 362: Ditto this at https://jigsaw.w3.org/css-validator/

PR #258, opened 2020-10-01
Tidy up MenuStructure code

windymilla:
1. Create separate sub to contain menu entries for each menu or submenu
2. Use common sub to create each menu & submenu
3. Use `command` menu items consistently, rather than some `command` and some
`Buttons`
4. For Unicode menu, instead of creating a separate hash in order to permit
`Sort by Name` and `Sort by Range`, just use the same hash with a sorting function
that sorts appropriately, thus avoiding two different sections of code to create it.

### DIFF
--- a/src/lib/Guiguts/MenuStructure.pm
+++ b/src/lib/Guiguts/MenuStructure.pm
@@ -13,10 +13,7 @@ BEGIN {
 # rebuild every time it is modified
 sub menurebuild {
 
-    # Delete any current menus
-    for ( 0 .. 12 ) {
-        $::menubar->delete('last');
-    }
+    $::menubar->delete( 0, 'last' );    # Delete any current menus
 
     menu_toplevel( '~File',        &menu_file );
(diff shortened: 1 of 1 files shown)


src/lib/Guiguts/MenuStructure.pm
```diff
@@ -9,6 +9,28 @@ BEGIN {
     @EXPORT = qw( &menurebuild );
 }
 
+# Menus are not easily modifiable in place. Easier to just destroy and
+# rebuild every time it is modified
+sub menurebuild {
+
+    # Delete any current menus
+    for ( 0 .. 12 ) {
```
cpeel, line 17: [This was in the original code] This may not be doable, but it feels weird to hard-code the number of menus here. I can just see a new menu getting added (or removed) and this value not getting changed in the future.

windymilla: The delete method has an option for a range, so deleting 0 to 'last' is simpler and safer. Done.

PR #263, opened 2020-10-03
Provide generic mechanism to display relevant manual page

### LINKED ISSUE #82: Provide mechanism to display manual page for any dialog
Allow the user to quickly find help on a currently popped dialog.
Possible mechanism: Bind F1 key (Windows) or other key combination to open manual page depending on which dialog has focus.

windymilla:
Press F1 key when focus is on a dialog to display the manual page for that dialog
in a browser window.

1. Create hash table using existing dialog keys to store the sub-page of the manual
2. A couple of dialogs are used for more than one purpose: append a context string
to the keys in this case, e.g. `errorcheckpop+Jeebies`
3. In generic dialog initialization sub, bind F1 to a sub to display the manual page.
4. Update hotkeys.txt to document feature.

Fixes #82

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -754,67 +754,67 @@ sub initialize {
     # manualhash stores subpage of manual for each dialog
     # Where dialog is used in several contexts, use 'dialogname+context' as key
     $::manualhash{'aboutpop'}    = '#Overview';
-    $::manualhash{'alignpop'}    = 'Guiguts_1.1_Text_Menu#Align_text_on_string';
-    $::manualhash{'asciiboxpop'} = 'Guiguts_1.1_Text_Menu#Draw_ASCII_Boxes';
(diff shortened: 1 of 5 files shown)


src/lib/Guiguts/Utilities.pm
```diff
@@ -751,6 +751,71 @@ sub initialize {
         $::geometryhash{wfpop}            = '+365+63';
     }
 
+    # manualhash stores subpage of manual for each dialog
+    # Where dialog is used in several contexts, use 'dialogname+context' as key
+    $::manualhash{'aboutpop'}    = '#Overview';
```
cpeel, line 756: This attempts to load the page: https://www.pgdp.net/wiki/PPTools/Guiguts/Guiguts_1.1_Manual/#Overview which doesn't  exist.

cpeel (commented):
Very nice! This also works in macOS with "fn+F1" (which is how the function keys are suppose to work).

windymilla:
Thanks for spotting that - fixed now

PR #266, opened 2020-10-04
Make Shift-clicking on Search buttons temporarily reverse search direction

### LINKED ISSUE #117: On S&R dialog, let Shift+Search mean "go in reverse just this once"
Like the same thing in Notepad++. When you just want to go back up to see something that flew by, without having to turn on "Reverse" and then turn it off again afterwards.

Why: Because I'm lazy. ;-)

windymilla:
If the Shift key is pressed while clicking on the Search or the Replace & Search
buttons in the S&R dialog, the following happens:
1. When the button is pressed, the "Reverse" checkbox is toggled alerting the user
to the fact that they are reversing the expected direction. (Code remembers it has
done this for later)
2. Releasing the button (whether or not Shift is still down) does the search or replace
& search according to the just-toggled direction
3. Releasing the button (whether or not Shift is still down) then toggles the direction
back to its original setting (if it had been toggled in step 1)

Fixes #117

### DIFF
--- a/src/lib/Guiguts/SearchReplaceMenu.pm
+++ b/src/lib/Guiguts/SearchReplaceMenu.pm
@@ -1195,6 +1195,7 @@ sub searchpopup {
             -padx   => 2,
             -anchor => 'w'
         );
+        search_shiftreverse( $::lglobal{searchbutton} );
         $::lglobal{searchentry} = $sf11->Text(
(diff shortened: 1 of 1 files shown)


cpeel:
Very clever solution and it works great. Nicely done!

Issue #261, opened 2020-10-03
Utilities-initialize may not initialize all dialog geometry/position values


windymilla:
**Describe the bug**
If the geometry for wfpop is defined (e.g. by loading a setting.rc file) then none of the geometry/position hash values are set. This means that a dialog which is new in a release would not get its geometry initialized if the user keeps their old setting.rc.

Instead of
```
    unless ( $::geometryhash{wfpop} ) {

        # positionhash stores user's window position
        # geometryhash stores user's window position and size
        $::geometryhash{aboutpop}         = '+312+136';
        $::geometryhash{alignpop}         = '+338+83';
...
```
it should be
```
    # positionhash stores user's window position
    # geometryhash stores user's window position and size
    $::geometryhash{aboutpop}         = '+312+136' unless $::geometryhash{aboutpop};
    $::geometryhash{alignpop}         = '+338+83' unless $::geometryhash{alignpop};
...
```
Should also be able to remove `$geometryhash{wfpop} = q{};` from guiguts.pl

windymilla:
Fixed by #268

PR #268, opened 2020-10-04
Set geometry defaults per dialog if undefined

windymilla:
Rather than just checking if wfpop geometry is undefined and then setting default
values for all dialogs, set geometry for each dialog that is undefined.
This should be safer if a new dialog is introduced in the future.


windymilla:
Oh, fixes #261

Issue #234, opened 2020-09-21
Bookloupe/gutcheck output in error window is incorrect if not verbose mode


windymilla:
**Describe the bug**
If the bookloupe/gutcheck verbose flag is turned off (not recommended) then the line reporting the number of queries is not in the correct position.

**To Reproduce**
1. Open Guiguts
2. Run Tools-->Bookloupe/Guiguts
3. Use Run Options to turn off verbose mode
4. Run Tools-->Bookloupe/Guiguts
5. Note that `-->NNN queries` line is not where it should be

**Expected behavior**
`-->NNN queries` line should come after the initial output and before the queries

windymilla:
Fixed by #269

PR #269, opened 2020-10-05
Correct reported bookloupe output in non-verbose mode

windymilla:
If verbose flag was turned off (not recommended), the summary of how many
queries there were was wrongly placed.
Simpler logic seems to work.


windymilla:
Fixes #234

cpeel:
Would it be better / simpler to always enforce verbose output from bookloupe?

windymilla:
It wouldn't simplify this bit of code. I've never altered the bookloupe
options when PPing or PPVing. I think if we wanted to remove some of all of
the run options, it should probably go in a separate issue/MR.

Issue #229, opened 2020-09-19
Load `My View` bookloupe options at start of program


windymilla:
**Describe the solution you'd like**
It is possible to `Save My View` and `Load My View` from the Bookloupe View Options dialog. To make this more useful, load the `My View` settings when the program starts, effectively giving user-defined default settings for bookloupe.

**Describe alternatives you've considered**
Turning specific options on by default, as suggested in comments to #226 

**Additional context**

windymilla:
Consider changing "My View" to "Defaults"

Issue #231, opened 2020-09-20
Running bookloupe/gutcheck doesn't use View Options if dialog closed


windymilla:
**Describe the bug**
View Options dialog allows switch off/on viewing of groups of Bookloupe/Gutcheck errors. If View Options (or main BL/GC dialog) is closed, and then BL/GC is re-run, all errors are shown until a toggle is clicked in the View Options dialog.

**To Reproduce**
1. Open Guiguts
2. Load a file
3. Run Tools-->Bookloupe/Gutcheck...
4. Click View Options in BL/GC dialog
5. Click Hide All & note all errors are hidden.
6. Close View Options dialog
7. Click Run Checks in BL/GC dialog & note all errors are shown
8. Click View Options in BL/GC dialog - note toggles are all checked despite errors being shown

**Expected behavior**
At step 7, there should be no errors shown to match state of View Options dialog

PR #275, opened 2020-10-08
Improve handling of GC/BL view options

windymilla:
Three commits:
1. Continue to use view options after view option dialog is hidden - they were remembered anyway
2. Remove test line left behind from previous commit
3. Load "My View" (renamed to Defaults) options at program start.


windymilla:
Fixes #229 and #231

PR #279, opened 2020-10-10
Speed up some operations by reducing update frequency

### LINKED ISSUE #153: Speed up rewrapping by refreshing less often
Speed up the rewrap process by only updating the screen every 10 paragraphs instead of every paragraph. Preliminary test suggests 3x speed improvement (4.5 seconds to 1.5 seconds on small file, 90 to 30 seconds on large file)

Removing updates completely, which does cause a further speedup, but no feedback to the user that it is working

**Additional context**
Relevant function calls in selectrewrap are textwindow->see, textwindow->update, and ::update_indicators

### LINKED ISSUE #277: Arguments to selectrewrap are ignored
Remove parameters passed during subroutine calls, since ignored within the routine.

windymilla:
New routine, `updatedrecently`, can be called before repeated window updates
to check if it's time to do another update. Update frequency set to 1 second.

Implemented for rewrapping, for "slow" Replace All (fast one is done in one call),
for several footnote operations(first pass, move to landing zone, and tidy up), and for
HTML body conversion. Speed-ups of 3x or more.

Also removed parameters passed to selectrewrap() since they were ignored in the
routine, paving the way for an argument to trigger silent mode which does no
screen updating when called from footnote code.

Fixes #153
Also fixes #277

### DIFF
--- a/src/lib/Guiguts/Footnotes.pm
+++ b/src/lib/Guiguts/Footnotes.pm
@@ -752,8 +752,7 @@ sub footnotefixup {
         ) if $::lglobal{fnsecondpass};
         $pointer = '';
         $anchor  = '';
-        $textwindow->yview('end');
-        $textwindow->see($start) if $start;
+        $textwindow->see($start) if $start and not ::updatedrecently();
         $textwindow->tagAdd( 'footnote', $start, $end );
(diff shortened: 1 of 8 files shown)


cpeel:
Very nice!

PR #302, opened 2020-10-18
Remove gutcheck

### LINKED ISSUE #298: Remove gutcheck from Guiguts
1. Change labels to say Bookloupe instead of Bookloupe/Gutcheck
2. Remove gutcheck from tools
3. In ErrorCheck.pm, remove special handling of utf8/bytes output to cope with gutcheck. Use the temporary file errors.tmp that is used by the other tools instead of writing a special tempfile for gutcheck

windymilla:
Bookloupe completely supercedes gutcheck and now users have had the joint labels
`Bookloupe/Gutcheck` for a full release, they are used to it being called bookloupe.

All references to gutcheck have been removed apart from variable names which
use `gc`. Some are saved in the setting.rc file and it doesn't seem worth just changing
the names when the user does not see them.

Bookloupe now uses the temporary output file which is written out for all error
checks rather than writing its own in `gutcheckrun`

Fixes #298


windymilla:
The two commits can be combined into one

PR #307, opened 2020-10-19
Remove bookloupe run options feature

### LINKED ISSUE #301: Remove `Run Options` for bookloupe
Bookloupe has various command line options. However, it is always run with the same options (the defaults). Remove the run options button, dialog and associated code. 

Benefits: simpler code, simpler interface, simpler manual

windymilla:
Bookloupe is always run with the same options. Just bake these into the code rather
than allowing the user to alter them via a dialog.

Fixes #301


cpeel:
I like this, but are we going to get any flak from removing the ability to change these options?

windymilla:
1. No PPer should be using any different options.
2. I asked in the forum and replied were either that they've never changed them or didn't know they existed. Noone said they made use of them

cpeel:
Thanks for assuaging my worries!

PR #319, opened 2020-10-28
Remove use of yview for mouse wheel and forcing scrollbars

### LINKED ISSUE #77: Tk::Error: Failed to AUTOLOAD 'Tk::Frame::yview' when using the mouse wheel in Unicode Character Search

windymilla:
It appears as though BindMouseWheel is not necessary for scrolled widgets to
scroll using the mouse wheel. With or without it, some do and some don't, probably
depending on focus and type of widget being scrolled.

Also, yview was used to scroll down and up one line on several widgets when they
get created, possibly to force optional scrollbars to become active. Without it, as
long as scrollbars are not marked as optional, their behaviour seems identical.

Please check on non-Windows platforms if possible.
Also, does this fix #77 - failed to autoload yview?

### DIFF
--- a/src/lib/Guiguts/CharacterTools.pm
+++ b/src/lib/Guiguts/CharacterTools.pm
@@ -307,7 +307,6 @@ sub utfcharsearchpopup {
             -sticky     => 'wne',
         )->pack( -expand => 'y', -fill => 'both', -anchor => 'nw' );
         ::drag($pane);
-        ::BindMouseWheel($pane);
         my $fontlist = $cframe->BrowseEntry(
             -label     => 'Font',
             -browsecmd => sub {
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -112,9 +112,7 @@ sub errorcheckpop_up {
             $textwindow->markUnset($_) for values %::errors;
         }
     );
-
     ::drag( $::lglobal{errorchecklistbox} );
-    ::BindMouseWheel( $::lglobal{errorchecklistbox} );
 
     # button 1 views the error
     $::lglobal{errorchecklistbox}->eventAdd( '<<view>>' => '<ButtonRelease-1>', '<Return>' );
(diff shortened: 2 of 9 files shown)


windymilla:
Note that yview is still used on main textwindow, as well as for keyed alpha search in Word Frequency box.

cpeel:
This does fix the error in https://github.com/DistributedProofreaders/guiguts/issues/77. We should probably align the dialog name with the menu at some point (per the last part of that bug):
> Aside: The "UTF Character Search" menu item should probably be renamed "Unicode Character Search". The "UTF Character Entry" menu doesn't match the dialog name that opens up -- both of which should probably be "Unicode Character Entry".

PR #348, opened 2020-11-21
Provide redesigned HTML Markup dialog

### LINKED ISSUE #96: Improve method of adding language markup
Requested in this [forum post](https://www.pgdp.net/phpBB3/viewtopic.php?f=13&t=11466&p=1207077#p1207077)
Options:
a) offer multiple "Surround selection with" options
b) add extra language options in HTML markup dialog (cite is already there, as is span with user-defined attributes)
c) add a language markup dialog
d) recommend/document regex search/replace solution, e.g. converting italic markup to language

### LINKED ISSUE #120: Several changes to the HTML Markup dialog
1. The following buttons generate incorrect (although not invalid) markups: mdash, ndash, dagger, Dagger, lsquo, rsquo, sbquo, ldquo, rdquo, and bdquo.  Instead of just the character indicated, _e.g._, `&mdash;`, they generate opening and closing tags that contain those characters, _e.g._, when the word 'called' is selected in the text, the 'mdash' button generates `<&mdash;>called</&mdash;>`. If nothing's selected, it generates `<&mdash;></&mdash;>`.  It should just generate `&mdash;` but the question is, "where?" at the beginning or the end of the selection?

2. Does anyone use the 'q' or 'del' tags?  I'd ask the same of most of the characters generated by the first question, even if the buttons worked properly. Many of those character-buttons probably were useful early-on, but with the tools we have now, they're archaic.

3. What would be more useful, IMHO, would be ways of letting each of us customize some of those buttons, or to allow multiple rows of &lt;div&gt; and &lt;span&gt; text-boxes/buttons.  Especially multiple &lt;span&gt;s. Something like this allows 4 lines (could be more), each of which can be a span or  a div, and might be a relatively easy modification:
Gg1 0-47d-html_markup_new

4. At the bottom, the 'Insert Header' and "Apply Poetry Markup to Sel." buttons are pretty much obsolete, as "Auto Generate" will already have done that.  I doubt whether "Find Some Orphaned Markup" finds anything that the HTML Validator missed, so it's redundant/archaic.

windymilla:
1. A few obsolete buttons removed
2. Other buttons re-ordered for more consistency
3. Top rows of buttons are now configurable: Mouse-3 or Ctrl/Meta-Mouse-1 pops
a small dialog to type in either a classname or some attribute styling. Subsequent
clicks will continue to include this additional HTML in the markup inserted.
4. Next section contains div, span and i buttons with text fields for additional
attributes. Currently 4 such rows.
5. Fields in point 4 have a history pulldown menu. Code commonised to provide
generic history for an entry widget - also used in S&R dialog.
6. The Internal link, External link and Named Anchor can be similarly configured
to include a class or other attributes.
7. Code for small dialog box with entry field used for config has been commonised -
very similar code was used for the Goto Page & related dialogs.
8. All the config and the history is saved in the setting.rc file between runs.

Fixes #96
Fixes #120

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -135,8 +135,6 @@ our $globalspellpath        = q{};
 our $globalviewerpath       = q{};
 our $globalprojectdirectory = q{};
 our @gsopt;
-our $htmldiventry           = ' class="i2"';
-our $htmlspanentry          = ' class="i2"';
 our $highlightcolor         = '#a08dfc';
 our $history_size           = 20;
 our $ignoreversions         = "none";                      # Don't ignore any updates by default
@@ -212,6 +210,11 @@ our @replace_history;
 our @search_history;
 our @sopt = ( 0, 0, 0, 0, 0 );    # default is not whole word search
 our @wfsearchopt;
+
+# html markup dialog
+our @htmlentry = ('') x 4;        # class/attributes for each div, span, i button
+our @htmlentryhistory;            # single shared history list htmlentry
+our %htmlentryattribhash;         # class/attributes for each element button
 
 our %htmllabels;
 our %convertcharsdisplay;
--- a/src/lib/Guiguts/FileMenu.pm
+++ b/src/lib/Guiguts/FileMenu.pm
@@ -867,8 +867,7 @@ EOM
             qw/alpha_sort activecolor auto_page_marks auto_show_images autobackup autosave autosaveinterval bkgcolor
             blocklmargin blockrmargin bold_char cssvalidationlevel defaultindent donotcenterpagemarkers epubpercentoverride failedsearch
             font_char fontname fontsize fontweight geometry
-            gesperrt_char globalaspellmode highlightcolor history_size
-            htmldiventry htmlspanentry ignoreversionnumber
+            gesperrt_char globalaspellmode highlightcolor history_size ignoreversionnumber
             intelligentWF ignoreversions italic_char jeebiesmode lastversioncheck lastversionrun lmargin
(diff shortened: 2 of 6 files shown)


windymilla:
No problem if you don't have time to try it on a Mac, but if you do try it, could you particularly check if the Meta-Mouse-1 functionality works:
1. Pop HTML-->HTML Markup...
2. Right click on first button `<em>`. It should pop a mini dialog you can type into, e.g. "xyzzy". When you click OK, it should paste some HTML into the main window: `<em class="xyzzy"></em>`
3. Now try using Meta-Mouse-1 instead of Mouse 3. It should behave just the same.
As I said, if it doesn't fit it with your other work, rest and play, don't worry - someone can check it when we get to an alpha/beta release.
Thanks

cpeel:
The new layout is a _huge_ usability improvement - nicely done! I confirmed that both right-clicking and control+click bring up the class attribute entry dialog on the Mac. Settings were persisted across program executions just as I expected.

Should the markup buttons that are using a non-empty class attribute have some visual indicator to that effect? It may not be necessary depending on how users use them, but it seems like it might be a surprise if you were expecting a regular `<em>` but had used one with a class a few months ago and it was retained.

windymilla:
That's a good idea - maybe appending a plus sign to the button label, e.g. `<em>+`, `Internal link+`, etc.
I think changing the background color might be a bit too intrusive.
I'll put together another commit implementing the +, unless you have a better idea.

cpeel:
I think a `+` is a great idea.

windymilla:
That seems OK with a plus sign - not too intrusive. Had to shorten a couple of labels to make them fit. See what you think...

cpeel:
This looks good to me!

PR #354, opened 2020-11-25
Improve font configuration

### LINKED ISSUE #345: Font in Search & Replace entry widgets (and others?)
**Describe the bug**
The change to use Entry instead of Text widgets in the S&R dialog has meant the font has changed from a mono serif font (Courier?) to a sans serif font (Helvetica?). This is the same as all the other Entry widgets, but is a change, and the font is less easy to read for text such as is needed in search & replace operations.

**Expected behavior**
Consider
1. Just change the font for the S&R Entry widgets: `-font => '{Courier New} 8',`
2. Change the font for all Entry widgets: see below
3. Allow the user control over the font size/type using the Prefs menu, or see below for user command line options.

**Additional context**
The user can specify command line options, such as
```
perl guiguts.pl -xrm "*font: Helvetica 10"
perl guiguts.pl -xrm "*font: Helvetica 10" -xrm "*Entry*font: Courier 10"
```
This can also be done programmatically:
```
$top->optionAdd('*Entry*font' => '{Courier New} 8');
```

windymilla:
1. Create named fonts for proofing, for unicode/Greek and for text entry boxes
2. Named fonts allow instant change to all widgets using the font when the font is
updated, rather than having to individually reconfigure the widgets
3. Text entry box font defaults to the system default rather than user-defined font.
4. Most widgets will present the same font as previously, but a few text entry boxes
have been changed from the proofing font to the default. Also Unicode character
entry dialog and Latin-1 dialog now use the 'Greek/Unicode' font. No widgets use
other arbitrary fonts, e.g. Times.
5. Updated Prefs dialog to allow configuration of all 3 fonts. Uses SpinBox widget
for font size.
6. Commonised code where possible, including setting of font in the Unicode dialog
and the Unicode Character Search dialog.

Fixes #345


cpeel:
The "Use System Default For Text Entry Fields" checkbox was a bit confusing to me at first, because it was selected but the Text Entry Fields font was set to "Courier New" so I assumed that was the system default (since I hadn't set it to anything myself) and the text entry fields were clearly not using Courier New.

This may be more effort than it's worth, but is there a way to disable the Text Entry Fields inputs (or hide it) if the checkbox is selected?

Bonus would be to only show the alert about needing to restart Guiguts after the user has hit the OK button and the value of the checkbox had changed since the user opened the dialog. On my system, when the warning pops up it puts the Font Preferences dialog behind the main window and I almost missed it back there and toggling the checkbox just makes it pop up yet again.

windymilla:
1. Good point about disabling the widgets when the system font box is checked - I had intended to do that once I had it all working, but forgot :)
2. The comments about the confusion with restarting made me redouble my earlier efforts to make the use of the system font smoother. Eventually worked out how to get system font info and incorporate it into the new mechanism, so restarts are no longer required at all.

Hopefully OK to merge now.

cpeel:
This works beautifully. Clever way of figuring out and using the system font information as well!

Issue #133, opened 2020-07-19
Replace Latin-1 chart with charts for DP character suites


windymilla:
**Describe the solution you'd like**
The Tools-->Character Tools-->Latin-1 Chart does not cover all the characters needed now that DP/PG use Unicode. Provide a chart for each (commonly-used?) [DP character suite](https://www.pgdp.net/c/tools/charsuites.php)

**Describe alternatives you've considered**
The Tools-->Character Tools-->UTF Character Entry and Character Search tools provide a way of entering characters, but not organised into DP character suites, i.e. what should be helpful for the project.

windymilla:
After discussion with @srjfoo also need to consider characters that are commonly used by PPers but not by proofers. Examples are curly quotes, emdashes, endashes, etc. Also, what about non-breaking spaces or other non-printing characters?

cpeel:
> The Tools-->Character Tools-->Latin-1 Chart does not cover all the characters needed now that DP/PG use Unicode. Provide a chart for each (commonly-used?) [DP character suite](https://www.pgdp.net/c/tools/charsuites.php)

It's also worth thinking broader than DP -- do folks at DPCanada use GG? What about other PG suppliers? If so, making the Unicode charts pgdp.net character suite specific might not be useful.

windymilla:
Forum comments from [here](https://www.pgdp.net/phpBB3/viewtopic.php?f=13&t=70756&p=1212558#p1212416) onwards suggest a preferred solution would be to just change what is in the chart to cover characters commonly needed by PPers.

> 3) Put Latin-1 supplement block into the Unicode menu. Retain the special chart, but move away from calling it a Latin-1 chart, and call it something like a "Commonly used" chart. It would have the relevant characters from Latin-1 supplement, but also very common characters from other blocks, e.g. curly quotes, em-dash, en-dash, maybe fractions other than 1/2, 1/4, 3/4, etc.

windymilla:
Also see #143

windymilla:
Change buttons to labels as was done in #361 to reduce spacing

windymilla:
Fixed by #371

PR #361, opened 2020-11-30
Improve Greek transliteration dialog

### LINKED ISSUE #162: Allow user to cancel Greek Transliteration dialog without inserting Greek
When there is text in the Greek transliteration dialog, crossing the dialog off inserts the text - it should be necessary to click "OK" or "Insert" to insert, and either "Cancel" or "crossing it off" should insert nothing.

This is a specific instance of #129

### LINKED ISSUE #351: Use the unicode/Greek font for Greek character buttons in Greek dialog
Possibly for historical reasons, the buttons in the Greek dialog that insert Greek letters use small images, rather than characters from the unicode/Greek font. 
This means the inserted character may not match the button very well. 
Also the size of the buttons does not change when the font size is altered up or down.
Also, the rough breathing mark (dasia) looks like an acute accent in the images, which could cause confusion.

Leave it alone - maybe some fonts might not have very good representations of Greek characters, but then that isn't a good font for doing Greek work. Hopefully DPSansMono has good representations.

### LINKED ISSUE #352: Commonise duplicated code in Greek dialog
1. betagreek is essentially two separate routines for converting in different directions
2. bettergreek similarly
3. command subs for 'ASCII->Greek' button & similar are all very similar - could be common, passing in a ref to the conversion routine
4. Button creation for different rows is also identical apart from passing in correct element of column array hash.
5. Could then loop from row 1 to 4, skipping unused spaces within the common function when undefined value is passed in.

windymilla:
1. Use text labels rather than images for the Greek letter buttons. This means the
inserted character in the Greek dialog will match the glyph on the button. Also, some
of the characters had an acute accent instead of a breathing mark in the old version.
This means Greekgifs and GifMaker files are no longer required.
4. If the ou ligature is clicked, it now inserts the ou ligature character. If the user
wants separate o and u, they can just the o and u buttons. This makes more sense
now that utf8 text & HTML versions are preferred.
5. Some bits of code have been commonised by splitting bi-directional conversion
routines with (virtually) no common code, paving the way to pass the conversion
routine as an argument to the common callback routine.
6. Extract a long validation routine to improve readability of the main dialog popup
routine.
7. Don't reset the type of input (Latin-1, name, HTML, utf8) to Latin-1  if dialog is
closed then re-opened. Default to utf8 when program starts.

Fixes #162
Fixes #351 
FIxes #352


cpeel:
It appears that on Mac (at least) the initial dialog size is the scrunched up zero-space version:
Screen Shot 2020-12-01 at 8 51 51 AM

And you have to make the dialog *very big* to see all of the controls ~ 1058x576 which might be a bit too large for some users?
Screen Shot 2020-12-01 at 8 52 31 AM

windymilla:
Regarding initial squished up dialog, is that on a fresh release, i.e. no setting.rc file? I can't reproduce it here.
Also, which font do you have selected for Unicode/Greek in the Prefs|Appearance|Set Fonts... dialog (default should be Courier New 14)? 
Can you adjust the font to a smaller size to make the dialog acceptably-sized? One thing I can think of is to reduce the font size by a couple of points where there are two letters on a button, e.g. th or gamma-gamma. That would add complication and might not look good. Another option might be to try to force the button width/height, except that I think that is specified in character widths/heights with the minimum being 1 - I could look into that though.
Also, when I use DP Sans Mono, I notice the buttons are a quite a bit taller and wider than with Deja Sans Mono - do you know anything about reasons for that?
The Unicode dialog, which I haven't changed yet looks as though its spacings etc are the same, depending on which Unicode block you choose - if you choose Greek, it looks the same, so I don't think I've done anything specific wrong, but happy to be enlightened if you have any ideas... 😄

cpeel:
> Regarding initial squished up dialog, is that on a fresh release, i.e. no setting.rc file? I can't reproduce it here.
> Also, which font do you have selected for Unicode/Greek in the Prefs|Appearance|Set Fonts... dialog (default should be Courier New 14)?

This is without a `settings.rc` file and is using Courier New 14 for the display.

> Can you adjust the font to a smaller size to make the dialog acceptably-sized? One thing I can think of is to reduce the font size by a couple of points where there are two letters on a button, e.g. th or gamma-gamma. That would add complication and might not look good. Another option might be to try to force the button width/height, except that I think that is specified in character widths/heights with the minimum being 1 - I could look into that though.

Making the font smaller does make the window smaller. There's a lot of whitespace on the sides of the buttons -- is it possible to make that smaller (say half of what it is now)? I'm happy to punt this and see if anyone complains about it. I'm not a real user.

> Also, when I use DP Sans Mono, I notice the buttons are a quite a bit taller and wider than with Deja Sans Mono - do you know anything about reasons for that?

Yes, the ascenders and descenders were tweaked in DP Sans Mono to be longer so the overall font height for DP Sans Mono is bigger than DejaVu Sans Mono.

> The Unicode dialog, which I haven't changed yet looks as though its spacings etc are the same

I hadn't looked at these recently, and I'm sad to be the bearer of bad news but these no longer show any character for me and there are no errors output. ☹️  I'll see if I can't find out where this broke for me.

windymilla:
Trying labels instead of buttons - they look and behave much the same as the flat buttons that were used previously, but they are smaller. If the size issue seems fixed, I think it would be good to merge this, unless you think this MR is what has made your Greek dialog squished. If not, let's add another MR for that.
If it's OK, then I'll add another MR to make the same (button->label) change to the Unicode dialog.
I'll change the Latin-1 dialog when I change its function from Latin-1 to "commonly needed characters" as per #133.

cpeel:
This indeed fixes the dialog being too big. It doesn't fix "dialog starts all scrunched up on every open" but I'm fine merging this in and opening that up as a separate bug.

Agreed on a separate bug on the Unicode dialogs. I can't _not_ reproduce that now on any GG checkout and more investigation is needed. I'll open that after some investigation.

PR #363, opened 2020-12-03
Fix Greek dialog being zero-sized on Macs

### LINKED ISSUE #362: Greek Transliteration dialog always opens at smallest possible window size
**Describe the bug**
The Greek Transliteration dialog always opens in the smallest possible window size that MacOS supports (the size of the 3-button window controls wide x resizing icon tall). The dialog can be resized bigger. The new size is retained when accessing the dialog during the same GG session, but is small again if GG is closed and reopened.

**To Reproduce**
1. Open Guiguts
2. Open Tools -> Character Tools -> Greek Transliteration
3. See the teeny-tiny dialog

**Expected behavior**
The dialog has a sensible initial size showing all the controls.

**Environment**
* Guiguts version: 53a7a4e4b
* OS and version: macOS Catalina
* Perl version: v5.30.3
* Perl distribution: homebrew

**Screenshots**
Screen Shot 2020-12-01 at 8 51 51 AM

**Additional context**
IIRC there was another dialog that had the same problem on the Mac but not elsewhere. I don't think we solved it there either but we gave the dialog a sensible default size and persisted the size in `settings.rc`. It would be interesting to understand what we're doing for some dialogs that size correctly and for others that don't.

windymilla:
The geometry reported by `$::lglobal{grpop}->geometry;` appears incorrect.
It shows width & height of 50x8 which are the width and height of the scrolled
text widget contained in a frame contained in grpop, when it is expected to
return the pixel width & height of the grpop dialog.

Whether this is a bug or not, and whether it is related to the Mac size issue, changing
from using positionhash to geometryhash fixed a similar issue in #204, so it's worth
a try.

Hopefully will fix #362

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -687,7 +687,7 @@ sub initialize {
     $::positionhash{gotolinepop}      = '+400+400'        unless $::positionhash{gotolinepop};
     $::positionhash{gotopagpop}       = '+400+400'        unless $::positionhash{gotopagpop};
     $::geometryhash{gcviewoptspop}    = '+264+72'         unless $::geometryhash{gcviewoptspop};
-    $::positionhash{grpop}            = '+144+153'        unless $::positionhash{grpop};
+    $::geometryhash{grpop}            = '700x500+100+100' unless $::geometryhash{grpop};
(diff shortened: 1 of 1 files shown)


cpeel:
This did indeed fix #362 - thanks!

PR #365, opened 2020-12-03
Improve Unicode dialog

### LINKED ISSUE #364: Reduce size of Unicode dialog
Due to the extended ascenders/descenders of the DP Sans Mono font and extra padding given to buttons, the Unicode
dialog is rather large. Change the buttons to labels as was done in #361.

(Check if geometry method returns widthxheight in expected units, i.e. pixels. If not, may need to use geometryhash instead of positionhash for Unicode dialog too as per #363 ).

windymilla:
Use labels instead of buttons since they take less space.
Reorganise long top line of widgets into 2 rows, so dialog can be narrower.
Other minor code restructuring, including speeding up creation of dialog by
removing unnecessary update every time a button was added.

Fixes #364 - note no problems with geometry in this dialog

### DIFF
--- a/src/lib/Guiguts/CharacterTools.pm
+++ b/src/lib/Guiguts/CharacterTools.pm
@@ -136,26 +136,19 @@ sub doutfbuttons {
             my $msg = "Dec. $ord, Hex. " . sprintf( "%04X", $ord ) . ", $cname";
 
             # Use label instead of button since it takes less space
-            my $w = $::lglobal{utfframe}->Label(
-                -activebackground   => $::activecolor,
-                -text               => $text,
-                -font               => 'unicode',
-                -relief             => 'flat',
-                -borderwidth        => 0,
-                -background         => $::bkgcolor,
+            my $w = $::lglobal{utfframe}->Button(
+                -activebackground => $::activecolor,
+                -text             => $text,
+                -font             => 'unicode',
+                -relief           => 'flat',
+                -borderwidth      => 0,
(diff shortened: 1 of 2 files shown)


src/lib/Guiguts/CharacterTools.pm
```diff
@@ -140,37 +128,45 @@ sub doutfbuttons {
...
+            my $text = chr($ord);
 
-            # FIXME: See Todo
-            $buttons[ ( $y * 16 ) + $x ] = $::lglobal{utfframe}->Button(
+            my $cname = charnames::viacode($ord);
+            next unless $cname;    # Unnamed - don't create button
+            my $msg = "Dec. $ord, Hex. " . sprintf( "%04X", $ord ) . ", $cname";
 
-                #    $buttons( ( $y * 16 ) + $x ) = $frame->Button(
+            # Use label instead of button since it takes less space
```
cpeel, line 138: I find the spacing with the button better as it's easier to read. With only 16 of these across (instead of 29) they fit easily on the screen.

cpeel (commented):
Relatedly, I figured out why the Unicode dialogs no longer work for me. `charnames::viacode()` is always returning `undef`. @srjfoo's system is working fine with the same middleware version so something is amiss with my install apparently. I'm digging into this some more. We use this function is other places too, but the Unicode dialog is the only place where we use it for more than supplementary data I believe.

windymilla:
Should be OK to merge now - thanks

PR #371, opened 2020-12-08
Replace Latin-1 chart with Commonly-Used Characters Chart

windymilla:
Since books are not generally Latin-1, more helpful to PPers to have common
characters, such as curly quotes, fractions, emdashes, etc., available in a chart.
1. Replace Latin-1 chart with Common Characters Chart
2. Mouse-3 copies to clipboard, like the Unicode Chart does
3. Spare buttons at the end are left blank (actually a space) which the user can
configure with Ctrl-Mouse-1
4. Config dialog accepts a single character, or a decimal ordinal, or a hex ordinal.
Prefix the hex ordinal with `0x`, `\x` or `x`, or if it has letters [a-f] in it, then prefix
may be omitted. For any other character string, just take the first character.
5. User-defined characters are stored between runs of the program, and in dialog
is upgraded in a future release, they should be preserved to fit into any unused
spaced in the new version.
6. Make a couple of helper functions common between new dialog and Unicode
dialog

### DIFF
--- a/src/lib/Guiguts/CharacterTools.pm
+++ b/src/lib/Guiguts/CharacterTools.pm
@@ -552,7 +552,7 @@ sub utfcharentrypopup {
                 }
                 $outentry->delete( '1.0', 'end' );
                 $outentry->insert( 'end', $char );
-                $charlbl->configure( -text => $name );
+                $charlbl->configure( -text => $name ) if $name;
                 return 1;
             },
         )->grid( -row => 1, -column => 2, -pady => 5 );
--- a/src/lib/Guiguts/CharacterTools.pm
+++ b/src/lib/Guiguts/CharacterTools.pm
@@ -168,6 +168,8 @@ sub usercharconfig {
     charbind3( $w, $text );                                            # Right-click button copies text to clipboard
     charbuttonballoon( $w, $blln, $ord );                              # Balloon message if user hovers over button
 
+    ::savesettings();                                                  # Ensure new button definition saved to setting file
+
     # stop class callback being called - possible due to binding reordering during button creation above
     $w->break;
 }
--- a/src/lib/Guiguts/CharacterTools.pm
+++ b/src/lib/Guiguts/CharacterTools.pm
@@ -6,87 +6,197 @@ BEGIN {
     use Exporter();
     our ( @ISA, @EXPORT );
     @ISA    = qw(Exporter);
-    @EXPORT = qw(&latinpopup &utfpopup &utfcharentrypopup &utfcharsearchpopup &cp1252toUni);
+    @EXPORT = qw(&commoncharspopup &utfpopup &utfcharentrypopup &utfcharsearchpopup &cp1252toUni);
 }
 
-sub latinpopup {
+#
+# Create popup containing commonly used characters
+# Buttons inserting a space can be defined by user, configured by Ctrl-click, stored in $::userchars
+# This allows as many as possible of user's definitions to be preserved if in a future upgrade,
+# more buttons are defined, preferably keeping at least as many user-defined buttons
+sub commoncharspopup {
     my $top = $::top;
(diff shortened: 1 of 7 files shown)


src/lib/Guiguts/CharacterTools.pm
```diff
@@ -131,11 +240,6 @@ sub doutfbuttons {
             my $ord  = hex($start) + ( $y * 16 ) + $x;
             my $text = chr($ord);
 
-            my $cname = charnames::viacode($ord);
-            next unless $cname;    # Unnamed - don't create button
```
cpeel, line 135: Oh lovey, this fixes the Unicode dialog for me, despite my "charnames::viacode() always returning undef" probem.

src/lib/Guiguts/CharacterTools.pm
```diff
@@ -6,87 +6,197 @@ BEGIN {
...
+    );
+
+    # Replace empty string with space character
+    my $text = $::lglobal{comcharconfigval} ? $::lglobal{comcharconfigval} : ' ';
+
+    # If more than one character, interpret as decimal ordinal, hex ordinal or just take first char
+    if ( length($text) > 1 ) {
+        if ( $text =~ /^\d+$/ ) {    # decimal integer
+            $text = chr($text);
+        } elsif ( $text =~ s/([\\0]?[x])?([0-9a-f]+)/$2/i ) {    # hex (with x, \x, 0x or no prefix)
```
cpeel, line 155: You might consider adding the `U+xxxx` syntax since that's what we are showing to PMs/PPs on the [Character Suites page](https://www.pgdp.net/c/tools/charsuites.php) and in [some of our docs](https://www.pgdp.net/wiki/Recommended_custom_characters_list) at pgdp.net. It still hexadecimal.

cpeel (commented):
This works great! And thanks for (inadvertently?) fixing my problems with the Unicode dialogs 😁

src/lib/Guiguts/CharacterTools.pm
windymilla: Good idea - new commit added

cpeel: The `U+` syntax is working great, thanks.

src/lib/Guiguts/FileMenu.pm
```diff
@@ -939,6 +939,13 @@ EOM
         }
         print $save_handle ");\n\n";
 
+        print $save_handle ("\@userchars = (\n");
+        for (@::userchars) {
```
cpeel, line 943: This isn't getting saved for me. I am able to add a letter and it persists in the dialog while GG is open, but when I exit GG the results aren't persisted. Here's the relevant section of `settings.rc` after saving an entry:
```
@userchars = (
);
```

The test was done starting with a non-existent `settings.rc` file.

windymilla:
Weird. I'll look into it tomorrow. Either a last minute change - oops - or
a platform dependent issue. I'll particularly look at coping with null
returns from via code too.
Thanks

src/lib/Guiguts/FileMenu.pm
windymilla: I began to type "It's difficult to see (i.e. probably glaringly obvious) how the dialog could re-populate correctly with the new user-defined button, but for the userchars array to still be zero-length when saved in the file." when I realised one obvious reason would be if I didn't save the settings file when a new definition was added, so the saved setting.rc was one saved before the new button was defined :) 
New commit added to fix that, and another to do the same for the HTML Markup dialog - if the Common Chars chart configuration now saves properly for you however immediately you exit after defining a button, then should be OK to merge.

windymilla:
All done, I think!

windymilla:
By the way, this should fix #133

cpeel:
Fixed and looks good!

PR #374, opened 2020-12-12
Remove debug button and code from MultiLingual spelling

### LINKED ISSUE #137: In the "Spell Check in Multiple Languages" dialog, change "Save Debug Files"
The problem that I am trying to address is in the "Spell Check in Multiple Languages" dialog, there are several buttons that enable actions by the user to bring about the intent of the dialog, and mixed in with them is "Save Debug Files". This should not be in the user's face. If you want to keep it, I am not sure the best way to segregate it: Change it to right click on an unusual part of the dialog? Hit a function key? As it is now, it appears that the dialog is still under development.

windymilla:
1. Remove the Debug Files button from the dialog
2. Remove chunks of debugging code
3. Create the Multilingual Spellcheck Help dialog in the standard manner, using
 initialize_popup_with_deletebinding()

Fixes #137

### DIFF
--- a/src/lib/Guiguts/MultiLingual.pm
+++ b/src/lib/Guiguts/MultiLingual.pm
@@ -9,7 +9,6 @@ BEGIN {
     @ISA     = qw(Exporter);
     @EXPORT  = qw(&spellmultiplelanguages);
 }
-my $debug = 0;    # debug set for now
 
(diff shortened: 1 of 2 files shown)


cpeel:
Nice diffstat! 😁

PR #376, opened 2020-12-13
Fix S&R bug when text substituted for $1 contains literal $2

### LINKED ISSUE #121: Search/replace error when TEXT contains a $ sign followed by a number
This was previously reported at https://sourceforge.net/p/guiguts/bugs/135/ 
And similar at (https://sourceforge.net/p/guiguts/bugs/134/
Guiguts 1.1.0 also does this 100% of the time.
okrick
_______________________________________________________
`Search: <table((.|\n)+?)</table>`
`Replace: <div class="center"><div class="ilb">\n<table$1</table>\n</div></div>`

Text:
```
<hr />
<table class="p1" summary="Errors in the books">
  <tr>
    <td class="tdl">An overstatement of net income of</td>
    <td class="tdr">$2,721,068</td></tr>
  <tr>
    <td class="tdl">A mischarge of worn-out equipment to profit and loss of</td>
    <td class="tdr">2,843,596</td></tr>
</table>
<hr />
```
Result:
```
<div class="center"><div class="ilb">
<table class="p1" summary="Errors in the books">
  <tr>
    <td class="tdl">An overstatement of net income of</td>
    <td class="tdr">
,721,068</td></tr>
  <tr>
    <td class="tdl">A mischarge of worn-out equipment to profit and loss of</td>
    <td class="tdr">2,843,596</td></tr>
</table>
</div></div>
<hr />
```
The error occurs in the fourth line of the original text: `<td class="tdr">$2,721,068</td>` has been split into two lines and the "$2" at the beginning of the number '"$2,721,068" has been deleted.

The error is repeatable with the example given above, but does not occur every time there is a $ sign in a table: it only happens occasionally. The project in question has 169 cells containing $ signs, but only two were altered this way, and in the other affected table, it occurred in the fourth row, not in the first.
Discussion
<hr />

    Charlie Howard
    Charlie Howard - 2017-07-24

    Turns out the error occurs frequently, not just twice, and the missing information always seems to be "$2", whether followed by another digit or by a comma. But, it doesn't occur every time nor in every table.

windymilla:
In order to facilitate GG's regex extensions such as `\C...\E` and `\G...\E`, some aspects
of the substitution are done in parts rather than as a single perl regex substitution.
As part of this, parenthesis grouping/capturing and substituting were done for one
group at a time. The problem occurred when the substitution for `$1` contained the
literal text `$2`, e.g. `$234.56`. Then later, the `$2` was substituted with the text that
the second paren pair captured.

Fixed by escaping dollar characters in the match strings, so the substitution for
`$1` will contain `\$2` and won't be further substituted. Of course, a `$2` in the replace
term is still replaced with the second group match string.

This took quite a long time to determine where the problem was arising and work
where the escaping was required. I believe it fixes the issue, and I haven't been able
to make other more complex regex replacements go wrong, even if they include $2
combined with extensions such as `\U...\E`

Code also made less repetitive by using an array to capture the matches rather than
8 scalar variables

Also, removal of superfluous and potentially confusing check on regex option
inside an `unless` that precludes regex matching.

Fixes #121

### DIFF
--- a/src/lib/Guiguts/SearchReplaceMenu.pm
+++ b/src/lib/Guiguts/SearchReplaceMenu.pm
@@ -7,7 +7,7 @@ BEGIN {
     our ( @ISA, @EXPORT );
     @ISA    = qw(Exporter);
     @EXPORT = qw(&update_sr_histories &searchtext &reg_check &getnextscanno &updatesearchlabels
-      &isvalid &swapterms &findascanno &reghint &replaceeval &replace &replaceall
+      &isvalid &swapterms &findascanno &reghint &replace &replaceall
       &searchfromstartifnew &searchoptset &searchpopup &stealthscanno &find_proofer_comment
       &find_asterisks &find_transliterations &nextblock &orphanedbrackets &orphanedmarkup &searchsize
       &loadscannos &replace_incr_counter &countmatches);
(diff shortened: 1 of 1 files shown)


cpeel:
Nice investigation and solutions!

PR #377, opened 2020-12-14
Fix and tidy `\L`-type regex directives

### LINKED ISSUE #375: Regex extensions such as `\U` fail if at at start of replace string and appear twice
**Describe the bug**
Given the text `abcdefghi` a search string of `(def)` and replace string such as `\U$1\E\U$1\E' gives the result `abcDEFdef\Eghi`.

While editing that bit of code, there is a lot of duplication for the different command types, and the setting of cfound, lfound, etc. can be made simpler.

**To Reproduce**
Perform the above S&R

**Expected behavior**
Resulting text should be `abcDEFDEFghi`

**Environment**
* Guiguts version: At least 1.1.1, probably 1.0.25 and possibly beyond

**Additional context**
Probably to do with the splitting code in replaceeval() for each command type which has a special case for the command coming at start of string and looks a bit bookey.

windymilla:
1. Replaced repeated code to handle `\L`-type directives with common routine
2. Fixed bug when `\L` was used at start of replacement term and also occurred later.
3. Replaced omitted escaping of backslash in previous commit.

Fixes #375

### DIFF
--- a/src/lib/Guiguts/SearchReplaceMenu.pm
+++ b/src/lib/Guiguts/SearchReplaceMenu.pm
@@ -700,25 +700,16 @@ sub replaceeval {
     $replaceterm =~ s/\\GX/\\X/g;
     $replaceterm =~ s/\\GB/\\B/g;
     $replaceterm =~ s/\\GG/\\G/g;
+    $cfound = ( $replaceterm =~ /\\C/ );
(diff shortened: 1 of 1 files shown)


cpeel:
Wow - this is so much simpler now!! Love it.

PR #378, opened 2020-12-15
Save word count threshold for marked up phrases

### LINKED ISSUE #112: Make word count markup threshold default user-configurable
Requested [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1211529#p1211442)

In Word Frequency dialog, the Ital/Bold/SC button pops the Word Count Threshold dialog. The default value is 4. Make the default configurable via Preferences.

windymilla:
In the Word Frequency dialog, the Ital/Bold/SC button checks for the same phrase
occurring with and without markup. A configuration dialog is also popped that sets
the maximum number of words in such a phrase. Previously this was set to 4 on each
run of the program. Now save the value in the setting.rc file. Also, consider 0 to mean
no limit, rather than a limit of zero words which is not useful.

Also allow config dialog to be popped from the Prefs menu where other preferences
are set, and remove a couple of "beta" labels on menu options, since they have been
in "beta" for 10 years.

Fixes #112


cpeel:
> since they have been in "beta" for 10 years.

🤣

Issue #381, opened 2020-12-16
Allow user to add/edit compose keystrokes


windymilla:
**Describe the solution you'd like**
User should be able to enter a character or its ordinal into an entry field, and in a separate field enter the compose keystrokes to define a new compose sequence. Also to edit or delete an existing sequence. The list of sequences should be stored in the setting.rc file.

Consider whether to just allow additional "user-defined" sequences, or whether to make the default sequences editable too.

windymilla:
Wait until a user requests this

Issue #380, opened 2020-12-16
Allow user to configure key to pop Compose dialog


windymilla:
**Describe the solution you'd like**
Compose dialog is popped with Right Alt key "Alt_R". This is stored in setting.rc, and could be manually edited there, but there
should be a Prefs option to select a key, or perhaps a range of options to choose from.

Possibly an entry field in a dialog with a binding for any key press and reports the keysym, with checkbox buttons to add Shift and/or Control.

windymilla:
Duplicate of #435

PR #382, opened 2020-12-16
Implement a form of Compose Key for Unicode characters

### LINKED ISSUE #143: Add "Compose key" feature
Under Linux (and elsewhere via downloaded tools such as [WinCompose](https://github.com/SamHocevar/wincompose)) you can use a compose key followed by a couple of keystrokes to enter an accented or other non-keyboard character.
Add a similar feature to GG for adding common characters, e.g. curly quotes, etc. 

Recommend use of WinCompose (or similar)

**Additional context**
Possibilities: 
1. Pop a dialog when "compose" key (e.g. Right Alt , but configurable) is pressed. Echo keys typed, and on "Enter" dismiss dialog and insert character (use same insert code as the Latin-1 chart).
2. Same as 1, but just dismiss and insert character once a match is found with a known compose sequence
3. Dismiss dialog without inserting if Escape is pressed
4. Don't pop a dialog, just absorb further keystrokes after compose key until a compose match is found - maybe implement early exit so exit from compose mode if keystrokes entered so far can't match any known compose sequence.

windymilla:
1. Default key is the Right Alt key (marked alt gr on some keyboards). At present there
is no UI to alter this, although you could manually edit the composepopbinding entry
in setting.rc
2. Once Compose key is pressed, small dialog appears and subsequent keys pressed
appear in an entry field.
3. If the keys match a defined sequence, the dialog disappears and the character
composed is entered into the main text window, search dialog field, etc.
4. Sequences are defined for most characters in the Commonly Used Characters
dialog.
5. In addition, the user can type 4 hex digits (optionally preceded by `U+, \x, 0x or x`)
which will be interpreted instantly as a hex ordinal and the character inserted
6. Alternatively, the user can type fewer than 4 hex digits, or can type # followed by
decimal digits, then type Enter or click OK, and relevant character will be inserted.
7. If OK is clicked, but field does not contain valid sequence, the characters from the
fields are inserted
8. If Cancel (or Escape) are used, the dialog is dismissed without inserting characters.
9. In the Help menu, it is possible to pop a dialog with a list of defined sequences.
10. Issues have been entered for a UI to configure the Compose button (#380) and
also to add, edit and delete sequences (#381)

Fixes #143

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -114,11 +114,12 @@ our $blockrmargin     = 72;
 our $poetrylmargin    = 4;
 our $blockwrap;
 our $booklang          = 'en';
-our $composepopbinding = 'Alt_R';    # Default key to pop the Compose dialog
-our %composehash;                    # Keystrokes to insert character
-our $cssvalidationlevel  = 'css3';          # CSS level checked by validator (css3 or css21)
+our $composepopbinding = 'Alt_R';    # Default key to pop the Compose dialog (Right hand Alt key, also labelled AltGr)
+$composepopbinding = 'Control-m' if $OS_MAC;    # Default to Ctrl+m on a Mac - Alt+RightArrow does the same indent operation
+our %composehash;                               # Keystrokes to insert character
+our $cssvalidationlevel  = 'css3';              # CSS level checked by validator (css3 or css21)
 our $defaultindent       = 2;
-our $epubpercentoverride = 1;               # True = override % img widths to 100% for epubs
+our $epubpercentoverride = 1;                   # True = override % img widths to 100% for epubs
 our $failedsearch        = 0;
 our $fontname            = 'Courier New';
 our $fontsize            = 10;
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -113,7 +113,9 @@ our $blocklmargin     = 1;
 our $blockrmargin     = 72;
 our $poetrylmargin    = 4;
 our $blockwrap;
-our $booklang            = 'en';
+our $booklang          = 'en';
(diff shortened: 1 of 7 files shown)


windymilla:
I'd quite like to know if the default key "Alt_R" is OK on a Mac. If it's OK, I might not do either of the customisation tasks (#380 & #381) for the initial release and gauge from user feedback what is necessary. Thanks

cpeel:
`Alt_R` isn't a think on Macs based on my testing and https://wiki.tcl-lang.org/page/Modifier+Keys

I edited `settings.rc` to set this to `Control-m` and was able to test it out. This is very cool and I can see it being very useful!

cpeel:
I'm happy to merge this in as-is and instruct Mac users they need to update `settings.rc` to a valid keystroke -- let me know your thoughts.

srjfoo:
> I'm happy to merge this in as-is and instruct Mac users they need to update `settings.rc` to a valid keystroke -- let me know your thoughts.

As another Mac user, that makes sense to me.

windymilla:
Or if there is another suggestion for the default, I could make it default
to that for OS_MAC. Are there any unused keys of a similar nature? Many
Ctrl- combinations are already allocated to actions at least on Windows.

cpeel:
I'm not sure I have great suggestions for a good MacOS default. I whack-a-mole'd my way to finding `Control-m` as available -- not that I think it would be a particularly good default. I was unable to get any single-meta key working like the `Alt_R` seems to on the PC.

windymilla:
I think I've tidied up the loose ends relating to making Ctrl-m work as a compose key for Mac users. So, hopefully good to merge now. Thanks

cpeel:
Working beautifully for me!

PR #385, opened 2020-12-17
Fix Greek dialog size issues

### LINKED ISSUE #360: Resizing Greek scrolled window jumps to full screen size
**Describe the bug**
When the resize handle is used on a scrolled window in the Greek dialog (check for other similar) the whole dialog jumps to full screen width

**To Reproduce**
1. Open Guiguts
2. Open Tools-->Character Tools-->Greek Transliteration
3. Grab the resize handle where the two scroll bars meet
4. Dialog size jumps to full screen width

**Expected behavior**
User should be able to grow/shrink dialog using resize handle

**Environment**
* Guiguts version: At least since 1.0.25

windymilla:
1. Revert edit made in #362, which shouldn't have been necessary
2. Change Greek dialog size to be determined in the normal way, rather than being
sized (in characters) via the scrolled text it contains (-setgrid option)
3. Since no releases have been made between #362 and now, it is not necessary to
cope with a setting.rc that has a geometryhash entry for grpop.

Fixes #360 hopefully without breaking #362

### DIFF
--- a/src/lib/Guiguts/FileMenu.pm
+++ b/src/lib/Guiguts/FileMenu.pm
@@ -797,6 +797,9 @@ sub readsettings {
         }
     }
 
+    # Greek dialog previously had position but now needs geometry
+    delete $::positionhash{grpop} if $::positionhash{grpop};
+
     # If someone just upgraded, reset the update counter
     unless ( $::lastversionrun eq $::VERSION ) {
         $::lastversioncheck = time();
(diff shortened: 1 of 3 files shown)


windymilla:
Please check Greek dialog works OK in a fresh Mac install if you have time. Thanks

cpeel:
Alas, this re-breaks #362 for me ☹️

windymilla:
Apologies - I hoped the setgrid was the cause of the problem. This should restore you back to a working Mac version, and has a check to cope with old setting.rc files that contain a positionhash entry - a fix that should probably have been in the original #362 anyway. Should be OK to merge now

cpeel:
Confirmed the Greek dialog loads correctly on the Mac.

PR #402, opened 2020-12-23
Improve display of warnings and errors

### LINKED ISSUE #163: Display (some) error messages in error box rather than in command window
Some error messages (such as those output by warn/die) appear in the command window which is usually not seen by the user. Some of these messages should pop up an error dialog instead of, or in addition to aborting code execution.

1. Overriding behaviour of warn/die - can be done using $SIG.  Unknown what interaction if any with Perl/Tk behaviour - currently program does not die, but just exits code in current callback. Current die might therefore be useful to retain if in low level routine (or call die from within handler after popping error dialog?)
2. Replacing all calls to warn/die with calls to our own routines (which still might call die/warn after displaying meaningful error). Since there are not many, this would be possible. An advantage of this would be that the dialog-popping behaviour would happen for errors catered for in GG code, not if some lower level lib function dies or Perl error.

**Additional context**
There are about 20 uses of "die" and 50 of "warn" to review. 
An example of one that needs attention is attempting to add Good Words to the project dictionary when good_words.txt is not in the current directory. There is no visible indication that the Good Words have not been added (except hidden in the command window).

windymilla:
Warnings and errors whether issued by Tk or via warn/die in the application code
were only output to stderr, which might not be noticed by the user until it was too
later, or which on some platforms would vanish when they exited the program.

1. Override the handling of warn and die to output to a message log dialog as
well as to stderr.
2. Sound the bell (or flash the warning icon) if a warning or error is output and pop
the message log to the front.
3. Allow the user to pop the message log via the Help menu.
4. In the case of die actually going to terminate the program, give the user a chance
to copy/paste messages first.
5. Tidy up a few places that used warn/die in the code.

Fixes #163

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -287,6 +287,27 @@ our $text_font;
 our $textwindow;
 our $menubar;
 
+# End of declarations - start of code
+
+# Override handling of die and warn to display errors via user interface
+
+# Redirect die in recommended way so it doesn't catch compilation errors.
+# Note that where die is used as a result of a Tk action such as a button
+# or key press, it will abort the action rather than the whole program
+# and will use warnerror below not dieerror.
+BEGIN { *CORE::GLOBAL::die = \&::dieerror }
+
+# Redirect warn by overriding signal handler
+local $SIG{__WARN__} = \&::warnerror;
+
+# An alternative for catching Tk background errors,
+# but they are caught by the warn override above
+# sub Tk::Error {
+#     my($w, $error, @msgs) = @_;
+#     warnerror($error);
+#     warnerror($_) for @msgs;
+# }
+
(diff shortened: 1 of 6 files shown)


windymilla:
For testing:
1. There is a true die at line 334 in guiguts.pl if you specify two filenames on the command line, e.g. "guiguts.pl a b"
2. You can change this to a warn to see the effect of a warn not triggered by a Tk action.
3. You can see the effect of a die after a Tk action by doing a Spellcheck on a file, then clicking "Add Goodwords to Proj Dic" button in the Spellcheck dialog - should be like a warn plus cancel the action
4. The initial message with the version number and asking the user to note any error messages is output using the infoerror routine.

cpeel:
This is a very clever solution and well executed!

Issue #409, opened 2020-12-26
Remove Remote HTML Validation


windymilla:
**Describe the solution you'd like**
Check if users make use of Remote HTML Validation. If not, remove it, since it adds extra complexity.

**Describe alternatives you've considered**
If it is useful or vital on some platforms, then keep it and close this issue.

PR #410, opened 2020-12-26
Make all the perl modules used mandatory

### LINKED ISSUE #243: Handle better if recommended Perl modules are not installed
Several modules are "recommended but optional", e.g. Image::Size
Either make the modules required, rather than recommended, since they are easily installed via cpanm
Or warn the user if a module isn't installed about what the consequences would be, e.g. "Image::Size module is not installed - Guiguts will be unable to determine the dimensions of HTML illustrations"

Related to #242

windymilla:
Some modules were mandatory and some were "optional" but removed or degraded
some functionality. Since all the modules are easily installed using cpanm via the
provided script, it's simpler to just require all of them.

Fixes #243.

See also #409

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -28,6 +28,7 @@ use FindBin;
 use lib $FindBin::Bin . "/lib";
 
 #use Data::Dumper;
+use charnames();
 use Cwd;
 use Encode;
 use FileHandle;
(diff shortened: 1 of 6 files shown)


cpeel:
Lovely. I'm happy to see this change!

PR #413, opened 2020-12-27
Make Goto dialogs remember their screen position

### LINKED ISSUE #313: Goto dialogs don't remember their on-screen position
**Describe the bug**
There are 3 `Goto` dialogs: gotolinepop, gotopagpop, gotolabpop.
If they are moved on screen , they do not remember their position when popped again.

**To Reproduce**
1. Open Guiguts
2. Press Ctrl-j to pop the Goto Line dialog
3. Move the dialog across the screen
4. Dismiss the dialog
5. Press Ctrl-j again
6. See the dialog go back to its previous location

**Expected behavior**
Dialog should remember previous location chosen by user

**Additional context**
Probably because `DialogBox` is used as the basis for these dialogs. Consider making them regular dialogs.
ETA: Look at `popover` in gotoline function – if it goes, could goto’s remember their place? Looks like you can popover a widget, or popover the cursor or be positioned?

windymilla:
Goto dialogs previously used the DialogBox convenience widget. Now replaced with
a standard dialog containing a label and entry field, with OK and Cancel buttons.

Fixes #313

The same common routine also works for the compose dialog.

Also fixes compose pop error where multiple copies of the compose dialog could be
popped and then not all dismissed (wrong hash key used for dialog)

Also very minor simplification and standardisation of regex used to enter decimal or
hex unicode ordinals in a couple of places.

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -7,7 +7,7 @@ BEGIN {
     our ( @ISA, @EXPORT );
     @ISA    = qw(Exporter);
     @EXPORT = qw(&openpng &get_image_file &arabic &roman &popscroll
-      &cmdinterp &nofileloadedwarning &win32_cmdline &win32_start &dialogboxcommonsetup &dialogboxpopup
+      &cmdinterp &nofileloadedwarning &win32_cmdline &win32_start &dialogboxcommonsetup &textentrydialogpopup
       &win32_is_exe &win32_create_process &dos_path &runner &debug_dump &run &launchurl &escape_regexmetacharacters
       &deaccentsort &deaccentdisplay &readlabels &working &initialize &initialize_popup_with_deletebinding
       &initialize_popup_without_deletebinding &titlecase &os_normal &escape_problems &natural_sort_alpha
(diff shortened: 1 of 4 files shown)


src/lib/Guiguts/Utilities.pm
```diff
@@ -2619,19 +2619,81 @@ sub entry_history {
...
+#
+# Accepts the following named argument pairs:
+# -key          => key to dialog in $::lglobal hash
+# -title        => title for dialog
+# -label        => label for entry field
+# -textvariable => reference to variable paired with entry field
+# -command      => reference to routine to be executed on OK/Enter
+# -defaulttext  => optional default text in entry field that user will append to rather than typeover
+#    e.g. given "Pg " for "Pg 25" on gotopage, just "25" will be selected and user will typeover
+sub dialogboxpopup {
```
cpeel, line 2634: `dialogboxpopup` to me does not imply "This is a dialog with a single text entry field". Would something like `textentrydialogpopup` perhaps be clearer?

I don't feel strongly about this and if you are happy with the former that's fine with me too.

cpeel (commented):
Only one naming question but the code looks good!

windymilla:
It was because the DialogBox widgets that it replaced had a text entry
widget. However, you're right, a standard DialogBox widget doesn't have a
text field - I'll rename it as you suggest. Thanks

windymilla:
Should be OK now - thanks

Issue #427, opened 2020-12-30
Saving a file for a new project causes endless Message Log error messages


charliehoward4dp:
1. Started GG 1.2.0 by dragging a brand-new project file (projectID5eb476b844ad2.txt) to its desktop shortcut (my normal way of giving it something not in the recent files list).  This file is in the attached zip.
2. Made some changes, but did not assign Page Labels
3. Used Save As and named the file t00-combine illos.txt
4. GG created the file and its bin (both attached in the zip)
5. The Message Log appeared and endlessly listed the following four messages:

> 13:51:53: Use of uninitialized value in concatenation (.) or string at D:/Documents/DistrProof/PP/GG 1.2.0/lib/Guiguts/FileMenu.pm line 304.
> 13:51:54: Use of uninitialized value in concatenation (.) or string at D:/Documents/DistrProof/PP/GG 1.2.0/lib/Guiguts/FileMenu.pm line 305.
> 13:51:55: Use of uninitialized value in concatenation (.) or string at D:/Documents/DistrProof/PP/GG 1.2.0/lib/Guiguts/FileMenu.pm line 306.
> 13:51:55: Use of uninitialized value in concatenation (.) or string at D:/Documents/DistrProof/PP/GG 1.2.0/lib/Guiguts/FileMenu.pm line 307.

The "OK" (or "Close", whatever is the button label) in the Message Log did not work; the X for Close Window did not work. I could not get to GG's text window, as the Message Log kept the Focus. I was able to close everything by X-closing the Command-Line Window.

Since there was not Message Log in most of the test versions of GG 1.2, I don't know if this was happening with messages appearing in the always-minimized Command-Line window.

Restarted GG and right-clicked "Lbl" to assign Page Labels. Didn't really do so, just clicked "Use these values", then did another "Save As" using a new name. It saved the file and its bin and did NOT report any errors, so the Message Log window did NOT appear.

Update: It does not happen with GG 1.1.1. It does happen with the last Beta (12-27) and the one before that (12-18); but the 12-18 one didn't have a Message Log, so the messages just appear in the Command Line window and eventually stop (so it isn't an endless loop).

[gg12filesavebug.zip](https://github.com/DistributedProofreaders/guiguts/files/5755851/gg12filesavebug.zip)

windymilla:
[...]

PR #241, opened 2020-09-26
Remove use of `no warnings` pragma from code

windymilla:
1. Most changes are minor, e.g. checking variable is not undefined.
2. Returning via next (html_convert_tb) fixed by returning a status code to calling
routine.
3. Improved comparison function for sorting in html_convert_pageanchors to cope
with Roman numerals instead of ignoring warnings.
4. Removed unused routines in ReflowGG.pm since these used `no warnings`
5. Max lengths of status bar fields initialised at start of program.
6. spawn.pl now uses recommended `die` after exec.

Fixes #230
Fixes #218

PR #432, opened 2020-12-31
Don't attempt to write uninitialized label info to the bin file

windymilla:
Historically, uninitialized errors were suppressed with a pragma, removed by #241

Check explicitly if label has been set before outputting label information.


windymilla:
Fixes #427

PR #434, opened 2020-12-31
Stop uninitialized error when Aspell version is unobtainable

### LINKED ISSUE #429: If Aspell can't report its version, Help --> Software versions outputs an error
**Describe the bug**
There is an uninitialised variable error when checking Help-->Software Versions, if Aspell is either not installed, or the path is not correct, or it cannot report its version number.

**To Reproduce**
1. Move Aspell or configure Guiguts to look in the wrong place for Aspell
2. Open Help --> Software versions 
3. See error

**Expected behavior**
It should just say "Not available" or similar for the version number

windymilla:
When Help-->Software Versions runs but Aspell cannot report its version (due to
bad path or Aspell not installed) it returned an uninitialized value causing an error.

Instead, return "Unknown", but keep the global variable uninitialized so that the
user can correct it by using Prefs-->File Paths, then recheck Software Versions.

Fixes #429

### DIFF
--- a/src/lib/Guiguts/SpellCheck.pm
+++ b/src/lib/Guiguts/SpellCheck.pm
@@ -293,7 +293,9 @@ sub get_spellchecker_version {
     }
     close $aspell;
     unlink 'aspell.tmp';
-    return $::lglobal{spellversion} = $aspell_version;
+    $::lglobal{spellversion} = $aspell_version;
+    $aspell_version = "Unknown" unless $aspell_version;
+    return $aspell_version;
 }
 
 sub aspellstop {


src/lib/Guiguts/SpellCheck.pm
```diff
@@ -293,7 +293,9 @@ sub get_spellchecker_version {
     }
     close $aspell;
     unlink 'aspell.tmp';
```
cpeel, line 295: Just a note that at least on Linux and MacOS, `aspell --version` reports the version without bothering with a temporary file to create and delete. Not for this MR but a simplification.

PR #435, opened 2020-12-31
Allow user to choose the Compose Key

### LINKED ISSUE #430: Compose key Alt_R is already used on LOTE keyboards to type common characters
**Describe the bug**
Reported [here](https://www.pgdp.net/phpBB3/viewtopic.php?f=13&t=11466&p=1231836#p1231829), the Spanish keyboard and maybe others use the AltGr or Alt_R key to type symbols such as braces. The new Compose key feature has hijacked that.

**Expected behavior**
Allow user to configure which key is used to initiate Compose Sequences

windymilla:
1. Let key binding routine take an undef argument to unbind a key
2. Add option to Prefs menu to pop a Set Compose Key dialog
3. Dialog allows selection of Shift/Control/Alt modifiers plus a base key

Fixes #430

### DIFF
--- a/src/lib/Guiguts/KeyBindings.pm
+++ b/src/lib/Guiguts/KeyBindings.pm
@@ -6,7 +6,7 @@ BEGIN {
     use Exporter();
     our ( @ISA, @EXPORT );
     @ISA    = qw(Exporter);
-    @EXPORT = qw( &keybindings );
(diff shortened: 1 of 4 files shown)


cpeel:
This is very nicely done and works great on the Mac too.

PR #466, opened 2021-01-13
Prepare for release 1.2.3

windymilla:
Version numbers and changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,16 +3,17 @@
 ## Version 1.2.3
 
 ### Bug Fixes
-- Faulty bin file could be created, depending on page label setup
-- `/P...P/` poetry was not always rewrapped correctly 
+- faulty bin file could be created, depending on page label setup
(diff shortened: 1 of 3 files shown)


cpeel:
Bugs happen and I'm very happy in how quickly you are addressing them. Is there value in having a couple of people beta-test a version for a week or so before tagging and releasing it? I'm a bit worried that the pace of new versions for fixes might outpace the patience of some of the PPers and/or frighten people off from using new releases. On the flip side, if what is out there is particularly problematic then getting a fixed version out might be more important.

I am happy to do whatever you think is best, but I wanted to raise the discussion.

windymilla:
In these cases, I think they are definitely necessary. We should have more
of a discussion before version 1.3.n about testing. In hindsight, 1.2
didn't get enough. I'd like to go ahead with 1.2.3 probably tomorrow as
it's late today.

cpeel:
Sounds great!

PR #482, opened 2021-01-23
Fix error message and ignored orphaned brackets

### LINKED ISSUE #479: 15:46:30: Use of uninitialized value in pattern match (m//) at C:/DP/guiguts/lib/Guiguts/SearchReplaceMenu.pm line 1944.
**Describe the bug**
Orphan bracket () search result:
10:32:01: Guiguts 1.2.3: If you have any problems, please include any error messages that appear here with your bug report.
15:46:30: Use of uninitialized value in pattern match (m//) at C:/DP/guiguts/lib/Guiguts/SearchReplaceMenu.pm line 1944.
15:46:30: Use of uninitialized value in pattern match (m//) at C:/DP/guiguts/lib/Guiguts/SearchReplaceMenu.pm line 1944.
15:46:30: Use of uninitialized value in pattern match (m//) at C:/DP/guiguts/lib/Guiguts/SearchReplaceMenu.pm line 1944.
15:46:30: Use of uninitialized value in pattern match (m//) at C:/DP/guiguts/lib/Guiguts/SearchReplaceMenu.pm line 1944.
15:46:30: Use of uninitialized value in pattern match (m//) at C:/DP/guiguts/lib/Guiguts/SearchReplaceMenu.pm line 1944.

**To Reproduce**
1. Open Guiguts
2. Open the Tools=>Check Orphaned Brackets...
3. Select Bracket or Markup Style ()
4. Click Search
5. See error

**Expected behavior**
I expect it to find the first case of unpaired parenthesis if any.

**Environment**
* Guiguts version: 1.2.3
* Perl version: Strawberry Perl (64-bit) 5.30.3.1-64bit
* Perl distribution: Strawberry Perl (64-bit) 5.30.3.1-64bit

**Additional context**
Here is the file.
[hazlitt5.zip](https://github.com/DistributedProofreaders/guiguts/files/5839280/hazlitt5.zip)

windymilla:
A piece of code checked if there were four identical brackets and skipped the check
on them. This meant that the user would be warned about 3 consecutive orphaned
brackets (or indeed the 5th one of 5)
```
(a)
b)
c)
d)
(e)
```
but not about four such brackets
```
(a)
b)
c)
d)
e)
(f)
```
If the final (f) is omitted, so that the four orphaned brackets are the final ones in the
file, then an "uninitialized value" error was output.

I can't see any reason why that behaviour would be desired and have removed the
code that did it. Now all the orphaned brackets in the above examples are warned
about correctly, and no error is output if the orphaned ones are the last ones in
the file.

Fixes #479

### DIFF
--- a/src/lib/Guiguts/SearchReplaceMenu.pm
+++ b/src/lib/Guiguts/SearchReplaceMenu.pm
@@ -1952,20 +1952,6 @@ sub orphanedbrackets {
             $::lglobal{brbrackets}[1] = $::lglobal{brbrackets}[1] || '';
             last unless @{ $::lglobal{brbrackets} };
         }
-        if ( ( $::lglobal{brbrackets}[2] ) && ( $::lglobal{brbrackets}[3] ) ) {
-            if (   ( $::lglobal{brbrackets}[0] eq $::lglobal{brbrackets}[1] )
-                && ( $::lglobal{brbrackets}[2] eq $::lglobal{brbrackets}[3] ) ) {
-                shift @{ $::lglobal{brbrackets} };
-                shift @{ $::lglobal{brbrackets} };
-                shift @{ $::lglobal{brindices} };
-                shift @{ $::lglobal{brindices} };
-                shift @{ $::lglobal{brbrackets} };
-                shift @{ $::lglobal{brbrackets} };
-                shift @{ $::lglobal{brindices} };
-                shift @{ $::lglobal{brindices} };
-                brnext( $brkresult, $brnextbt );
-            }
-        }
         if ( @{ $::lglobal{brbrackets} } && $::lglobal{brindices}[0] ) {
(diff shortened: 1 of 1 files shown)


cpeel:
> A piece of code checked if there were four identical brackets and skipped the check
> on them. This meant that the user would be warned about 3 consecutive orphaned
> brackets (or indeed the 5th one of 5)
> ...
> but not about four such brackets

*Weird* -- I wonder what the original use case was for that?

windymilla:
I wonder if it was bracketing several lines together by using a brace at
the end of each to simulate a large brace in the original text. But if that
was it, it shouldn't do it for all brackets, angle speech marks etc. Also,
just doing it for 4 is odd, but I suppose if 11 lines were bracketed
together it would reduce the number of false positives from 11 to 3.

If someone wants a proper feature to detect and not report that case, they
can ask for it :)

Issue #488, opened 2021-01-25
Clicking on "0" in Word Frequency outputs an error


windymilla:
**Describe the bug**
In Word Frequency, clicking on 0 in the character list gives the following error
'''
10:49:14: Tk::Error: not a Tk object
Tk::die_with_trace at C:/Strawberry/perl/site/lib/Tk.pm line 251
Tk::__ANON__ at C:/guiguts/lib/Guiguts/SearchReplaceMenu.pm line 92
Guiguts::SearchReplaceMenu::searchtext at C:/guiguts/lib/Guiguts/WordFrequency.pm line 355
Guiguts::WordFrequency::__ANON__ at C:/Strawberry/perl/site/lib/Tk.pm line 424
(eval) at C:/Strawberry/perl/site/lib/Tk.pm line 424
Tk::MainLoop at C:\guiguts\guiguts.pl line 371
<<find>>
(command bound to event)
'''

Note, the bug is specific to "0" - it's being treated as false in a test that should have been checking for "is defined" instead of just checking the variable: `unless ($searchterm)` SearchReplaceMenu.pm:92 (and possibly elsewhere)

Also reported separately, but appears to be triggered by the above bug: subsequent double click on other characters or words always find the first occurrence in the file instead of the next occurrence. This is reset after doing a normal search with the S&R dialog.

PR #490, opened 2021-01-26
Search correctly for 0 from Word Frequency dialog

windymilla:
Searching for the character "0" from the Character Counts list in Word Frequency
caused an error message because "0" was interpreted as "false" due to incorrect
test. Bug introduced during implementation of S&R Count button in release 1.1.
Changed to explicitly check search term was empty.


windymilla:
Fixes #488

PR #515, opened 2021-02-13
Implement Greek characters in Compose Key feature

### LINKED ISSUE #183: Typing Greek from polytonic keyboard does not insert expected character
**Describe the bug**
If a Greek keyboard is installed, then either character with tonos is inserted or a question mark

**To Reproduce**
1. Open Guiguts
2. Switch to Greek keyboard
3. Type "qa" which are the keys for alpha with oxia
4. A "?" appears
5. Switch to MS Word, type "qa" - alpha with oxia appears.
6. Copy paste the character from Word to GG - character appears correctly.

**Expected behavior**
When typing "qa", alpha with oxia should appear

**Environment**
* Guiguts version: all
* OS and version: Windows
* Perl : Strawberry Perl

**Additional context**
[Useful link with oxia/tonos discussion & code points](https://www.pgdp.net/wiki/DP_Code_-_Unicode/Greek)
[Typing Greek wiki page](https://www.pgdp.net/wiki/Typing_Greek)

windymilla:
1. All Greek characters compose sequences begin with `=`
2. Simple characters just require the equivalent English letter.
3. Order of compose characters is breathing `()`, accent `'/`\`, iota subscript `|`, letter
4. About 400 Greek sequences defined to cover the vast majority of the Extended
Greek Unicode block.

Doesn't exactly fix #183 but best we can do for now given Perl/Tk Unicode limitations

### DIFF
--- a/src/lib/Guiguts/CharacterTools.pm
+++ b/src/lib/Guiguts/CharacterTools.pm
@@ -779,6 +779,30 @@ sub composeinitialize {
     composeinitcase( "\x{2021}", "\x{2021}", 'DDAG' );               # double dagger
     composeinitcase( "\x{A7}",   "\x{A7}",   'SEC', 'S*', '*S' );    # section
     composeinitcase( "\x{B6}",   "\x{B6}",   'PIL', 'P*', '*P' );    # pilcrow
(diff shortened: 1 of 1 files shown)


cpeel:
Will this be documented in the GG Manual?

PR #524, opened 2021-02-15
Update system to use perl 5.32 and perltidy version 20210111

windymilla:
Also include all perl files affected by the perltidy change


windymilla:
Abandoning for now

PR #526, opened 2021-02-15
Move to Github Actions for CI/CD

cpeel:
This moves CI/CD from TravisCI to Github Actions with identical functionality. TravisCI recently moved to a paid model and the DistributedProofreaders project is running out of the free credits. Github Actions is (at least currently) free with comparable functionality.

### DIFF
--- a/.github/workflows/perltidy.yml
+++ b/.github/workflows/perltidy.yml
@@ -1,0 +1,19 @@
+name: perltidy
+on: [push, pull_request]
+jobs:
+  perltidy:
+    runs-on: ubuntu-latest
+    container:
+      image: perl:5.32
+    steps:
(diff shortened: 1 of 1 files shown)


cpeel:
Hmm.. I don't understand why the Action wasn't run as part of the MR as it [was on the push of the branch](https://github.com/cpeel/guiguts/actions/runs/569847024). Looking into it.

cpeel:
I'm more than a little confused at how to make Github Actions work. I've updated this MR to have both TravisCI and Github Actions running in hopes that after this is merged in the latter magically start working and we can disabled TravisCI.

windymilla (approved):
I'm happy for this to go in to see if it works then.

PR #531, opened 2021-02-16
Stop using TravisCI

windymilla:
Now that Github Actions is working, no need to use Travis


windymilla:
That seems to have worked in the sense that no TravisCI check was done, and the Github Actions check worked

PR #533, opened 2021-02-16
Fix infinite loop doing Count on regex of "^"

### LINKED ISSUE #529: A regex which returns zero-length result causes Count to hang.
**Describe the bug**
A regex search for `^`, for example, will work, but clicking the Count button hangs the program

**To Reproduce**
1. Open Guiguts
2. Open the Search & Replace dialog
3. Enter the search string `^` and turn on Regex
4. Click Count - GG will hang

**Expected behavior**
Ideally, report the number of matches for the regex, which would be the number of lines in the file in this case.
Probably where it advances to search for the next one, it needs to always advance at least one character.

windymilla:
If last line of file is empty, search never fails to find regex "^" due to the way searches
wrap back to the start. Avoid this by checking if already at last character of file, i.e. 1
char from end, then forcing code to report failure to find.

Fixes #529

### DIFF
--- a/src/lib/Guiguts/SearchReplaceMenu.pm
+++ b/src/lib/Guiguts/SearchReplaceMenu.pm
@@ -112,124 +112,131 @@ sub searchtext {
     my $length = '0';
     my ($tempindex);
 
-    # Search across line boundaries with regexp "this\nand"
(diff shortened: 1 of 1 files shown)


cpeel:
+1 for Github's ability to view diffs ignoring whitespaces!

PR #535, opened 2021-02-17
Make undo work correctly after S&R inserts string terminating in newline

### LINKED ISSUE #357: Undo failure after regex search and replace
**Describe the bug**
Undo does not always restore the text correctly after a regex search/replace - long-standing bug

**To Reproduce**
1. Open [attached file](https://github.com/DistributedProofreaders/guiguts/files/5610679/undobug.zip) in Guiguts
2. Perform following regex Search & Replace:
Search: `<p><span class="pagenum"([^\n]+?)</p>\n*?<p([^\n]+?)(\n|</p>\n)`
Replace: `<p$2\n<span class="pagenum"$1$3`
Start at the top of the attached file, do the Search, then do the Replace (not S&R). The result is correct.
3. Click Undo button
4. Text is not restored, e.g. word Bullen is missing

**Expected behavior**
Undo should restore text to before Replace was executed

**Environment**
* Guiguts version: Ever since 1.0.25 or before

windymilla:
A bug in the Tk::TextUndo method insert_UNDO causes an undo to delete too much
when a replaced multi-line string ends in a newline character.

Override the method by including it in TextUnicode and adjusting the method of
calculating the length of the residue of the string after the final newline.

Fixes #357 and other cases of undoing after inserting a multi-line string that terminates in a newline

### DIFF
--- a/src/lib/Guiguts/TextUnicode.pm
+++ b/src/lib/Guiguts/TextUnicode.pm
@@ -189,6 +189,46 @@ sub ntdelete {    # no undo tracking delete
     $w->Tk::Text::delete( $start, $end );
 }
 
+# Override TextUndo::insert_UNDO to avoid bug when undoing multi-line insert in the case
+# when the string ends in a newline character. See code for details.
+sub insert_UNDO {
+    my $w      = shift;
+    my $index  = shift;
+    my $string = '';
+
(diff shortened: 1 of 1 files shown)


cpeel:
Nice comment in the function about how this differs from the original.

PR #542, opened 2021-02-19
Join footnotes correctly after user has made manual edits to file

### LINKED ISSUE #497: Footnote Fixup=>Join with previous sometimes fails
**Describe the bug**
Occasionally the second footnote in a merge is erased instead of merging with the first one.

**To Reproduce**
1. Open Guiguts
2. Open the Tools=>Footnote fixup menu ...
3. Select First Pass
4. Double click on the first pink row. (footnote #12)
5. Select "Join with Previous" (this usually works once)
6. Edit anything in the newly merged footnote.  In this case, I joined the poetry stanza by deleting four rows including the ending `*/` and the beginning `/*`.
7. Double click another pink row such as footnote #25. - **See comment below for clarification of error**
8. The footnote that was #25 and the closing bracket for footnote #24 are removed.

**Expected behavior**
I expected the two footnotes to be merged without losing data.

**Environment**
* Guiguts version: 1.2.4-beta and 1.0.25
* Perl version: Strawberry Perl (64-bit) 5.30.3.1-64bit
* Perl distribution: Strawberry Perl (64-bit) 5.30.3.1-64bit

**Additional context**
[hazlitt5.zip](https://github.com/DistributedProofreaders/guiguts/files/5896888/hazlitt5.zip)

I have noticed that if I take the trouble to save the file every time after a merge I do not have this problem.

Editing the previous merged footnote seems to be necessary for this file for this fault to occur.  I don't feel that is always the case but can't reproduce it without editing in either of the two footnoted projects I'm currently working on.

windymilla:
If the user runs Footnote First Pass, then manually edits the file before attempting to
Join Footnote to Previous, the joining fails. This is because the code stores indexes to
the line and column where the footnotes start. These indexes are shown to user in the
Footnote List dialog, but aren't safe to be used for manipulation of footnotes
when the user has edited the text.

Switch to using the permanent footnote marks in the joining routine.

Also, exit the routine immediately if user hasn't chosen at least the second footnote
to join to the previous.

Fixes #497


windymilla:
Oops - should be OK to merge now, thanks

PR #549, opened 2021-02-21
Commonise code in PageNumbers.pm

### LINKED ISSUE #548: Reduce duplicated code
PageNumbers.pm in particular has several almost duplicate subs, such as pgprevious and pgnext, pmoveup and pmovedown, pmoveleft and pmoveright. 
Reduce duplicated code either by extracting common elements or combining into single routines with an argument.

windymilla:
One set of 2 routines and one set of 4 have each been combined into one routine
that takes an argument to avoid lots of duplicate code, which did not all work
consistently with one another, e.g. some moves would not allow one marker to
"overtake" another, but others would.

PPer now has full freedom to move page markers, which is appropriate, since the
only time this dialog is used is when markers have gone wrong for some reason and
need correcting.

Fixes #548


cpeel:
Loving this diff!! 👍

PR #554, opened 2021-02-23
Remove global debug flag

### LINKED ISSUE #453: Remove debug global variable
Small bits of code output some debug info if the `debug` variable is set, but not widespread or recent enough to be useful.
Remove the variable and everything that depends on it.

Also remove debug_dump routine - too general to be useful now product is more mature.

windymilla:
Since only partially implemented in limited sections of the code, the debug flag is no
real use and the debug code just detracts from the main flow. Also the debug_dump
routine for similar reasons.

Fixes #453

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -22,8 +22,6 @@ our $VERSION = '1.2.4';
 
 # DON'T FORGET to update the version number in makefile too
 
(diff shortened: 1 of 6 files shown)


cpeel:
100% agreed.

PR #563, opened 2021-02-25
Include Extended European Latin charset accented characters in WF

### LINKED ISSUE #136: Word Frequency Accents reports only works with Basic Latin accented letters
I would like the ability to configure the Accents report in Word Frequency to work with accents other than Basic Latin, for instance with EEL or other character suites. It would be important to be able to select more than one character suite at a time.

While this could be done by specifying individual letters, I think that by character suite would make more sense.

windymilla:
In Word Frequency, the Accented filter only included characters in the Latin-1
Supplement block. Now includes DP's three Extended European Latin character sets.

Fixes #136

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -491,7 +491,7 @@ sub escape_regexmetacharacters {
 
 sub deaccentsort {
     my $phrase = shift;
-    return $phrase unless ( $phrase =~ y/\xC0-\xFF// );
+    return $phrase unless ( $phrase =~ /[$::convertcharssinglesearch]/ );
     eval "\$phrase =~ tr/$::convertcharssinglesearch/$::convertcharssinglereplace/";
     $phrase =~ s/([$::convertcharsmultisearch])/$::convertcharssort{$1}/g;
     return $phrase;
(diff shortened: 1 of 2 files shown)


src/lib/Guiguts/Utilities.pm
```diff
@@ -534,8 +534,27 @@ sub readlabels {
     $::convertcharsdisplaysearch = join( '', keys %::convertcharsdisplay );
     $::convertcharsmultisearch   = join( '', keys %::convertcharssort );
 
-    $::convertcharssinglesearch  = "��������������������������������������������������������������";
-    $::convertcharssinglereplace = "AAAAAAAaaaaaaaCcDdEEEEeeeeIIIIiiiiOOOOOOooooooNnsTtUUUUuuuuYyy";
+    # Contains accented characters from Latin-1 Supplement block and DP's Extended European Latin character suites
```
cpeel, line 537: You might consider adding a comment pointing to the character suites in question for future explorers, such as either in the code:
* https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/charsuite-extended-european-latin-a.inc
* https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/charsuite-extended-european-latin-b.inc
* https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/charsuite-extended-european-latin-c.inc

Or the UI:
* https://www.pgdp.net/c/tools/charsuites.php?charsuite=extended-european-latin-a
* https://www.pgdp.net/c/tools/charsuites.php?charsuite=extended-european-latin-b
* https://www.pgdp.net/c/tools/charsuites.php?charsuite=extended-european-latin-c

srjfoo: Is there a reason for including only EEL a, b and c? Excluding the Greek character suites, there's the Medievalist supplement and Semitic and Indic transcriptions (assuming you don't want the Symbols collection).

windymilla: The original request, #136, led me down the character suite route, but looking at it afresh, I think it makes more sense to just include all accented Latin characters in the Latin Extended A, Latin Extended B, and Latin Extended Additional Unicode blocks. This would cover all the relevant characters from the Medievalist and Semitic/Indic transcriptions, and would future-proof against another unusually accented letter from these blocks being included in a new character suite, such as Ḁ or Ẅ. Just 700 or so numbers and letters to type/match up :)

cpeel: I feel like there has to be a transliteration function that will decompose the characters down into Latin-1 so it doesn't have to all be done by hand. Maybe something like https://metacpan.org/pod/Text::Unidecode ?

windymilla: Thanks. I've looked at that previously - it's a "transliterator of last resort", so may not do exactly what we want - would probably take as long to check out as to do manually. Also, and more importantly, I only want to do accented characters, not eng or long-s or other oddities that live in these blocks, so I have to know which specific characters they are. I exaggerated when I said 700, it's just under 600 in those 3 blocks and some are not accented. I quickly made the base blocks with the hex digits in using GG regex, and am just going through typing the unaccented equivalent, which is pretty quick. I've finished the first block.

srjfoo: Definitely a good idea not to tie it to any one character suite.

windymilla:
I think that does it. The characters are generally capital followed by small, with a notable exception being the start of Latin block B.

PR #571, opened 2021-02-27
Pass title and author to locally-run ebookmaker

### LINKED ISSUE #168: Pass Title and Author to EBOOKMAKER
When invoking ebookmaker, include --title TITLE --author AUTHOR  if they are available, as they normally are in the <title> section of the `<head>`.

Why: These can (and should) be included when running online ebookmaker, and including them with the Local version will allow us to upload MOBI/EPUB versions for Smooth-reading without using the online version. That'll be quicker & easier (the main reasons) and infinitesimally reduce the load on PG's servers.

windymilla:
Extract title and author from `<title>` element in HTML file and pass as arguments
to ebookmaker. This causes the epub/Kindle files to be named with the book title
and the title and author are also stored in the files, so they show correctly on the
reader's e-reading device. This is helpful for creating smooth-reading versions.

#Fixes #168

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -2053,7 +2053,7 @@ sub ebookmaker {
         my $tstring = $textwindow->get( $tbeg . '+7c', $tend );    # Get whole title/author string
         $tstring =~ s/\s+/ /g;                                     # Join into one line, single spaced
         if (
-            $tstring =~ s/The Project Gutenberg EBook of//i         # Strip PG part - 2 formats
+            $tstring =~ s/The Project Gutenberg EBook of//i        # Strip PG part - 2 formats
             or $tstring =~ s/&mdash;A Project Gutenberg eBook//i
         ) {
             HTML::Entities::decode_entities($tstring);             # HTML entities need converting to characters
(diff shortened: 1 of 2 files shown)


src/lib/Guiguts/Utilities.pm
```diff
@@ -2043,6 +2043,34 @@ sub ebookmaker {
...
+    # Get title and author information
+    my $ttitle  = $fname;      # Title defaults to base filename
+    my $tauthor = 'Unknown';
+    my $tbeg    = $textwindow->search( '-exact', '--', '<title>',  '1.0', '20.0' );
+    my $tend    = $textwindow->search( '-exact', '--', '</title>', '1.0', '20.0' );
+    if ( $tbeg & $tend ) {
+        my $tstring = $textwindow->get( $tbeg . '+7c', $tend );    # Get whole title/author string
+        $tstring =~ s/\s+/ /g;                                     # Join into one line, single spaced
+        if (
+            $tstring =~ s/The Project Gutenberg EBook of//         # Strip PG part - 2 formats
```
cpeel, line 2056: Do we want `//i` here to ignore case? I'm looking at `EBook` and `eBook` and shaking my head...

windymilla: Good idea - whatever case the PPer has used, we want to strip the phrase :)

windymilla: OK - done now, thanks

PR #572, opened 2021-02-27
Convert all accented characters during "display" conversion

### LINKED ISSUE #569: Deaccent for display doesn't handle all accented characters
**Describe the bug**
#563 included all Latin accented characters in "sort" deaccenting, but "display" deaccenting should also use the same mapping to clean up any remaining accented characters once the language-specific work has been done.

windymilla:
If an accented character was not handled by the language-specific conversions for
display purposes, then only the Latin-1 supplement characters were also converted.

Changed to include all Latin characters from the Extended and Additional blocks.
Also ordered and split list of converted characters in the Latin-1 Supplement block
to match the handling of the other blocks.

Fixes #569

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -37,6 +37,7 @@ use File::Spec::Functions qw(catdir);
 use File::Copy;
 use File::Compare;
 use File::Which;
+use HTML::Entities;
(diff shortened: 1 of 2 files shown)


windymilla:
Ignore - git user error!

PR #591, opened 2021-03-04
Provide fraction conversion functionality

### LINKED ISSUE #169: html and text fraction converter
Currently, GG's html generator has a function which converts some, but not all fractions in the DP rounds format (e.g. 1-1/2) to html entities (if "Convert fractions" is selected prior to generation). There is no equivalent for text files.

I would find helpful a revised fraction converter, which:

a) converts a wider range of fractions, ideally all possible fractions;

b) can be used in text and/or html files.

This tool could produce:

a) unicode fractions (e.g. ⅓), where they exist; or

b) fractions using the superscript/subscript format (e.g. ¹⁄₃), which potentially covers all possible fractions.

Ideally the user could choose:

a) only unicode fractions, and leave all others in DP rounds format;

b) only superscript/subscript format;

c) unicode fractions where possible, falling back on superscript /subscript format where unicode fractions do not exist.

windymilla:
1. Converts fractions in DP format such as 2/3 or 1-1/4 to Unicode characters
2. Three options:
a. Only convert those that have a single Unicode character representation - leave
others in DP format
b. Convert all possible to a single Unicode character and use superscripts and
subscripts for the rest.
c. Use superscripts and subscripts for all.
3. Accepts either the Fraction Slash or standard slash characters as input.
4. Always uses Fraction Slash on output with subscripts and superscripts.
5. If character preceding match is an alphabetic, then don't convert the fraction,
to avoid altering strings such as ABC1/2 or A-B-1/2, which are more likely a
code number, serial number, etc. In maths, there would normally be a space, e.g.
"A - B - 1/2" - this fraction will be converted.

#Fixes #169

### DIFF
--- a/src/lib/Guiguts/CharacterTools.pm
+++ b/src/lib/Guiguts/CharacterTools.pm
@@ -9,7 +9,7 @@ BEGIN {
     our ( @ISA, @EXPORT );
     @ISA    = qw(Exporter);
     @EXPORT = qw(&commoncharspopup &utfpopup &utfcharentrypopup &utfcharsearchpopup &cp1252toUni
-      &composepopup  &composeinitialize &composeref);
+      &composepopup  &composeinitialize &composeref &fractionconvert);
 }
 
 #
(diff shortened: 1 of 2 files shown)


cpeel:
Very nicely done 👍

PR #595, opened 2021-03-06
Provide hanging indent rewrap option for ASCII Table Effects

### LINKED ISSUE #366: ASCII Table Effects: add support for hanging indents in selected column(s)
Usually, multi-line text cells in columns are printed as hanging indents (1 or 2 leading spaces on continuation lines). Adding an option to do this for the currently-selected column, while adjusting that column's width, would save pp'ers a lot of time in preparing the Plain Text versions of projects containing such tables ... if they know how useful ASCII Table Effects already is.

windymilla:
1. Adds field to ASCII Table Effects dialog, which accepts postive or negative integer
hanging indent (negative interpreted as normal indent)
2. Other contents of dialog reorganised to make association between various
elements more obvious.
3. Resizing of columns more robust - does not error when column shrunk too small,
and no blank line below.
4. Table Right Column field has validation and does not error if illegal value entered.
5. Selection of Next/Previous column made more robust - does not error. If Next is
chosen when already at end, wraps back to start, and vice-versa.
6. Bug with non-wrapped column adjustment fixed to avoid affecting line below the
table.

Fixes #366

### DIFF
--- a/src/lib/Guiguts/ASCIITables.pm
+++ b/src/lib/Guiguts/ASCIITables.pm
@@ -94,8 +94,8 @@ sub tablefx {
         }
         my $f1 = $::lglobal{tblfxpop}->LabFrame( -label => 'Adjust Column' )
           ->pack( -side => 'top', -anchor => 'n', -expand => 'yes', -fill => 'x' );
-        my $f1a = $f1->Frame->pack( -side => 'top', -anchor => 'n' );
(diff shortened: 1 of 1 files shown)


windymilla:
Made a small improvement after query from @srjfoo and suggestion from @charliehoward4dp
Hope it doesn't mess with your testing - thanks

PR #599, opened 2021-03-07
Change Mac packaging/install and README for Jeebies

tangledhelix:
# Background

On macOS, browsers mark a file as being downloaded from the internet. This is true even for files downloaded inside a .zip file, like the files in Guiguts.

For executables such as Jeebies, being marked in this way has two consequences.

1. When run, the system will ask "this came from the internet, are you sure you want to run it?"
2. The executable must be signed by a developer certificate

On more recent versions of macOS, the signing requirement is an absolute; if the executable isn't signed, it cannot run.

Note that the thing doing this marking is the browser.

# Changes in this PR

This PR adds an arm64 (M1) build of Jeebies, and also a script to download and install Jeebies. The script detects the CPU architecture (x86_64 or arm64) and downloads the correct file.

README.md and package.sh were also modified to reflect these changes.

The reason this works is that curl doesn't mark downloaded files the way that browsers do; as a result, the certificate signature requirement is no longer in play.


tools/jeebies/README.md
```diff
@@ -5,8 +5,16 @@ https://www.pgdp.net/wiki/PPTools/Jeebies
 
 ## Packaging
 
-We include pre-built binaries for Windows and MacOS. For other platforms we
-ship the files necessary to build Jeebies.
+We include pre-built binaries for Windows.
+
+On macOS, we can provide pre-built binaries, but not in the .zip download of Guiguts. This is because it's an unsigned executable program, and anything you download using a browser (including files inside a .zip), must be signed in order to run on modern versions of macOS.
```
cpeel, line 10: The signing requirement is only for arm64 binaries -- the x86_64 ones should not require signing based on my reading and they should still run through Rosetta (at a minimum they don't for those of us not running Big Sur). We should continue to bundle the x64_64 binaries and provide a way for users to build the arm64 version if they want.

Please be sure and wrap Markdown files at 80 characters.

cpeel (commented):
GitHub is showing `tools/jeebies/jeebies-mac-arm64` as a zero-length file which seems wrong.

tangledhelix:
> GitHub is showing `tools/jeebies/jeebies-mac-arm64` as a zero-length file which seems wrong.

When I look at Github it tells me it's a 67.2kb executable file... where do you see zero length?

https://github.com/tangledhelix/guiguts/blob/jeebies-arm64/tools/jeebies/jeebies-mac-arm64

tools/jeebies/README.md
tangledhelix: My concern with providing a way to build is that most users won't have a compiler, so that requires they install one for this compile step. Presumably this concern is the same reason we distribute the x86_64 binary today rather than only source code.

Also, my testing disagrees with your suggestion it's only an issue for arm64 binaries. I just downloaded the jeebies-mac binary from Github using Chrome, and it gives me the unsigned executable error. So does the jeebies executable downloaded via the .zip file.

It's also not an arm64 chip thing (I get the same error on an x86_64 iMac), nor a Big Sur thing (I get the same error on x86_64 Catalina).

cpeel: > Well bugger. This solidifies that I'm going back to Linux for my next laptop when my 2015 MBP running Catalina dies.

Our MacOS instructions tell the user to install XCode, so if they follow the instructions they'll have a compiler but having a way to download a prebuilt one is better I think.

So overall I think your approach is good, but suggest that instead of having two binaries we just build a Universal one and keep the filename the same. This also means we need to update the `INSTALL.md` file and tell the user to run this script

cpeel:
Also ensure your commit message follows the git best practices of being wrapped at 72 characters.

tools/jeebies/README.md
tangledhelix: 🤦 - I completely overlooked that Homebrew requires the compiler tools already...

I'll look into how to build a universal binary.

srjfoo:
> > GitHub is showing `tools/jeebies/jeebies-mac-arm64` as a zero-length file which seems wrong.
> 
> When I look at Github it tells me it's a 67.2kb executable file... where do you see zero length?
> 
> https://github.com/tangledhelix/guiguts/blob/jeebies-arm64/tools/jeebies/jeebies-mac-arm64

I'm seeing it, too:
image

tangledhelix:
I'm not sure there's a point to distributing a binary at all, even via a download script, if the user has already installed the compiler tools... we can't simply include the binary, due to the code signature problem, so the user has to do something either way.

1. Run a build
2. Run a download script

So providing the binary in the download doesn't gain anything (the user has to run a terminal command either way). And the build option has no significant barrier to entry, as the user was already instructed to install the Xcode build tools.

I don't run this project, so I'll defer to @cpeel and @windymilla - if you feel we should ship the universal binary and a download script, as well as the Makefile to build the binary, then I'll arrange it that way. If you agree that there's little point to shipping a binary, and we should just have the user run make, then I'll arrange it that way. What are your thoughts?

See latest commit - the Makefile has been adjusted to build either intel or arm binaries (if desired), but by default it will build a universal binary. I also added  a `clean` target.

tangledhelix:
I also see the empty file thing now, if I look in this PR. But it's not showing empty in my repository... strange.

In any case, that file won't matter anymore, as things have shifted and we'll either ship a universal binary (under different name) or no binary at all. So I'll ignore it and await feedback from Casey and Nigel about that question. That file path won't end up shipping no matter what.

cpeel:
I think having them build the binary themselves is probably the best solution -- like we do for Linux. Speaking of Linux, note that the `Makefile` needs to work there as well (and possibly for Windows? @windymilla?).

What I'd like to see happen is:
1. Remove the `elif [[ $OS == "mac" ]]; then` in `tools/jeebies/package.sh` and the MacOS binary so we fall back to copying over the source code like we do for everything else (i.e. Linux).
2. We probably(?) don't need the `Makefile` changes if we're having users build their own anyway.
3. Keep your `README.md` changes with the details on why we don't ship MacOS binaries (with slightly updated wording)
4. Add a `tools/build-tools.sh` script that we include with non-Windows packaging that will build the appropriate tools and update the `INSTALL.md` file with instructions for MacOS users to run it. This should probably basically be a copy of `tools/package-tools.sh` except that it finds `Makefiles` and runs them. `tools/package-tools.sh` should likely be the one that copies over the new `build-tools.sh` script for non-Windows OSs.

I'm happy to do some or all of 4 depending on how much you want to do since that's way more involved.

I appreciate you opening this MR and helping us understand the challenges with what we're shipping -- to my knowledge you're the first person to _use_ GG on Mac. Thank you for helping make this better!

windymilla:
@cpeel, I'm happy with all of what you suggest. I don't think the Makefile does need to work on Windows, since we just copy jeebies.exe when packaging the Windows release. Windows users have no need to rebuild jeebies at all.

tangledhelix:
Since @cpeel and @windymilla seem to agree here, I'll work on those changes and update this PR.

tangledhelix:
Pushed updates with Casey's suggestions. Let me know if I missed something...

cpeel:
This looks excellent to me, thank you. @windymilla any comments / feedback before I merge this in?

windymilla:
I tried making a win, mac and generic release file on my Windows system, but I am no longer able to. Although I think `make win` runs fine, both `make mac` and `make generic` give the following output:
```
$ make generic
rm -rf guiguts
mkdir guiguts
# Start with src/
cp -a src/* guiguts
# Remove untracked files & directories that might be in src/
rm -rf guiguts/tools/ guiguts/header.txt guiguts/setting.rc guiguts/data/labels_en.rc
# Copy common tools
cp -a CHANGELOG.md INSTALL.md UPGRADE.md LICENSE.txt README.md THANKS.md guiguts
# Build tools
mkdir guiguts/tools
./tools/package-tools.sh generic $(pwd)/guiguts/tools
cp: cannot stat `build-tools.sh': No such file or directory
make: *** [generic] Error 1
```

I'm guessing somewhere it is confusing whether the platform of my machine with the platform of the release I am building for, but I don't know.

tangledhelix:
@windymilla Can you try packaging again? I made one update so that the full path to `build-tools.sh` is now specified. I didn't take the top-level Makefile into account and didn't realize this would be run from outside of the `tools/` directory.

windymilla:
Thanks Dan. I'm afraid it still didn't work, but looks like it got a bit further:
```
$ make generic
rm -rf guiguts
mkdir guiguts
# Start with src/
cp -a src/* guiguts
# Remove untracked files & directories that might be in src/
rm -rf guiguts/tools/ guiguts/header.txt guiguts/setting.rc guiguts/data/labels_en.rc
# Copy common tools
cp -a CHANGELOG.md INSTALL.md UPGRADE.md LICENSE.txt README.md THANKS.md guiguts
# Build tools
mkdir guiguts/tools
./tools/package-tools.sh generic $(pwd)/guiguts/tools
./tools/package-tools.sh: line 21: TOOLS_DIR: command not found
cp: cannot stat `/build-tools.sh': No such file or directory
make: *** [generic] Error 1
```

PR #616, opened 2021-03-13
Improve curly quote conversion

windymilla:
1. Spot and convert ditto marks (2 spaces each side)
2. Force quotes at start of line to be open quotes
3. Force quotes at end of line to be close quotes
4. Add Curly Quote Corrections submenu - tear-off for easy use
5. Allow selection of next line with @ symbol - other correction options work on
selected region, so user can alter it if needed
6. Correction option 1: Flip all double quotes in selected region - once quotes are
out of sync, every quote for the rest of the paragraph often needs flipping.
7. Correction option 2: Use spacing to decide quotes in selected region, e.g. if a
quote is preceded by a space and followed by a non-space it must be an open quote.
8. Allow removal of all `@` signs in the selected region
9. Allow insertion of the four types of quote.

### DIFF
--- a/src/lib/Guiguts/MenuStructure.pm
+++ b/src/lib/Guiguts/MenuStructure.pm
@@ -538,8 +538,9 @@ sub menu_txt {
             'command', "Auto-Convert ~Options...", -command => sub { ::text_convert_options($top); }
         ],
         [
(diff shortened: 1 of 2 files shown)


windymilla:
Fixes #618

PR #650, opened 2021-03-27
Improve default selection of path to tools

### LINKED ISSUE #649: Improve transfer of settings during release upgrade
Paths to tools are stored in the setting.rc file. If this is copied to a new release, it can cause problems.

If the tool is under the tools folder of the release, and the saved version in the setting.rc is in a similar folder, it would be better to just use the relative path to the tool instead of the absolute path to the old release.

windymilla:
When upgrading from a previous release, if the user copied across their setting.rc,
it contained the path to each tool in the old release, which could cause problems
if the old release was subsequently deleted, or became out of date.

Now, if the tool lies under a `tools` folder, and the user's saved path looks the same
as the new suggested path under the current release, e.g. both end with
`tools\ebookmaker\ebookmaker.exe`, then use this relative path to the tool instead
of the absolute path.

Fixes #649

### DIFF
--- a/src/lib/Guiguts/Preferences.pm
+++ b/src/lib/Guiguts/Preferences.pm
@@ -15,16 +15,23 @@ BEGIN {
 
 #
 # Find path to tool, either from suggested default path, or on the system path.
-# If user already has a path set up for this tool, then we may not want to override it.
-# However, we do if the path is in the "same" location under another previous release.
+# If user already has a path set up for this tool, then we may not want to override it,
+# since perhaps it is a modified or updated version they keep on their system.
+# However, a frequent reason for the user having a path already in their setting.rc
+# is if they copied the setting.rc from another installed release, for example when they
+# upgraded. In that case, the tail end of the saved path will match the tail end of
+# the default path for the tool, e.g. "...tools\jeebies\jeebies.exe". In this case,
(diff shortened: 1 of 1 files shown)


src/lib/Guiguts/Preferences.pm
```diff
@@ -13,23 +13,46 @@ BEGIN {
       &saveinterval &reset_autosave &setcolor &locateExecutable &filePathsPopup &setDPurls &composekeypopup);
 }
 
+#
+# Find path to tool, either from suggested default path, or on the system path.
+# If user already has a path set up for this tool, then we may not want to override it.
+# However, we do if the path is in the "same" location under another previous release.
```
cpeel, line 19: Something that is implied is that "previous release" == "different directory". Since the problem we're solving is because of the second, not the former, consider updating the comment to be clearer.

I was originally quite confused on what the Issue and this MR were solving until I realized that it's not actually about the release as it is about the directory names.

windymilla: As usual, it seemed clear when I wrote it, but now less so, and obviously not clear to anyone but me, which makes it useless. I'll update it tomorrow as it's getting late now. Thanks

windymilla:
OK - that should be clearer, I hope :)

PR #652, opened 2021-03-27
Prepare for release 1.3.0-beta

windymilla:
Version number and changelog


windymilla:
Should be OK for merging and 1.3.0-beta will be ready to consider for release - thanks

INSTALL.md
```diff
@@ -40,27 +40,38 @@ _Note that you must do step 6, even if you have done it previously, as
...
 8. See the [Guiguts Windows Installation](https://www.pgdp.net/wiki/PPTools/Guiguts/Install)
-   wiki page for information on installing the Aspell spell checker and an image
-   viewer to display scans and edit images.
+   wiki page for information on installing the Aspell spell checker and an
+   image viewer to display scans and edit images.
+
+_Only if step 6 above was unsuccessful,_ then do the following to re-associate
+`.pl` files with the Perl program/app. These instructions may vary slightly
+for different versions of Windows:
+ * Close Notepad if it is running.
```
cpeel, line 67: Since these are ordered steps we should have this be a numbered list instead of a bulleted one.

cpeel (commented):
Nice addition of the strawberry-perl-before-GG-download reorder.

INSTALL.md
windymilla: The reason I didn't was that I thought it could be confusing with the numbered steps in the main sequence. For example, step 6 in the additional steps would say
6. Now return to reattempt step 6.
I looked into whether markdown supported a, b, c, etc or Roman numerals, but I can't see that it does.
Any ideas?

cpeel: Hmm... yes, I see that being confusing. If this is all isolated to step 6, perhaps make this a sublist to step 6 so that it's proximal to the problem?

windymilla: Yes, that's how it was before, but it got so long that it seemed to detract from the sequence of main steps, which seemed excessive given that shouldn't actually be needed by anyone. 

It would look like this:

6. Double click the file `install_cpan_modules.pl`. This should display a
   command window listing the Perl modules as it installs them. Note that this
   can take several minutes to complete.
   If instead, Windows says it does not know how to run that file, or it opens
   the file in a text editor like Notepad, follow the additional instructions
   below to re-associate `.pl` files with the Perl program/app. These
   instructions may vary slightly for different versions of Windows:
     * Close Notepad if it is running.
     * Right-click `install_cpan_modules.pl` in File Explorer, and choose
       `Open with`.
     * Choose `More apps` (may say "Choose Default Program" on some systems)
     * Scroll to the bottom then choose `Look for another app on this PC`
       (may say "Browse" on some systems)
     * Navigate to `c:\Strawberry\perl\bin` and choose `perl.exe`.
     * Now return to reattempt the start of step 6.
7. Double click the `run_guiguts.bat` file in the same folder,...

Although if I numbered the sublist the rendered form would use Roman numerals, the raw form would still have to use regular numbers which would be very confusing and is probably how most users would see it.

I'll have a think while I'm eating now - if you have another good idea, I'm happy to go with it.

cpeel: Ugh, yeah, that's a lot of extra mess in the main path for an in-theory rare edgecase. Perhaps keep this here with a numbered listing and have your last one be:
```
Now return and re-attempt the "Double click the file `install_cpan_modules.pl`" step.
```
? That would remove the awkward step 6 referring to step 6 bit (which, as a bonus, we don't have to remember to update the `6` if the numbered steps above change).

If you'd rather keep what you have, I'm OK with that too.

srjfoo: I don't know if it would help, or make things even more confusing, but how about changing ...
>If instead, Windows says it does not know how to run that file, or it opens the file in a text editor like Notepad, follow the additional instructions below to re-associate `.pl` files with the Perl program/app. These instructions may vary slightly for different versions of Windows: 

to

> If instead, Windows says it does not know how to run that file, or it opens the file in a text editor like Notepad, you will need to re-associate `.pl` files with the Perl program/app.[*]

And then add a separate numbered list as a footnote -- or perhaps in a separate troubleshooting section?

PR #681, opened 2021-04-01
Stop bundling kindlegen with Guiguts releases

### LINKED ISSUE #208: Remove kindlegen binary and point people to KindlePreviewer
Frustratingly, Amazon pulled the stand-alone [kindlegen download](https://smile.amazon.com/gp/feature.html?ie=UTF8&docId=1000765211&sa-no-redirect=1) and instead are pointing people to [KindlePreviewer](https://smile.amazon.com/gp/feature.html?ie=UTF8&docId=1000765261&sa-no-redirect=1). This breaks packaging for Windows and Macs. The Previewer includes kindlegen somewhere in it, but the Previewer is an installation executable and we can't extract the binary for packaging.

We should stop trying to package kindlegen in GG and instead point people to download and install KindlePreviewer. We can make this slightly less painful by adding the (presumptive) kindlegen path to the GG batch file.

windymilla:
Now Amazon don't supply kindlegen as a standalone executable, user must install
Kindle Previewer 3 and tell Guiguts where kindlegen is if they want ebookmaker to
produce .mobi files.

If kindlegen is not set up, ebookmaker will just make epub files (and give a helpful
message)

Also, set default install path for kindlegen to be where Kindle Previewer 3 is
installed by default on Windows.

Also, ebookmaker was not using the TIdy path set up by the user, but code was
assuming it was under `...tools/tidy`.

Fixes #208


src/lib/Guiguts/Utilities.pm
```diff
@@ -1114,6 +1115,14 @@ sub initialize {
         ::catfile( '\Program Files', 'Aspell', 'bin', 'aspell.exe' ) );
     $::validatecommand = ::setdefaultpath( $::validatecommand,
         ::catfile( $::lglobal{guigutsdirectory}, 'tools', 'W3C', 'onsgmls.exe' ) );
+    $::kindlegencommand = ::setdefaultpath(
```
cpeel, line 1118: FWIW, the default path on a Mac is: `/Applications/Kindle Previewer 3.app/Contents/lib/fc/bin/kindlegen` if we want to include that based on OS.

cpeel (commented):
Do we want to update `INSTALL.md` and mention the need to download Kindle Previewer 3 with a link to it now that we don't include the `tools/kindlegen/README.md`? (e.g. https://www.amazon.com/gp/feature.html?docId=1000765261)

windymilla:
Good idea. I'll update tomorrow

windymilla:
Should be OK now - thanks.

src/lib/Guiguts/Utilities.pm
```diff
@@ -1115,6 +1116,26 @@ sub initialize {
     $::validatecommand = ::setdefaultpath( $::validatecommand,
         ::catfile( $::lglobal{guigutsdirectory}, 'tools', 'W3C', 'onsgmls.exe' ) );
 
+    if ($::OS_MAC) {
+        $::kindlegencommand = ::setdefaultpath(
+            $::kindlegencommand,
+            ::catfile(
+                ' Applications',
```
cpeel, line 1123: You have an extra space here in the string.

windymilla:
Doh! Thanks for your careful reviews as ever!

charliehoward4dp:
Perhaps the "Install.md" instructions should offer the alternative suggestion of locating and using a copy of kindlegen.exe in an existing, earlier installation of Guiguts. That only will be helpful to people with those earlier versions.  Any copy of kindlegen.exe can be copied to a permanent user folder, making it available to future versions of Guiguts while protecting it from uninstalls of older versions.

PR #685, opened 2021-04-02
Fix errors revealed by test suite

windymilla:
1. Bug introduced into wrapping - occasionally adds a blank line after a block, due
to error in extending end of block to end of line - should go to start of next line,
but only if not already at start of line. Caused by #665. 
2. Bug introduced into de-accenting for display when separating out Latin1
characters into their own array for sorting speed up. Caused by #667. 
3. Baseline test files needed updating now that error display dialog has its own
count of active errors, so error list doesn't need it. Caused by #597.


cpeel:
Yay for test suites!

windymilla:
Indeed - both of those could have remained hidden for some time.

PR #713, opened 2021-04-16
Remove test files from release package

### LINKED ISSUE #700: Remove unwanted files from distributed release and/or whole system
Some files may want to remain within Guiguts, but not be distributed. Some might go completely
1. scannos/GutWrenchGutAxe folder?
2. tests folder?
3. tools/bookloupe/*.tst files?

windymilla:
1. Move the GG test directory from under src to the directory above so it does not
get included. Edit the test code to cope with the new location
2. Remove the bookloupe test program and test files when packaging it
3. Fix typo in bookloupe README

Note that the GutWrenchGutAxe scannos were unknown to PPers, but once they
knew they existed, they didn't want to lose them, so this commit leaves them alone.

Fixes #700


tests/epubfriendly.html
```diff
@@ -1,76 +1,76 @@
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
```
cpeel, line 1: I'm confused on why these files are listed as changed -- apparently by whitespace? Did the line endings change? How did you move the files, `git mv src/tests .`?

cpeel (commented):
> Note that the GutWrenchGutAxe scannos were unknown to PPers, but once they knew they existed, they didn't want to lose them, so this commit leaves them alone.

I'm not sure if that's a reason to leave them or a reason to remove them. Such hoarders! 😁

tests/epubfriendly.html
windymilla: No, I just moved the folder in Windows. Was that wrong?

cpeel: That should have been fine. I'm not too worried about it as viewing the diffs without whitespace shows no changes. Mostly curious.

windymilla: Thanks

PR #716, opened 2021-04-18
Highlight characters in WF that are not in an enabled character suite

### LINKED ISSUE #691: Add additional CP support to WF Character Count

1. Maintain a list of selected DP character suites that can be popped from the CP menu.
2. Option in CP menu to toggle highlighting of characters in WF Character Count that are not in the selected suites.
3. Allow suggestion of additional char suite to select, e.g. by Ctrl-clicking highlighted character in list. Maybe pops a OK/Cancel dialog asking "Add XYZ char suite to selected list?" if character is in a char suite. Pop warning "Character not in any known char suite" if not.

windymilla:
1. List of available character suites is available via the File->CP menu.
2. User can enable/disable them (except Basic Latin).
3. Check button in File->CP menu controls highlighting of any characters in the WF
Character Count list that are not in an enabled charsuite.
4. Control-clicking a highlighted line either pops a dialog asking if the user wants to
enable the first charsuite alphabetically that contains the character, or if there is no
such charsuite a warning dialog is popped
5. A bug (from 1.0.25) where the user can control-click in a word listing in the WF
dialog to display 1st harmonics has been fixed (missing argument to routine)
6. An undocumented broken feature (again since 1.0.25) to add a word to the project
dictionary with control-mouse 2 or 3, has been removed.

Fixes #691

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -116,8 +116,9 @@ our $blocklmargin     = 1;
 our $blockrmargin     = 72;
 our $poetrylmargin    = 4;
 our $blockwrap;
-our $booklang          = 'en';
-our $composepopbinding = 'Alt_R';               # Default key to pop the Compose dialog (Right hand Alt key, also labelled AltGr)
+our $booklang             = 'en';
+our $charsuitewfhighlight = 0;                  # Don't do charsuite availability highlighting in WF dialog
+our $composepopbinding    = 'Alt_R';            # Default key to pop the Compose dialog (Right hand Alt key, also labelled AltGr)
 $composepopbinding = 'Control-m' if $OS_MAC;    # Default to Ctrl+m on a Mac - Alt+RightArrow does the same indent operation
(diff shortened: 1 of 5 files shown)


cpeel:
On the Mac if I go to File -> Content Providing -> Manage Character Suites, the resulting dialog is so small as to be unreadable and I can't resize it. I know we fixed this for some other dialogs, looks like we need to fix that here too.
Screen Shot 2021-04-21 at 2 56 04 PM

windymilla:
Should be OK, though I note it had/has conflicts. Do I need to re-do it in some way? I think it's because there are two new entries in the File->CP menu, so the code changes are adjacent.

cpeel:
The Mac fix works, thanks! Please rebase this off the latest master and I'll get this merged in.

windymilla:
Hopefully I've done that right - the program seems to behave correctly, so if the code looks OK to you, it can be merged.

PR #726, opened 2021-04-21
Include File::HomeDir as a package dependency

cpeel:
The homebrew-provided perl doesn't include `File::HomeDir` by default, so we need to be explicit about having it installed.


windymilla (approved):
Looks good and works on my computer (saying HomeDir was up to date). Thanks

PR #729, opened 2021-04-24
Optimise status bar updating

### LINKED ISSUE #720: Bookloupe error message
**Describe the bug**
13:15:56: Guiguts 1.3.0: If you have any problems, please include any error messages that appear here with your bug report.
13:19:23: Tk::Error: Tk::Toplevel=HASH(0x8f94200) is not a widget at C:/DP/guiguts/lib/Guiguts/Utilities.pm line 1599.
 Tk callback for .toplevel1.frame.frame.button1
 <ButtonRelease-1>
 (command bound to event)
13:19:23: Tk::Error: Usage $widget->update(...)
Tk::Frame=HASH(0x8f7c760) is not a Tk object at C:/DP/guiguts/lib/Guiguts/ErrorCheck.pm line 388.
 Guiguts::ErrorCheck::errorcheckpop_up at C:/DP/guiguts/lib/Guiguts/ErrorCheck.pm line 388
 Guiguts::KeyBindings::__ANON__ at C:/DP/guiguts/lib/Guiguts/KeyBindings.pm line 40
 <Key-F6>
 (command bound to event)

**To Reproduce**
1. Open Guiguts
2. Press F6
3. Hide all except for Short lines
4. Close the Bookloupe window (either by clicking the x in the upper right corner or Alt+F4)
5. See error message

**Expected behavior**
I didn't expect to see the error message.  

Also, I noticed the Perl interpreter took less than 1% of the CPU resources until Bookloupe was started then it went to 24% of CPU and stayed that way as long as the Bookloupe window was open.  For comparison, the WF window does the same temporarily then after the processing was done Perl's drain on the CPU dropped to less than 1% with the WF window still open.  I don't know if this has any bearing on anything but thought I'd mention it.  The reason I was checking the resources was that I was seeing a slow down in GuiGuts and other programs.

**Environment**
* Guiguts version:  1.3.0
* Perl version: Strawberry Perl (64-bit) 5.32.1.1-64bit
* Perl distribution: Strawberry Perl (64-bit) 5.32.1.1-64bit

**Additional context**
[13949.zip](https://github.com/DistributedProofreaders/guiguts/files/6338931/13949.zip)

I closed and reopened GuiGuts and still get the same error message.

windymilla:
The status bar is updated every 200 milliseconds. The update routine can take more
than 200ms to run under some circumstances. The slow part is finding the current
page number when in a large file without page numbers and lots of marks have
been inserted by a checking tool. The code takes time going through those marks
looking for the previous/next page mark. As a result there is no time for anything
else to happen and the program becomes unusably slow.

Fix is to remember the current line/column and only search to find what page
we are on if the line/column have changed since the previous update.

Fixes #720

### DIFF
--- a/src/lib/Guiguts/StatusBar.pm
+++ b/src/lib/Guiguts/StatusBar.pm
@@ -12,77 +12,87 @@ BEGIN {
 
 # Routine to update the status bar when something has changed.
 #
-sub update_indicators {
-    my $textwindow = $::textwindow;
-    my $top        = $::top;
-    my ( $last_line, $last_col ) = split( /\./, $textwindow->index('end') );
-    my ( $line, $column )        = split( /\./, $textwindow->index('insert') );
-    $::lglobal{current_line_label}
-      ->configure( -text => "L:$line/" . ( $last_line - 1 ) . " C:$column" )
-      if ( $::lglobal{current_line_label} );
-    my $mode             = $textwindow->OverstrikeMode;
-    my $overstrke_insert = ' I ';
-    if ($mode) {
-        $overstrke_insert = ' O ';
-    }
-    $::lglobal{insert_overstrike_mode_label}->configure( -text => " $overstrke_insert " )
-      if ( $::lglobal{insert_overstrike_mode_label} );
-    my $filename = $textwindow->FileName;
-    $filename = 'No File Loaded' unless ( defined($filename) );
-    $::lglobal{highlightlabel}->configure( -background => $::highlightcolor )
-      if ($::scannos_highlighted);
-    if ( $::lglobal{highlightlabel} ) {
-        $::lglobal{highlightlabel}->configure( -background => 'gray' )
-          unless ($::scannos_highlighted);
-    }
-    $filename = ::os_normal($filename);
-    $::lglobal{global_filename} = $filename;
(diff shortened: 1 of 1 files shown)


src/lib/Guiguts/StatusBar.pm
```diff
@@ -12,77 +12,87 @@ BEGIN {
...
+        # window label format: GG-version - [edited] - [file name]
+        if ($edit_flag) {
+            $top->configure( -title => $::window_title . " - " . $edit_flag . " - " . $filename );
+        } else {
+            $top->configure( -title => $::window_title . " - " . $filename );
+        }
+        update_ordinal_button();
+        $textwindow->idletasks;
+
+        # get_page_number() can take a significant time to execute, so check if line/column
```
cpeel, line 59: Page numbers can change in the middle of a line? Is this because of where page markers are? I'm confused on why we need to check the column and not just the line.

windymilla:
Yes, once the text is rewrapped, page boundaries can be anywhere.

cpeel:
> Yes, once the text is rewrapped, page boundaries can be anywhere.

Ah yes, I wasn't thinking about rewrapping. Thanks!

PR #731, opened 2021-04-24
Allow user to copy/paste from Version numbers dialog

### LINKED ISSUE #724: It would be nice if the Versions pop up under the Help menu could be copy/pasted

windymilla:
It would be useful for the user to be able to copy the version info when they want
to report an issue.

Replace the Dialog containing a message with a DialogBox containing a read-only
Text widget containing the information.

Fixes #724


cpeel:
Oh, nicely done.

PR #758, opened 2021-05-13
Stop using charset ascii when autogenerating HTML

### LINKED ISSUE #755: Character Encoding in HTMLConvert.pm Should Be Updated
Although the HTML Validator which is packaged with Guiguts does not see it as an error and technically speaking it is not an error, however, the on-line W3C HTML Validator warns that "charset=ascii" may not be widely supported. This encoding is equivalent to "charset=us-ascii" which may be better supported across platforms.

Therefore, the encoding should be updated to produce "charset=us-ascii"

```
    if ( $::lglobal{leave_utf} && ::currentfileisunicode() ) {
        $headertext =~ s/BOOKCHARSET/utf-8/;
     } elsif ( $::lglobal{keep_latin1} && ::currentfileislatin1() ) {
         $headertext =~ s/BOOKCHARSET/iso-8859-1/;
    } else {
         $headertext =~ s/BOOKCHARSET/us-ascii/;
     }
```

windymilla:
If a file had no non-ascii characters, and the default option to treat all files
as utf-8 was turned off, the HTML header set the charset to ascii.
Although this is valid, some tools whine about it and suggest us-ascii instead.

Since for an ascii file, the charset iso-8859-1 is equally valid, use that when
ascii would have been used previously

Fixes #755

### DIFF
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -1407,10 +1407,8 @@ sub html_parse_header {
     $headertext =~ s/BOOKLANG/$::booklang/g;
     if ( $::lglobal{leave_utf} && ::currentfileisunicode() ) {
         $headertext =~ s/BOOKCHARSET/utf-8/;
-    } elsif ( $::lglobal{keep_latin1} && ::currentfileislatin1() ) {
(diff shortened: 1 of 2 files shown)


windymilla:
Replied to comments in issue - if you're happy with them, then OK to merge - thanks

PR #771, opened 2021-06-13
Add auto-correction features to error check dialog

### LINKED ISSUE #753: Add auto-correction features to error check box
Currently
- Mouse-1 views the error
- Mouse-3 (or 2) removes the error and views the next

Add
- Ctrl+Mouse-1 to make the change suggested by the error, if possible
- Ctrl+Mouse-3 to make the change, remove the error and view the next
- Ctrl+Shift+Mouse-3 to remove all identical errors reported on other line numbers

windymilla:
Add features to process queries in the error check dialog, initially for Jeebies
and Load Checkfile (to supoprt OCRfixr output)

For Jeebies, the change will be to swap he/be.
For Load Checkfile, if the text is `Suggest 'X' for 'Y'` then replace Y with X at
marked position.

1. Ctrl+Mouse-1 makes the change suggested by the query
2. Ctrl+Mouse-2/3 makes the change and removes the query (as Mouse-2/3
currently does)
3. Ctrl+Shift+Mouse-2/3 removes all queries matching the clicked on, i.e.
to get rid of multiple wrong identical suggestions

Fixes #753

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -127,23 +127,53 @@ sub errorcheckpop_up {
     $::lglobal{errorchecklistbox}->bind(
         '<<remove>>',
         sub {
-            ::busy();
-            my $xx = $::lglobal{errorchecklistbox}->pointerx - $::lglobal{errorchecklistbox}->rootx;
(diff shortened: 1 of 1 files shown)


cpeel:
I apologize for the long delay on getting to this MR. I kept thinking I'd find / make time to try this out on my Mac but that clearly hasn't happened. The code looks good to me so I'll get it merged in.

PR #786, opened 2021-07-17
Clarify match numbers displayed by spellchecker

### LINKED ISSUE #760: Change to Info displayed in Spell Checker window
Currently when spell checker finds something not in the dictionary it may show four different things after the "Not in Dictionary" message on the top line. It can be a mix of exact match, hyphenated match, different case match, and possessive.

Up to four numbers may appear above the current word:

1 number is just the number of occurrences (exact matches)
2 numbers mean number of occurrences and number of occurrences as part of a hyphenated word, e.g. "abc" twice and "abc-d" once would say "2, 1 hyphens in text"
3 numbers mean number of occurrences, number of occurrences with different case, e.g. "Abc", and number of possessives, e.g. "Abc's"
4 numbers mean occurrences, case, possessive, hyphen

It would be nice if some kind of comment field appeared under the count to explain what each of the numbers meant in any of the four situations above.

windymilla:
When spell check finds a misspelled word, it displays up to 4 numbers which
detail variants on the word: exact matches, case variants, possessive variants
(add or remove 's), hyphen variants.

Since there can be 1, 2, 3 or 4 numbers displayed, and the meaning of the
numbers varies depending on how many are displayed, it's not easy for the
user to know/remember what each one means. Previously only hyphens were
labelled.

Now, output an extra word to describe the type of match, eg.
instead of
`3, 2, 1, 4 hyphens in text.`
output
`3 exact, 2 case, 1 possessive, 4 hyphen in text.`

This is only intended as a reminder; the manual will give more detailed
explanations of the meaning of those words, e.g. whether case is all case
variants or just capitalised first letter, that possessive means either adding
or removing `'s`, and that hyphen means the word is used in a longer
hyphenated word, not that it appears solo in a hyphenated form,  e.g.
`tomorrow` would flag `tomorrow-people`, not `to-morrow`.

There was also a bug previously where the first spelling found did not display
the additional types of match, only exact matches. This was due to partly
duplicated code in subs spellcheckfirst and spellchecknext. Changed to call
spellchecknext at end of spellcheckfirst and removed duplicated code.

Fixes #760

### DIFF
--- a/src/lib/Guiguts/SpellCheck.pm
+++ b/src/lib/Guiguts/SpellCheck.pm
@@ -18,35 +18,13 @@ sub spellcheckfirst {
     ::hidepagenums();
     $::globalspelldictopt = $::booklang;
     spellloadprojectdict();
-    $::lglobal{lastmatchindex} = '1.0';
 
     # get list of misspelled words in selection (or file if nothing selected)
     spellget_misspellings();
-    my $term = $::lglobal{misspelledlist}[0];               # get first misspelled term
-    $::lglobal{misspelledentry}->delete( '0', 'end' );
-    $::lglobal{misspelledentry}->insert( 'end', $term );    # put it in the appropriate text box
-    $::lglobal{suggestionlabel}->configure( -text => 'Suggestions:' );
-    return unless $term;                                    # no misspellings found, bail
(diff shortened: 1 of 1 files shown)


cpeel:
Very nice - love code reuse!

PR #802, opened 2021-07-29
Fix long-standing right-click bug

### LINKED ISSUE #801: Right-clicking 100 times causes "deep recursion" error
**Describe the bug**
Using right-click to bring up the popup menu 100 times caused deep recursion, followed by a crash if the user continued to use right-clicks.

windymilla:
If you right-clicked 100 times, you would hit a "deep recursion" error. If you persisted, the program would crash, presumably once the recursion got too
deep. The fault was that the code called `Popup` instead of `post`. The former
does not return, leading to an extra level of recursion each time the right-click
menu was used. The `post` method is the correct one for this case.

Fixes #801


cpeel:
Oh, nice find!!

PR #807, opened 2021-08-02
Preserve white background for Windows check boxes

### LINKED ISSUE #806: Check button highlight color changes under Windows even when cancelled
**Describe the bug**
Preferences->Set Button Highlight Color changes the color a button is highlighted when the mouse passes over it. On non-Windows systems, this color is also used for the background of checkboxes. On Windows systems checkboxes are always white. Lack of brackets combined with the ternary conditional operator meant that checkboxes were changed to the highlight color on Windows systems (even if the color selection dialog was cancelled rather than OKed)

windymilla:
The logic was failing due to lack of brackets to enforce operation precedence
with the ternary operator, meaning that although on Windows systems check
boxes should remain white, they did not.
Note that this code was executed even if the Color Choose dialog was
cancelled, which is fine because with the correction it will have no effect, as
was probably the original coder's intent, and is the structure used in similar
Color Choose code nearby.

Fixes #806

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,9 @@
   such as emdashes. The shortest line determination was also incorrect
 - Right-clicking the text window over 100 times would cause an error, and
   eventually the program would exit
+- Checkbox backgrounds on Windows changed color instead of remaining white when
+  Preferences->Appearance->Set Button Highlight Color was used, even if the
+  color choosing was cancelled
 
 
 ## Version 1.3.1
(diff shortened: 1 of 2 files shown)


charliehoward4dp:
Is this fix in Beta2 or only in a later, unreleased Beta?  In Beta2, it still occurs in most checkboxes, with the exception of the Search & Replace Dialog.

windymilla:
No, it was done after Beta2.

PR #818, opened 2021-08-22
Make searches wrap during curly quote processing

### LINKED ISSUE #816: Have "Select Next Straight Single Quote" wrap around when searching
"Select Next Straight Single Quote" starts searching from the current cursor position but stops searching at the end of the file. 

Could it be enhanced to wrap around and continue searching from the start of the file in the way that "Find Asterisks w/o slash" does?

windymilla:
To avoid the user thinking prematurely that they have done all straight to
curly quote processing, make the searches for "@" and straight single quotes
wrap at the end of the file. Now if the user gets a bell and icon flash, it's
because there are no more "@" or straight single quotes left in the file.

Fixes #816


cpeel:
Nice. I can see this being much more intuitive.

PR #819, opened 2021-08-22
Avoid crash when tool executable doesn't exist under Linux

### LINKED ISSUE #814: Missing tools cause crash under Linux
**Describe the bug**
If a tool, e.g. bookloupe, is not in the folder it is configured to be in, the "system" call causes the whole program to crash when run under Linux. Under Windows, it just gives an error but doesn't crash. See [this post](https://www.pgdp.net/phpBB3/viewtopic.php?p=1254388#p1254388) for details

**Expected behavior**
Check if the file to be run exists before attempting to run it, so it can avoid the fatal error.

windymilla:
Under Linux, the system call is used to spawn external tools like bookloupe.
If the file didn't exist, GG crashed, whereas under Windows it just gave an
error.

Add a check that the intended tool exists and is executable before attempting
to run it.

Fixes #814

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -342,7 +342,11 @@ sub dos_path {
 sub run {
     my @args = @_;
     if ( !$::OS_WIN ) {
-        system { $args[0] } @args;
+        if ( -x $args[0] ) {
+            system { $args[0] } @args;
+        } else {
+            warn "Executable file $args[0] not found";
+        }
     } else {
         require Win32;
         require Win32::Process;
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -342,10 +342,10 @@ sub dos_path {
 sub run {
     my @args = @_;
     if ( !$::OS_WIN ) {
-        if ( -x $args[0] ) {
+        if ( File::Which::which( $args[0] ) ) {
             system { $args[0] } @args;
         } else {
-            warn "Executable file $args[0] not found";
+            warn "Executable $args[0] not found nor is it on the path";
         }
     } else {
         require Win32;


src/lib/Guiguts/Utilities.pm
```diff
@@ -342,7 +342,11 @@ sub dos_path {
 sub run {
     my @args = @_;
     if ( !$::OS_WIN ) {
-        system { $args[0] } @args;
+        if ( -x $args[0] ) {
```
cpeel, line 345: This assumes that `$args[0]` is an absolute path (or presumably a relative one to the Guiguts dir?). Would it be better to use `File::Which` here like we do in `Preference.pm`?

windymilla: I don't think so, because if I understand correctly, File::Which checks if it's on the path. By the time we are here, about to call system, we are using the (relative or absolute) path to the tool that is stored in the settings file - whatever the system has defaulted to or the user has set up. This test is solely to check that it exists in the place we expect it before trying to run it. I'm pretty confident if the `-x` test returns true, then the system should be able to run successfully. Maybe I don't know enough about File::Which to know the answer?

windymilla: Oh, unless you are thinking that perhaps this routine might be called in other circumstances, not a Guiguts tool? Hmm...maybe. I'll investigate a bit more and try to make a more informed decision. I know this edit fixes the problem the user reported because they have tried it, but I don't know for sure that it doesn't break something else. I'll get back to you. Thanks

windymilla: You were right ⭐ 
Tools like pphtml are run by executing perl, which my previous code would say didn't exist.
I've sent a new version (using Which) to the user to check out. If it's OK, I'll add another commit.
Thanks

windymilla: New version works fine for user, who also confirmed the previous version failed to run PPhtml.
So, should be good to merge now - thanks

cpeel: Excellent!

cpeel: I provided value this time! 😆

windymilla: Always ❤️

PR #830, opened 2021-09-05
Ensure replacement uses the same regex options as search

### LINKED ISSUE #829: Dollar replacement failure when newline is included in certain regexps
**Describe the bug**
The regex `(abc.def)\n` matches
```
abc
def
xyz
```
This is because the `\n` causes GG to treat the period as a newline.

However, if the replacement string is `XXX$1XXX`, the `$1` is not substituted.

Reported [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1255652#p1255652)

windymilla:
When replacing `$n` with a bracketed section of the regex, the "s" flag was
not used on the regex search. This flag causes "." to include newlines. Since
the flag is used when the user is searching for the regex in the first place,
it needs to also be used when doing the replacement. Otherwise, the user
does a Search and finds a match, but when they try to replace, they just get
`$1` left in their file.

Note this should have no effect on the commonly used `(.|\n)` construction.

Fixes #829

### DIFF
--- a/src/lib/Guiguts/SearchReplaceMenu.pm
+++ b/src/lib/Guiguts/SearchReplaceMenu.pm
@@ -127,6 +127,10 @@ sub searchtext {
                     my $wholefile = $textwindow->get( '1.0', $end );
 
                     # search is case sensitive if $::sopt[1] is set
+                    # Note that regex match options "sm" are needed here and when handling ()/$ grouping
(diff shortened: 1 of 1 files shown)


cpeel:
Should we leave comments here and in the search part so that future code spelunkers know that the modifiers need to match?

windymilla:
Good point - comments added - should be OK to merge now - thanks

Issue #823, opened 2021-08-27
Incorporate the lastest ebookmaker into Guiguts


windymilla:
Version 0.11.4 has [just been announced](https://www.pgdp.net/phpBB3/viewtopic.php?p=1254893#p1254893), and is being used in production at PG.

In the announcement
> This version on ebookmaker has not been tested on Windows, as I don't currently have access to a Windows box for development.

PR #837, opened 2021-10-02
Allow ebookmaker 0.11.13 to create HTML output files

windymilla:
1. Bundle 0.11.13 ebookmaker
2. Add menu option to trigger ebookmaker output of HTML file
3. Pass correct args to ebookmaker to output the requested files

HTML file (with images) is output to a subfolder "out" by ebookmaker because
output folder is the same as input folder.

### DIFF
--- a/src/lib/Guiguts/MenuStructure.pm
+++ b/src/lib/Guiguts/MenuStructure.pm
@@ -788,7 +788,7 @@ sub menu_html {
         [
             'command', 'EB~ookMaker epub/mobi Generation', -command => sub { ::ebookmaker("epub"); }
         ],
-        [ 'command', 'EB~ookMaker HTML Generation', -command => sub { ::ebookmaker("html"); } ],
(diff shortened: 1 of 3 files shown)


windymilla:
Fixes #823

src/lib/Guiguts/MenuStructure.pm
```diff
@@ -785,7 +785,10 @@ sub menu_html {
             }
         ],
         [ 'separator', '' ],
-        [ 'command',   'EB~ookMaker epub/mobi Generation', -command => sub { ::ebookmaker(); } ],
+        [
+            'command', 'EB~ookMaker epub/mobi Generation', -command => sub { ::ebookmaker("epub"); }
+        ],
+        [ 'command', 'EB~ookMaker HTML Generation', -command => sub { ::ebookmaker("html"); } ],
```
cpeel, line 791: Should the two EBookmaker entries have different shortcut keys rather than both `o`?

windymilla: Oops - new commit added - thanks

PR #854, opened 2021-11-07
Correct misspelling of `caret` in bookloupe output

### LINKED ISSUE #851: Spelling of caret as carat is wrong in bookloupe output
**Describe the bug**
For historical reasons, caret is spelled as carat in bookloupe output, and the corresponding error in the Bookloupe View Options dialog. It would soothe some proofreading brains if it could be corrected.

windymilla:
`Caret` is spelled `carat` in bookloupe output, and by necessity in the
Bookloupe View Options dialog.

Correct the spelling before displaying it, since bookloupe is not directly
part of Guiguts.

Fixes #851


cpeel:
It's pretty amazing how ... detail-oriented our users are!

PR #858, opened 2021-11-15
Output line number in undefined CSS message from pphtml

### LINKED ISSUE #849: PPHTML should include the line numbers of "possibly not defined" CSS classes
PPHTML recently was enhanced to include the line numbers of "CSS possibly not used." If it also reported the line numbers of "CSS possibly not defined", it would make it much easier to track them down.  See the (https://www.pgdp.net/phpBB3/viewtopic.php?p=1260847#p1260847) Q&A in the "Help With Guiguts" Forum.

windymilla:
#784 added line numbers to unused CSS messages. This adds line numbers
to undefined CSS messages from pphtml.

Minimal changes required to store and track the line number where CSS is
used in order to be able to output it later. I do not want to make substantial
edits to pphtml because it may be used separately from GG, because its output
is very familiar to experienced PPers and because it's a relatively simple
peripheral program that is easy to understand with its current structure, and
best not meddled with.

Fixes #849

### DIFF
--- a/src/lib/ppvchecks/pphtml.pl
+++ b/src/lib/ppvchecks/pphtml.pl
@@ -8,12 +8,13 @@ use Getopt::Long;
 # command line version of pphtml.pl
 # author: Roger Frank (DP:rfrank)
 
-my $vnum = "1.16";
+my $vnum = "1.17";
 
 my @book         = ();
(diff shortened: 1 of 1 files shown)


cpeel:
Nice solution!

PR #863, opened 2021-12-04
Remove option to create HTML5 ebookmaker output

windymilla:
Things have moved on since the HTML5 output feature was added by #823
PPers can now submit an HTML5 master file, so it is confusing to have
ebookmaker create HTML5 files, which incidentally should not be submitted.
So, remove the menu option, but leave the argument to the ebookmaker
routine, since it, or something like it could easily be needed again.


windymilla:
Fixes #863

PR #872, opened 2021-12-20
Generate HTML5 files

windymilla:
HTML autogenerate should create HTML5 files:

1. Always read/write utf8 files
2. Update default HTML header.txt for HTML5 (need to load it as utf8 now)
3. Do not write summary attribute in tables
4. Use numeric, not named, entities apart from a few permitted exceptions
5. Stop pphtml complaining about missing table summary
6. Update test suite to accept numeric emdash entities

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -197,7 +197,6 @@ our $utfcharentrybase      = 'dec';           # 'dec' or 'hex' allowed
 our $utffontname           = 'Courier New';
 our $utffontsize           = 14;
 our $utffontweight         = 'normal';
-our $utf8save              = 1;               # True = always save utf8, false = only if unicode characters in file
 our $verboseerrorchecks    = 0;
 our $vislnnm               = 0;
 our $wfstayontop           = 0;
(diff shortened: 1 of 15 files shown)


windymilla:
Will do a separate MR after this merge to incorporate the Nu HTML Checker

cpeel:
Looks great! Love seeing all the deleted code 😁 

How do you see these bigger changes as part of the GG release cycle? Are these going into a 1.4 or a 2.0 version? Mostly trying to wrap my head around how (and if) we should be thinking about managing the code with potential 1.3.x updates alongside these bigger changes.

windymilla:
I'm hoping these will be in a 1.4 version, but I'm not envisaging any more 1.3.x updates. I more or less have the changes necessary for the new "Nu" HTML Checker to be bundled and run, in place of the old HTML validator. Once those are in, I can make a test release up for a few eager beavers, then probably a beta release for wider trials of the HTML5 business, then hopefully full release 1.4.0. The changes are really not that huge. It's mostly just a different HTML header at the moment and avoiding named entities.
So, unless you see problems with that scheme, I'd be happy for this to be merged in, and I'll prepare another MR for the Nu HTML Checker.

cpeel:
Sounds like a great plan to me. 👍

windymilla:
Thanks

PR #873, opened 2021-12-22
Bundle the Nu HTML Checker

windymilla:
The Nu HTML checker is used by W3C online checker if an HTML5 file is being
validated. Bundle this with Guiguts in place of the old onsgmls.exe. The .jar
file is quite large, but in time, the old CSS jar file might be retired, because
the Nu checker also checks CSS, but retaining the existing separate CSS
checker for now.

Replace the previous W3C HTML validation with two options to use the Nu
checker. The first checks as HTML, the second as XHTML. Although it had
been postulated that the XHTML test was sufficient, there are messages,
e.g. warning about use of "name" attribute, that only occur in the HTML check.
The XHTML check is necessary to catch unclosed elements, e.g. `<hr>`

Added new XHTML test, and adjusted several other tests due to the
input test html file having a different (HTML5) header, which affected
line numbers.


tools/W3C/package.sh
```diff
@@ -13,13 +13,12 @@ cp README.md $DEST
...
 
-# For OpenSP we only have the pre-built Windows files
-if [[ $OS == "win" ]]; then
-    VERSION=1.5.2
-    curl -L -o opensp.zip https://sourceforge.net/projects/openjade/files/opensp/$VERSION/OpenSP-$VERSION-win32.zip/download
-    # we only want onsgmls.exe and it's dependency DLL
-    unzip opensp.zip onsgmls.exe osp152.dll -d $DEST
-    rm opensp.zip
-    # and the xhtml DTDs and related files
-    cp xhtml* $DEST
```
cpeel, line 24: Does this mean we can remove these files as well?
```
xhtml-lat1.ent
xhtml-special.ent
xhtml-symbol.ent
xhtml.soc
xhtml1-strict.dtd
xhtml1.dcl
```

windymilla: Oops, yes. I was thinking by removing the command from package.sh, I was getting rid of them, but of course they live in the tools folder. I'll add a commit

windymilla: Done - thanks

cpeel: Awesome!

PR #875, opened 2021-12-27
Fix MacOS dependency install order and command

cpeel:
XQuartz is required for TK installation so put that above the perl section. Also update the cask install syntax that changed in [homebrew 2.6.0](https://brew.sh/2020/12/01/homebrew-2.6.0/).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -128,8 +128,8 @@ to be done once. These instructions walk you through it.
 * Guiguts code
 * [Xcode Command Line Tools](https://developer.apple.com/library/archive/technotes/tn2339/_index.html)
 * [Homebrew](https://brew.sh/)
-* Perl & [Perl modules](#perl-modules)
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -169,6 +169,19 @@ Open Terminal.app and install Homebrew with:
...
+[XQuartz](https://www.xquartz.org/) is an X11 windows manager. If you don't
+have it installed already, you can either download and install it manually
+via the link _or_ install it with Homebrew using:
+
+```
+brew install --cask xquartz
+```
+
+After you install XQuartz, you must **log out and back in** before Guiguts can
+use it as the X11 server.
```
srjfoo, line 183: Definitely confirmed that this is the case.

srjfoo (approved):

windymilla (approved):
Thanks both

PR #891, opened 2022-01-18
Convert HTML4 header.txt to HTML5

windymilla:
If user has copied a customized header.txt from a previous release, it will have
an (X)HTML4 header. As this is read, convert the few lines necessary to make
it valid HTML5.

This allows the user to upgrade fairly painlessly from a 1.3 (HTML4) release to
a 1.4 (HTML5) release.

Also, beef up the upgrade notes a bit to clarify procedure for retaining any
header.txt customizations.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -33,13 +33,21 @@ steps below work for all operating systems.
      relevant ones to the new release. This will not apply to most users.
    * The file `headerdefault.txt` has had important changes made to it in this
      release. When you first run Guiguts, it copies this file to one named
-     `header.txt`. If you made changes to `header.txt`, or are unsure whether
-     you did or what they were:
+     `header.txt`. If you have customized your `header.txt` in a previous
+     release, you may want to retain those changes in the new release.
+     If you are upgrading from a 1.3.x release, you can copy `header.txt` from
+     the old release to the new one after you have run and exited Guiguts once.
+     If you are upgrading from an older release, e.g. 1.0.25, 1.1.x or 1.2.x,
+     you should follow the steps below to ensure you get all the updates:
      1. In the old directory compare `header.txt` with `headerdefault.txt`,
         making note of any differences.
      2. Run the newly installed Guiguts program to create the new `header.txt`
         then immediately exit.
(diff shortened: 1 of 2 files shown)


UPGRADE.md
```diff
@@ -33,13 +33,21 @@ steps below work for all operating systems.
...
+     the old release to the new one after you have run and exited Guiguts once.
+     If you are upgrading from an older release, e.g. 1.0.25, 1.1.x or 1.2.x,
+     you should follow the steps below to ensure you get all the updates:
      1. In the old directory compare `header.txt` with `headerdefault.txt`,
         making note of any differences.
      2. Run the newly installed Guiguts program to create the new `header.txt`
         then immediately exit.
      3. Edit `header.txt` in the new directory and make the changes you want
         to carry over from the old release.
+     If you were using `@media handheld` in your customized header file, please
```
cpeel, line 48: [This is rendering](https://github.com/windymilla/guiguts/blob/html45/UPGRADE.md) as a continuation of the list item above it. Maybe a newline before this one will get it in the right place?

UPGRADE.md
```diff
@@ -33,13 +33,21 @@ steps below work for all operating systems.
...
+     note that it is now deprecated. The same effects can be achieved using
+     the `x-ebookmaker` class as described [here](https://www.pgdp.net/wiki/DP_Official_Documentation:PP_and_PPV/DP_HTML_Best_Practices/Case_Studies/Media_Types_and_Mobile_Formats).
+     Alternatively, ask in the [DP forum Help with: guiguts thread](https://www.pgdp.net/phpBB2/viewtopic.php?t=11466).
```
srjfoo, line 51: Also add a blank line above, here, if "Alternatively" is meant to start a new paragraph.

UPGRADE.md
windymilla: I decided it would be better as a completely separate bullet point to avoid confusion with the steps immediately above. It applies to everyone, whichever release they are upgrading from

UPGRADE.md
windymilla: No, this is part of the same paragraph - thanks

PR #893, opened 2022-01-22
Convert sub/superscripts correctly in block markup

### LINKED ISSUE #892: Inconsistent convertion of superscript inside /* */
GG correctly converts abc^x and abc^{xy} in normal text paragraphs.

However inside a nowrap section /* */ it converts some superscripts but not all of them.
In the attached file (pre-generation) there are 28 caret elements.
After generation (I used the default header file) 11 were converted correctly, and 17 were left unchanged.

I have also seen the case (but not with this test file) where an element such as abc^{xy} was converted, but incorrectly,  to abcxy.

I suspect this issue was present in 1.0.25, but I don't have that on my system anymore.

**To Reproduce**
1. Load the attached file gg-caret test.txt into Guiguts
2. Open the HTML Generator menu, and run generation.
3. Search for ^ and `<sup>`

**Expected behavior**
All caret elements should be converted to <sup>xx</sup>

**Environment**
* Guiguts version:  1.3.3
* OS and version:  Windows 10
* Perl version: 
* Perl distribution: Strawberry
* 
[gg-caret test.txt](https://github.com/DistributedProofreaders/guiguts/files/7916789/gg-caret.test.txt)

windymilla:
During HTML generation, the routines to convert sub/superscripts were called
for each line of the file near the beginning of the conversion.
They edited the file directly to add the sub/sup markup.
The calling routine ignored this change and continued to work with the line
it had extracted from the file before the sub/sup conversion. This meant that
if further work was done on the line, e.g. adding a span to indent it for a table,
the sub/sup edit was lost.

Fixed by returning the modified line to the calling routine - it didn't seem
worth trying to restructure, since that would open a can of worms regarding
the structure of the HTML conversion with its massive outer loop.

Fixes #892 

This has been a bug since 1.0.25 at least; very surprising it hasn't been
reported/fixed before. Possibly HTML files have been posted with the
`^` or `_` markup still in where sub/sups occur inside block markup, since
it would be easily overlooked by the PPer post-conversion.

### DIFF
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -28,15 +28,19 @@ sub html_convert_tb {
     return 0;
 }
 
+# Convert subscripts in the textwindow at this line
+# Also return modified line so remainder of HTML conversion acts on the edited version
 sub html_convert_subscripts {
     my ( $textwindow, $selection, $step ) = @_;
     if ( $selection =~ s/_\{([^}]+?)\}/<sub>$1<\/sub>/g ) {
         $textwindow->ntdelete( "$step.0", "$step.end" );
         $textwindow->ntinsert( "$step.0", $selection );
     }
-    return;
+    return $selection;
 }
 
+# Convert superscripts in the textwindow at this line
+# Also return modified line so remainder of HTML conversion acts on the edited version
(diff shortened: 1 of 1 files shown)


cpeel:
Returning the selection is a very elegant fix!

windymilla:
Thanks!

PR #895, opened 2022-01-29
Use CRLF line endings when saving main text file

### LINKED ISSUE #894: Output CRLF line endings for the main text file on all platforms
Currently CRLF is forced on Windows only. On other platforms, the system default is used.
However, CRLF is pretty baked into the DP software, and many other editors can cope with whatever line endings there are.

Ask in the forums before implementing. Could possibly have a Prefs flag?

windymilla:
DP text files use CRLF line endings, and some tools expect this.
Guiguts already forced this for Windows when saving the main text file.
Mac/Linux users have said it would be easier if it did the same for them.

A variety of line endings are accepted when loading a file.

Fixes #894


cpeel:
This make sense and it looks like a good change. As a Linux/Mac user I recommend featuring this change prominently in the changelog since it will be unexpected to people who weren't already part of the discussion/decision.

PR #898, opened 2022-02-12
Add TIA OCR import feature

windymilla:
Accept either a zipped or unzipped Abbyy.gz file from TIA.
If file is zipped (extension .gz) then user must have gzip on their PATH.

Code was written to cope with importing several different types, and a new
type could be easily added. Hopefully this doesn't make it look too odd now
there is only one type of import that turned out to be successful.

Thanks to @srjfoo for coming up with the idea, investigations and testing.


srjfoo (approved):
And thanks to you, @windymilla for the implementation. I just asked a general question; you're the one who took the ball and ran.

windymilla:
Thanks. Possibly we could include gzip if the feature takes off, but I
suspect it will be very niche. There are several unzip tools a Windows PM
could use on a downloaded ABBYY file, then they can import the unzipped
version.

PR #902, opened 2022-02-17
Strip all rewrap markup from input to bookloupe

### LINKED ISSUE #901: Bookloupe only ignores 3 of the 8 rewrap makers when checking for "Forward slash"
**Describe the bug**
When Bookloupe looks for the "/" character in a file ["Forward slash" search enabled in options], it correctly ignores those characters when they are part of certain rewrap markers, namely /**/, /$$/ and /##/. In the case of the other five rewrap markers, however, the accompanying "/" characters are not ignored and thus queried by Bookloupe. This is inconsistent behaviour. 

**To Reproduce**
1. Open Guiguts with 
[rewrap-marker-test.txt](https://github.com/DistributedProofreaders/guiguts/files/8082647/rewrap-marker-test.txt)
2. Open the Tools>Bookloupe menu 
3. Enable "Forward slash" option in Options menu
4. Click "Run Checks" radio button
5. See which rewrap marker pairs the queried "/" characters are part of.

**Expected behavior**
I would expect the "/" character in any rewrap marker pair being ignored.

**Environment**
* Guiguts version: 1.4-test3
* OS and version: Windows 10
* Perl version: 
* Perl distribution: 

**Screenshots**

**Additional context**

windymilla:
Bookloupe only knows about the 3 oldest rewrap markup types, which it
does by stripping them from each line before it checks it.
Since we are not editing bookloupe code, in order to make bookloupe ignore
the newer rewrap markup types, we need to strip them from the file before
sending it to bookloupe.

Only strip if markup is on its own on the line. Note that opening markup can
be followed by `[` to specify margins.

Fixes #901

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -516,7 +516,14 @@ sub errorcheckrun {    # Runs error checks
     ::savesettings();
     $top->Busy( -recurse => 1 );
 
-    savetoerrortmpfile($tmpfname);
+    # Bookloupe strips /*, /$ and /#, but doesn't know about newer rewrap types,
+    # so strip them all here if they are the only text on the line,
+    # or if the open rewrap marker is optionally followed by margins in square brackets
+    my $striptext =
+      $errorchecktype eq 'Bookloupe'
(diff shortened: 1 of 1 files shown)


cpeel:
Clever solution 👍

Issue #904, opened 2022-03-19
Simple regex Rpl All seems to go into a loop in the patched 1.4.0 test version


charliehoward4dp:
When the following regex is used in a Rpl All on the attached file, GG 1.4-test3 seems to go into a loop-like state, deleting one character at a time (or none).  It doesn't happen with single/manual S&R's, and didn't happen in the past.  I suspect the "recent" patch associated with "Rpl All" is the cause (I made the patch by editing "SearchReplaceMenu.pm" according to instructions.)

Here's the regex:
  S:  \\[Blank Page]\\n

The Replace term is null. The regex is just supposed to delete lines containing [Blank Page]
[regex_bug.zip](https://github.com/DistributedProofreaders/guiguts/files/8310257/regex_bug.zip)

EDIT: The backslashes were there, but Github originally didn't display them. This edit doubled them so they now show.

windymilla:
Shouldn't there be backslashes to escape the square brackets, otherwise
it's deleting all those letters individually

PR #906, opened 2022-03-20
Revert "Make Replace All replace adjacent matches"

windymilla:
This reverts commit 8ef975e427058de2f5e0229174eb3e4dfccfb011.

The previous commit caused the bug reported in issue #904 

I didn't feel totally confident with the fix, and should have heeded my own concerns. I don't want to try other fixes to that part of the code - it really needs a rewrite (but probably retaining some behaviour which is not obvious)

Since the original bug wasn't reported for 10 years, and it is not catastrophic, I suggest we live with it.


windymilla:
Fixes #904

PR #920, opened 2022-04-25
Improve text fraction conversion

### LINKED ISSUE #913: Incorrect conversion of fractions
GG 1.4 (from 12/30/2021) "Convert unicode fractions only" converted 1/2,000th to  ½,000th  Should not have converted it at all.

windymilla:
Previously, `1/2,000` would be converted to `½,000`

Beefed up the regex to include a negative lookahead for `,\d` so conversion
will not be attempted in this case.

Fixes #913


cpeel:
> Previously, 1/2,000 would be converted to ½,000

Ha! That/s one interesting fraction! Nice fix.

PR #934, opened 2022-05-21
Add a better start experience for macOS users

cpeel:
Add a simpler way for macOS users to start Guiguts. Thanks to @srjfoo for all of the investigative work to make this possible. The `guiguts.command` file has been added to git with the executable bit set, which persists when packaged on a Mac so no changes are necessary after unpackaging it. We should confirm that this is still the case when packaged on a Windows machine.

This also introduces a new packaging script which removes the start-up files from operating system packages where they are not useful.


srjfoo (approved):
Works for me.

cpeel:
@windymilla, can you please do a package of this branch so that Sharon and/or I can validate the execute bits are set correctly when packaged on a Windows box? You're welcome to upload it somewhere to the TEST server for us to access if that's easiest.

windymilla:
Done - replying by PM with location

src/guiguts.command
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+# Startup script for macOS
+SCRIPT_DIR=$(dirname "$0")
```
windymilla, line 4: Does this work if the script is a symbolic link? I don't know anything about Macs, but got the impression online that if you drag the script to the Dock it creates a link, and it's possible that `dirname` will give the directory the link is in, not the original script. 
Suggested ideas for getting round it involved the used of `readlink/greadlink` ([Found something here](https://stackoverflow.com/questions/59895/how-can-i-get-the-source-directory-of-a-bash-script-from-within-the-script-itsel))

windymilla: Does it matter that there is no newline on the last line of the script?

windymilla:
For what it's worth, I do the packaging using MinGW Shell.
Within that shell, `ls -l` shows the new file with exec permissions before I package it.
If I unzip the Mac package, it appears the exec permissions are retained

src/guiguts.command
srjfoo: @windymilla, it worked fine for me. Both double-clicking on it, and dragging it to the dock. I also verified that the permissions were properly retained in the zip file.

cpeel: The Dock doesn't appear to use symlinks which is why I think this works as-is. `readlink` on macOS doesn't have the `-f` param, unfortunately, so we can't use that.

It doesn't matter that there is no newline, but it would be better if it did so let me fix that.

cpeel:
UGH. macOS isn't letting me run it because it doesn't see it as secure. I'm looking if there is a way to work around this or if this is something we just need to document. Security is a PITA.

cpeel:
I've figured out how to work around Apple's security fixes for this and updated the `INSTALL.md`. We also need the file to have Unix-style line endings so I've added a `.gitattributes` file which should force the right line endings on Windows just for that file.

Nigel can you please package the branch up again so we can test the line endings fix? Also please confirm this doesn't mess up how git views the line endings in the rest of the repo for you -- I think this should only impact the one file.

srjfoo:
> UGH. macOS isn't letting me run it because it doesn't see it as secure. I'm looking if there is a way to work around this or if this is something we just need to document. Security is a PITA.

Weird -- I didn't have that issue.

srjfoo:
> 

I take that back -- @cpeel, when I actually tried the right copy of GG, I got the same response. (I accidentally used my dev version for testing instead of @windymilla's package.)

src/guiguts.command
windymilla: Excellent

windymilla:
> Nigel can you please package the branch up again so we can test the line endings fix? Also please confirm this doesn't mess up how git views the line endings in the rest of the repo for you -- I think this should only impact the one file.

Will do, hopefully this evening. Thanks

windymilla:
It seems `guiguts.command` has LF endings, but `guiguts.pl` (for example) has CRLF endings, so that looks as though it's worked OK. PM on the way with new package location.

srjfoo:
> It seems `guiguts.command` has LF endings, but `guiguts.pl` (for example) has CRLF endings, so that looks as though it's worked OK. PM on the way with new package location.

I'm still not being able to open it with the packaged guiguts.command script. I wonder if it has anything to do with being downloaded via dropbox?

cpeel:
> > It seems `guiguts.command` has LF endings, but `guiguts.pl` (for example) has CRLF endings, so that looks as though it's worked OK. PM on the way with new package location.
> 
> I'm still not being able to open it with the packaged guiguts.command script. I wonder if it has anything to do with being downloaded via dropbox?

Is that even after running the `xattr` command in the INSTALL.md? I'm not at a computer but can look at it later.

srjfoo:
> > > It seems `guiguts.command` has LF endings, but `guiguts.pl` (for example) has CRLF endings, so that looks as though it's worked OK. PM on the way with new package location.
> > 
> > 
> > I'm still not being able to open it with the packaged guiguts.command script. I wonder if it has anything to do with being downloaded via dropbox?
> 
> Is that even after running the `xattr` command in the INSTALL.md? I'm not at a computer but can look at it later.

🤦‍♀️ -- no; I made the false assumption that there would be no user intervention necessary, so I hadn't looked back at INSTALL.md. I have now done so, and yes that fixes it.

Meanwhile, I learned that I could list all attributes on a file. Hopefully this isn't getting too esoteric. 😁

srjfoo:
Update: Regarding the quarantine -- I don't know how relevant this is, since most Mac users will be using a download, but the quarantine is not present in my dev copy that I cloned via github. However, I just checked the 1.4.0 Mac download, and the files in the unzipped directory do include the quarantine, so the `xattr -d` will still be needed for anyone installing that way.

cpeel:
Yes, this is an artifact of Apple's security. macOS tags all downloaded files with that unless they are signed by a recognized Apple Developer key. It then refuses to run programs / scripts with that extended attribute. The attribute isn't added for git checkouts -- since they aren't technically downloads

cpeel:
> 



> It seems `guiguts.command` has LF endings, but `guiguts.pl` (for example) has CRLF endings, so that looks as though it's worked OK.

Excellent, thanks for checking.

> PM on the way with new package location.

The new package works great for me. I think this PR is complete and ready for final review / approval.

srjfoo (approved):

windymilla (approved):

PR #936, opened 2022-05-24
Convert line endings to native form in WF comma check

### LINKED ISSUE #935: Word Frequency "comma, Upper" check broken on Linux
**Describe the bug**
Due to #895, line endings in the saved file under Linux are Windows format. When a file is loaded normally, they are converted back to native format, but WF has its own "slurp" routine that reads the file, without converting line endings.

Fix by adding `$wholefile =~ s/\cM\cJ|\cM|\cJ/\n/g;` to slurpfile in WordFrequency.pm

windymilla:
Word Frequency has its own mini load file routine, which doesn't convert
line endings, unlike the main load file in `TextUnicode.pm`.
Since #895 changed the saved line endings under Linux/Mac, it broke this
file load, stopping WF from locating in the text file the "comma, Uppercase"
errors it detected.

Fixes #935

### DIFF
--- a/src/lib/Guiguts/WordFrequency.pm
+++ b/src/lib/Guiguts/WordFrequency.pm
@@ -1392,11 +1392,12 @@ sub slurpfile {
     my $wholefile;
     ::savefile() unless ( $textwindow->numberChanges == 0 );
     {
-        local $/;    # slurp in the file
+        local $/;                               # slurp in the file
         open my $fh, '<', $filename;
         $wholefile = <$fh>;
         close $fh;
         utf8::decode($wholefile);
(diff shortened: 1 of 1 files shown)


cpeel:
This change LGTM. Bit of a bummer that we have multiple places where we load the file. Is there value in having a single, reusable function that loads the file, or perhaps a single function that normalizes line endings post-load?

windymilla:
Tbh there would be some value of course, but I couldn't be bothered. Perhaps I shouldn't be so lazy and take the opportunity to improve the code?

cpeel:
Nah, at this point we pick our battles and the refactor seems like one we can skip. But if we run into this in a 3rd place I would give that some serious consideration.

windymilla:
Thanks. :)

PR #939, opened 2022-05-28
Fix opening of scratchpad.txt

### LINKED ISSUE #938: alt+ctrl+s to open scratchpad.txt does not work
When `<alt>+<ctrl>+s` is pressed, this error message appears:

 13:27:36: Guiguts 1.4.0: If you have any problems, please include any error messages that appear here with your bug report.
 13:51:19: Failed to run start scratchpad.txt: The system cannot find the file specified.

Guiguts apparently creates that file in its own folder, if the file doesn't exist, but doesn't open it.  Having a file of that name in the folder of the text being edited doesn't work, either.

windymilla:
Ctrl-Alt-s should open a scratchpad.txt file on Windows/Mac systems.
It hasn't worked since at least GG 1.0.25. No idea if anyone used to use
it, or would use it now, but anyhow...

Fix it by calling the cmdinterp routine that escapes commands, etc. before
they are sent to the OS.

Fixes #938

### DIFF
--- a/src/lib/Guiguts/KeyBindings.pm
+++ b/src/lib/Guiguts/KeyBindings.pm
@@ -229,20 +229,6 @@ sub keybindings {
     keybind( '<Alt-Left>',  sub { ::indent( $textwindow, 'out' ); } );
     keybind( '<Alt-Right>', sub { ::indent( $textwindow, 'in' ); } );
 
-    # Scratchpad
-    keybind(
-        '<Control-Alt-s>',
-        sub {
-            my $rcmd = $::OS_WIN ? 'start' : $::OS_MAC ? 'open' : '';
-            return unless $rcmd;
-            unless ( -e 'scratchpad.txt' ) {
-                open my $fh, '>', 'scratchpad.txt'
-                  or warn "Could not create file $!";
-            }
-            ::runner( ::cmdinterp("$rcmd scratchpad.txt") );
-        }
-    );
-
     # Help
     keybind( '<Control-Alt-r>', sub { ::display_manual("regexref"); } );
 
--- a/src/lib/Guiguts/KeyBindings.pm
+++ b/src/lib/Guiguts/KeyBindings.pm
@@ -233,12 +233,13 @@ sub keybindings {
     keybind(
         '<Control-Alt-s>',
         sub {
+            my $rcmd = $::OS_WIN ? 'start' : $::OS_MAC ? 'open' : '';
+            return unless $rcmd;
             unless ( -e 'scratchpad.txt' ) {
                 open my $fh, '>', 'scratchpad.txt'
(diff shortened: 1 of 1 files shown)


srjfoo:
On a Mac -- assuming that the `Alt` key maps to the `option` key on a standard Mac keyboard, I've tried
    `ctrl`+`option`+`s`
    `ctrl`+`shift`+`s`
    `command`+`option`+`s`
    `command`+`ctrl`+`s`
All of which map to "Save as" until I do an initial save. After I do an initial save, `ctrl`+`shift`+`s` still opens the "Save as" dialog, but all the others just re-save the file.

`shift`+`option`+`s` does nothing

@cpeel, I *_think_* I'm running the right code, but I'd appreciate a re-test in case I'm doing something wrong.

windymilla:
I don't think there's any way to access it except through the key stroke, so perhaps it's never worked on Macs?

windymilla:
You should have seen something, maybe just an error, even if you were running the old version

srjfoo:
> I don't think there's any way to access it except through the key stroke, so perhaps it's never worked on Macs?

Entirely possible. Update -- Instead of pulling the code into my local branch, I downloaded the zip to a different location. Same results in testing, but this time I noticed an error message. For all the help it may give:
```
[1]+  Bus error: 10           perl guiguts.pl
```

srjfoo:
> 

Oh ... and in addition, after my initial testing on my github branch (which I don't remember seeing an error for, but may have just missed it), I switched back to `master` and did the same tests, and got:
```
[1]+  Segmentation fault: 11  perl guiguts.pl
```

Not sure which key combination I got it for, but apparently it doesn't like one of them?

windymilla:
No idea.

srjfoo:
Just out of curiosity, is there a real need for a scratchpad (which I assume is just a place where the PPer can take notes)? When I was PP'ing 12+ years ago, I never even knew the feature existed; if I needed to take notes, I either used a different text editor or opened another GG document.

If it's not really needed, wouldn't it be better to remove it and simplify the code?

windymilla:
Yes. I don't think it's really worth it. I'll submit another commit tomorrow probably. Thanks

charliehoward4dp:
I would delete it, at least from the documentation, which would be the least work.  Thanks to Miller's matchless [PP instructions](https://www.pgdp.net/wiki/User:Miller), I've always kept a scratchpad file in each project folder, and never knew there was supposed to be one  managed by Guiguts until yesterday.

cpeel:
I think deleting it makes the most sense. Less is more!

windymilla:
OK - it's gone. Documentation of shortcuts is not now stored in the release. Documentation edits will take place in the manual when the next version is released.

srjfoo (approved):

PR #941, opened 2022-06-09
Improve word navigation, selection and deletion

### LINKED ISSUE #932: <ctrl>+<backspace> (delete word backwards) and  <ctrl>+<delete> (delete word forwards ) don't work as expected
**Describe the bug**
When using the keyboard shortcuts
    ctrl+Backspace -- delete word backwards
    ctrl+Delete -- delete word forwards
each fails to work as I expect them to work.

If the word is enclosed in any type of 'bracket', each shortcut will delete the first one it encounters plus any letter characters but stop at the opening/closing 'bracket' it encounters.

In the case of a word containing an apostrophe each shortcut behaves differently and does not delete the whole word, including the apostrophe. I describe how to demonstrate this behaviour below.

**To Reproduce**
1. Open Guiguts with 
[grocers-apostrophe.txt](https://github.com/DistributedProofreaders/guiguts/files/8743606/grocers-apostrophe.txt)
2. Position cursor just before the word **grocers'** and do ctrl+Delete. It will delete all but the apostrophe.
3. Undo that edit and position the cursor just after the apostrophe. Then do ctrl+Backspace and it will delete the whole of the word **grocers'**, including the apostrophe. This is inconsistent with the action in (2).
4. Now position the cursor just before the word **banana's** and do ctrl+Delete. It will delete the word up to the apostrophe.
5. Undo that edit and position the word at the end of the word **banana's**. Then do ctrl+Backspace and it will delete the letter "s" then stop. It is at least consistent with the action in (4) but in both cases I would expect the whole word to be deleted, including the apostrophe.

**Expected behavior**
If a word contains an apostrophe then I expect both forward and backward deletes to delete the whole word, including the apostrophe.

Where a word is enclosed in brackets I would expect either no deletion at all or if deletion is done then both brackets are deleted with the word they enclose.
 
**Environment**
* Guiguts version:  1.4.0-test4
* OS and version:  Windows 10
* Perl version: 
* Perl distribution: 

**Screenshots**

**Additional context**
If the word to be deleted contains certain non-ASCII characters such as Greek letters then the whole word is deleted, However inclusion of other non-ASCII characters in a word, such as the 'dagger' symbol (&#8224;), causes the actions for an apostrophe described above.

windymilla:
Navigating by word using Ctrl+arrow keys, selection using Shift+Ctrl+arrows
or mouse double-clicks, and deletion with Ctrl+Delete/Backspace, were
inconsistent in their interpretation of a "word".
New version mostly emulates behavior of MS Word. Word boundary is
beginning of sequence of word characters or sequence of non-word
characters.

Fixes #932

### DIFF
--- a/src/lib/Guiguts/KeyBindings.pm
+++ b/src/lib/Guiguts/KeyBindings.pm
@@ -71,27 +71,18 @@ sub keybindings {
     keybind(
         '<Control-BackSpace>',
         sub {
-            $textwindow->addGlobStart;
-            my $pos = $textwindow->search( '-backwards', '-regexp', '--', '\W',
-                'insert -1c', 'insert linestart' );
-            if ($pos) {
-                $pos = "$pos +1c";
-            } else {
-                $pos = 'insert linestart';
-            }
-            $textwindow->delete( $pos, 'insert' );
-            $textwindow->addGlobEnd;
+            my $end = $textwindow->index('insert');
+            $textwindow->SetCursor('insert wordstart');
+            $textwindow->delete( 'insert', $end );
         },
         '<<BackSpaceWord>>'
     );
     keybind(
         '<Control-Delete>',
         sub {
-            $textwindow->addGlobStart;
-            my $pos = $textwindow->search( '-regexp', '--', '\W', 'insert +1c', 'insert lineend' );
-            $pos = 'insert lineend' unless $pos;
-            $textwindow->delete( 'insert', $pos );
-            $textwindow->addGlobEnd;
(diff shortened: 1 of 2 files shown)


cpeel (approved):
Code-wise this LGTM. I would appreciate @srjfoo kicking the tires on this as an end-user if she has time, otherwise I'm happy to merge it in.

srjfoo (approved):
Tires duly kicked. It seems to work as advertised on a Mac, also.

cpeel:
Thanks Sharon!

PR #942, opened 2022-06-09
Add Combining Characters to Compose Key Sequences

windymilla:
PPers are being encouraged to use combining characters when a precomposed
Unicode character does not exist.
Add the more common characters to the available Compose Sequences.
Each is introduced with a plus sign, e.g. `+~` for combining tilde, `+-` for
combining macron, etc.

Also, add ability to provide a description in the Help dialog, to make it
clearer which character can be created with the Compose Sequence.

### DIFF
--- a/src/lib/Guiguts/CharacterTools.pm
+++ b/src/lib/Guiguts/CharacterTools.pm
@@ -693,7 +693,7 @@ sub composekeyaction {
     my $str    = $::lglobal{composepopstr};    # Get string typed so far in dialog
 
     if ( $::composehash{$str} ) {              # Does it match one of the defined compose sequences?
-        insertit( substr( $::composehash{$str}, 0, 1 ) );    # First character only - remainder is optional help text
+        insertit( $::composehash{$str} );
         ::killpopup('composepop');
     } elsif ( $str =~ s/^(([\\0]?x)|U\+)?([0-9a-f]{4})$/$3/i ) {    # or 4 digit hex, (optional \x, 0x, x or U+)
         insertit( chr( hex($str) ) );
@@ -706,7 +706,7 @@ sub composekeyaction {
         } else {
             my $term = "\n";
             if ( $::composehash{ $str . $term } ) {                                      # Is it a string with a forced terminator, e.g. Greek betacode
-                insertit( substr( $::composehash{ $str . $term }, 0, 1 ) );              # First character only - remainder is optional help text
+                insertit( $::composehash{ $str . $term } );
             } else {
                 insertit($str);                                                          # just insert the string as it is
             }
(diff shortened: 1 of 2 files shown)


src/lib/Guiguts/CharacterTools.pm
```diff
@@ -746,66 +746,78 @@ sub composeinitialize {
...
+    composeinitsyms( "\x{2018}",                  '\'', '<', '\'', '6' );     # left single quote
+    composeinitsyms( "\x{2019}",                  '\'', '>', '\'', '9' );     # right single quote
+    composeinitsyms( "\x{201C}",                  '"',  '<', '"',  '6' );     # left double quote
+    composeinitsyms( "\x{201D}",                  '"',  '>', '"',  '9' );     # right double quote
+    composeinitsyms( "\x{201A}",                  '\'', ',' );                # low single quote
+    composeinitsyms( "\x{201B}",                  '\'', '^' );                # high reversed single quote
+    composeinitsyms( "\x{201E}",                  '"',  ',' );                # low double quote
+    composeinitsyms( "\x{201F}",                  '"',  '^' );                # high reversed double quote
+    composeinitsyms( "\x{B1}",                    '*',  '+' );                # plus/minus
+    composeinitsyms( "\x{B7} Middle dot",         '.',  '^', '*', '.' );      # middle dot
```
cpeel, line 781: It took me entirely too long to figure out what was going on with the additional text you've added to some of these. I kept trying to figure out how it was consumed (or not) by the function. Eventually I realized these were consumed (or just ignored?) by the changes in lines 696 and 709. I grok what you're doing and the limitations with the current functions, but this is super duper confusing.

Is there a better / different way we can associate the optional help text with the character? Maybe a secondary global hash that associates the character with help text?
```perl
composeinitsyms( "\x{B7}",         '.',  '^', '*', '.' );      # middle dot
composeinithelp( "\x{B7}", "Middle dot");
```
?

rtonsing:
I haven't really used this feature before, but need it. I find it confusing to figure out which combining character to use.

It says above "+- for combining macron", but for me both "+=" and "+-" produce Greek characters, "+=" immediately, with "+-" I have to click OK.

Windows 11 OS, latest pulled code + this file.

windymilla:
@rtonsing, I don't see that effect in my branch. I can think of 3 possible options, that you may be able to eliminate before we try other tests:
1. You don't have the correct code: CharacterTools.pm should have calls to a routine `composeinitcombining` in lines 799-810
2. You have some other "compose key" software installed which is intercepting the characters?
3. You aren't using the feature as it was intended:
    a. Type some characters into the main text window
    b. Press AltGr (and release it) - a small dialog pops up
    c. Type `1/3` into that dialog - the dialog vanishes and the character `⅓` appears in the text window. This should work in any 1.3 or 1.4 release.
    d. Type the letter `a` into the main window, so the cursor is positioned immediately after the `a`.
    e. Use AltGr again, and type `+-` into the small dialog
    f. A macron should appear over the `a`. This should only work in with the code from this branch.
    g. Depending on font, OS, etc., the macron may appear to the right of the `a`
    h. Click in the status bar at the bottom, where it says `Dec: nnn Hex: xxx`, and it changes to also show you the name of the character you are positioned at
    i. Use the arrow keys to step back before the `a` and macron, it should show "Latin Small Letter A" when before the `a`, then one step to the right, the cursor will not move if the font is displaying the macron above the `a` correctly, but the status bar will display "Combining Macron"

src/lib/Guiguts/CharacterTools.pm
windymilla: I felt guilty writing the code :)
Should have gone with my instinct to change it. Will get onto it now

windymilla:
@rtonsing, a comment from @srjfoo gave me an idea. When you said `+-` gave you Greek, I now realise you perhaps meant, "a letter followed by an minus sign", i.e. you thought the plus sign meant "any letter". In fact `+-` means literally that: "type a plus sign followed by a minus sign". It just creates the combining character - it does not make a precomposed (canonically composed?) form. So if you follow an `a` with a combining macron, you end up with an `a` followed by a combining macron (which should display as an `a` with a macron over it, but is two separate characters like this: ̄a. This is different to the single character "Latin Small Letter A with Macron": ā
At the moment, Guiguts makes no attempt to normalise where a PPer has used a combining character but there is a precomposed version available. I might enter a separate issue to consider adding that feature, since we want characters to be normalised.

rtonsing:
1. You don't have the correct code: CharacterTools.pm should have calls to a routine composeinitcombining in lines 799-810
	It does.
2. You have some other "compose key" software installed which is intercepting the characters?
	No.
3. You aren't using the feature as it was intended:
	True. I was typing the character into the box. It seems to be working as you describe now.

Note that if I use Right Alt key, the next key I type is lost. So to enter '+-' I type the '+' twice. Probably an OS behavior - Windows 11. Doesn't happen if I use the menu.

windymilla:
> Note that if I use Right Alt key, the next key I type is lost. So to enter '+-' I type the '+' twice. Probably an OS behavior - Windows 11. Doesn't happen if I use the menu.
You can redefine which key to use if Right Alt doesn't work for you. It's in Preferences->Processing->Set Compose Key...
Hope that's helpful

src/lib/Guiguts/CharacterTools.pm
windymilla: Added another commit, but hold off merging - may add another commit with a few changes to key sequences. Thanks

windymilla:
Please wait before merging, @cpeel, thanks

windymilla:
Final commit added, so ready to merge, assuming new code is OK. Many thanks

src/lib/Guiguts/CharacterTools.pm
cpeel: The new code is much clearer, thanks that update!

cpeel:
Looks great, thanks for your work on this!

PR #944, opened 2022-06-10
Add Cmd-Up/Down key bindings for Mac

windymilla:
Cmd-Up should move to top of file
Cmd-Down should move to bottom
Holding the Shift key should select instead of just moving


windymilla:
I haven't been able to test this request from @srjfoo, so would be grateful if you can check it on a Mac please.

srjfoo (approved):
Works beautifully on my Mac, thanks!

cpeel:
Nice!

PR #946, opened 2022-06-13
Add option to normalize selected text

windymilla:
For HTML to validate, text needs to be in Unicode Normalization Form C.
PPer might have composed a character by using a combining character when
in fact there was a precomposed character they should have used.
When they get the error from the validator, they can run the new command
on selected text. The reason for doing selected text, rather than the whole
file, is there are rare circumstances, e.g. Greek oxia, where normalizing it
may not be what an expert Greek PPer wants.

### DIFF
--- a/src/lib/Guiguts/CharacterTools.pm
+++ b/src/lib/Guiguts/CharacterTools.pm
@@ -3,13 +3,14 @@ use strict;
 use warnings;
 
 use Unicode::UCD 'prop_invmap';
+use Unicode::Normalize;
 
 BEGIN {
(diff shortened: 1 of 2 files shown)


cpeel:
> The reason for doing selected text, rather than the whole
> file, is there are rare circumstances, e.g. Greek oxia, where normalizing it
> may not be what an expert Greek PPer wants.

This was my exact question when I started reading it. And upon reflection I can't believe I forgot about that edgecase given how much trouble it was during our conversion to Unicode!

PR #948, opened 2022-06-14
Expand compose sequences for combining characters

windymilla:
Add combining characters "below" to match existing "above" versions.
Introduce "below" with underscore, "above" with plus.
Match the characters used in the DP proofing guidelines for diacriticals,
e.g. comma for cedilla, open parenthesis for inverted breve, etc.
This will make it more logical and memorable for PPers, e.g. if faced with
`[x:]` from DP rounds, meaning diaeresis below `x`, they use `x` followed by
ComposeKey `_:`


windymilla:
After suggestions from @srjfoo...
Hopefully, this scheme is more logical for DP users.

srjfoo (approved):
Looks good to me.

PR #955, opened 2022-07-03
Set kindlegen path to an absolute path (macOS)

tangledhelix:
The default path to kindlegen on startup (without a setting.rc) is
`Applications/Kindle Previewer 3.app/Contents/lib/fc/bin/kindlegen`
when run on macOS.

Without a leading `/` or `./` or `../` Guiguts will consider this
a relative path. Unless the user happens to start guiguts while
in the root directory, this will never work.

This patch adjusts the default value for `$kindlegencommand`
to include the leading `/`

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -1128,8 +1128,8 @@ sub initialize {
         $::kindlegencommand = ::setdefaultpath(
             $::kindlegencommand,
             ::catfile(
-                'Applications', 'Kindle Previewer 3.app',
-                'Contents',     'lib', 'fc', 'bin', 'kindlegen'
+                '/Applications', 'Kindle Previewer 3.app',
+                'Contents', 'lib', 'fc', 'bin', 'kindlegen'
             )
         );
     } else {


windymilla:
The edit looks fine to me, though I have no way of testing it. However, at the moment it fails the perltidy check, so that needs fixing before merging.

tangledhelix:
That's weird... I didn't even touch the line it complains about in the output. Anyway, I'll try to fix up push an update.

cpeel:
`/` isn't a file part, though. I would change it to `/Applications` which would mirror what we do for `\Program Files (x86)` higher up.

And yes, perltidy is a frustrating piece of technology.

cpeel (approved):
Thanks for finding & fixing this!

tangledhelix:
Two notes just as FYIs.

First, this is the output I got with the existing (relative, non-functional) path:

```
12:44:57: Guiguts 1.3.1: If you have any problems, please include any error messages that appear here with your bug report.
12:45:26: Beginning ebookmaker
12:45:26: Files will appear in the directory /Users/dan/dp/pp/fisherboy/
12:45:26: For mobi files, use Set File Paths to locate kindlegen which you can obtain by installing Kindle Previewer 3
X connection to /private/tmp/com.apple.launchd.ch3S5hvyau/org.xquartz:0 broken (explicit kill or server shutdown).
```

Fixing the path changed the error, but on my system at least, it didn't fix the crash. I'm not sure, but I think this could be yet another arm64/amd64 problem. I saw similar issues before XQuartz shipped a universal binary.

```
12:45:49: Guiguts 1.3.1: If you have any problems, please include any error messages that appear here with your bug report.
12:46:09: Beginning ebookmaker
12:46:09: Files will appear in the directory /Users/dan/dp/pp/fisherboy/
[xcb] Unknown sequence number while processing queue
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
Assertion failed: (!xcb_xlib_threads_sequence_lost), function poll_for_event, file xcb_io.c, line 272.
```

I also attempted to set the kindlegen value to the below, with no luck. (`arch` can be used to force the platform used at runtime on an arm64 system with the Rosetta emulator.)

```
arch -x86_64 /Applications/Kindle Previewer 3.app/Contents/lib/fc/bin/kindlegen
```

I'm not sure how much it's worth chasing this; if people can't use the built-in ebookmaker, they can still use ebookmaker online easily enough.

windymilla (approved):
Looks good

cpeel:
> Fixing the path changed the error, but on my system at least, it didn't fix the crash. I'm not sure, but I think this could be yet another arm64/amd64 problem. I saw similar issues before XQuartz shipped a universal binary.

I suspect this is an issue with how perl is spawning an external program and communicating back via the file descriptors -- I'm not confident that it's explicitly related to kindlegen / ebookmaker.

Sometimes it feels like this really great program is built on a very shake tech foundation -- perl & tcl are very long in the tooth and not well supported these days. I have a faint hope that @willcohen's work on getting us moved to the better supported perl/tcl framework will improve things but that's going to be a bigger lift for GG compared to guiprep.

windymilla:
> Sometimes it feels like this really great program is built on a very shake tech foundation -- perl & tcl are very long in the tooth and not well supported these days. I have a faint hope that @willcohen's work on getting us moved to the better supported perl/tcl framework will improve things but that's going to be a bigger lift for GG compared to guiprep.
Rewriting the whole thing completely using something more modern would be good, but a massive task for one person - it's great to see one or two more people getting involved in the development, so maybe there's hope for the future of GG after all.

tangledhelix:
> I suspect this is an issue with how perl is spawning an external program and communicating back via the file descriptors -- I'm not confident that it's explicitly related to kindlegen / ebookmaker.

Maybe so, but my instinct was that it was an arm64 / x86_64 problem. I didn't think of this before, but due to my current project of building a Guiguts.app, I actually have both arm64 and x86_64 setups for Guiguts right now, so I tried the same test using the x86_64 rig. In that situation, kindlegen runs successfully.

cpeel:
It also works fine on my Intel Mac, so I believe you that it's arm64 related, but I suspect it's not unique to kindngen / ebookmaker and might be a larger problem cross-instruction set.

PR #957, opened 2022-07-10
Add `\IP` extension to regex syntax to insert page label

windymilla:
`\IP` gets replaced with the page number from the book by using the labels
from the pagenumbers hash.

The primary identified use of this is to better automate the naming of links
and their associated ids, e.g. for chapter headings or illos.

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -2442,10 +2442,13 @@ sub poetrynumbers {
     }
 }
 
+#
+# Accepts an optional index to find the page number at that index
+# Otherwise returns page number at index of the current insert position
 sub get_page_number {
     my $textwindow = $::textwindow;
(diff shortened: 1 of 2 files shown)


src/lib/Guiguts/Utilities.pm
```diff
@@ -2442,10 +2442,13 @@ sub poetrynumbers {
...
 }
 
+#
+# Accepts an optional index to find the page number at that index
+# Otherwise returns page number at index of the current insert position
 sub get_page_number {
     my $textwindow = $::textwindow;
     my $pnum;
-    my $markindex = $textwindow->index('insert');
+    my $markindex = shift // $textwindow->index('insert');
```
cpeel, line 2451: I did not know about the `//` operator! TIL!

cpeel (approved):

PR #958, opened 2022-07-14
Implement --home flag

tangledhelix:
Adds a -`-home` flag, which takes an argument (a path). When invoked with `--home <path>`, Guiguts will store its data & settings files in the specified directory rather than in the Guiguts code directory.

This may be desired because:

1. It may make it easier to maintain settings between upgrades
2. I am working on a macOS app bundle for Guiguts; these bundles are signed, and the contents can't change or the signature is invalidated. This flag permits us to write our metadata outside of the app bundle directory into another location,       e.g. `$HOME/.guiguts/`

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -326,14 +326,14 @@ $lglobal{runtests} = ( @ARGV == 1 and $ARGV[0] eq 'runtests' );
 
 # We need to check @ARGV for --home now, before initialize() and readlabels()
 # are run, otherwise we'll write files in the wrong location
-if (@ARGV and $ARGV[0] eq '--home') {
+if ( @ARGV and $ARGV[0] eq '--home' ) {
     shift @ARGV;
     die "ERROR: malformed command (--home must specify a path)\n" unless @ARGV;
 
-    $lglobal{homedirectory} = ::rel2abs(shift @ARGV);
-    ::infoerror("Resolved --home to " . $lglobal{homedirectory});
+    $lglobal{homedirectory} = ::rel2abs( shift @ARGV );
+    ::infoerror( "Resolved --home to " . $lglobal{homedirectory} );
 
-    if (-e $lglobal{homedirectory}) {
+    if ( -e $lglobal{homedirectory} ) {
         die "ERROR: --home directory must be a directory\n" unless -d $lglobal{homedirectory};
     } else {
         die "ERROR: --home directory could not be created\n" unless mkdir $lglobal{homedirectory};
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -324,6 +324,24 @@ local $SIG{__WARN__} = \&::warnerror;
 # otherwise it will load setting.rc which could influence the test results.
 $lglobal{runtests} = ( @ARGV == 1 and $ARGV[0] eq 'runtests' );
 
+# We need to check @ARGV for --home now, before initialize() and readlabels()
+# are run, otherwise we'll write files in the wrong location
+if (@ARGV and $ARGV[0] eq '--home') {
+    shift @ARGV;
+    die "ERROR: malformed command (--home must specify a path)\n" unless @ARGV;
+
+    $lglobal{homedirectory} = ::rel2abs(shift @ARGV);
+    ::infoerror("Resolved --home to " . $lglobal{homedirectory});
+
+    if (-e $lglobal{homedirectory}) {
+        die "ERROR: --home directory must be a directory\n" unless -d $lglobal{homedirectory};
+    } else {
+        die "ERROR: --home directory could not be created\n" unless mkdir $lglobal{homedirectory};
+    }
+
+    die "ERROR: --home directory is not writeable\n" unless -w $lglobal{homedirectory};
+}
+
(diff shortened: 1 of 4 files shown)


src/guiguts.pl
```diff
@@ -324,6 +324,24 @@
 # otherwise it will load setting.rc which could influence the test results.
 $lglobal{runtests} = ( @ARGV == 1 and $ARGV[0] eq 'runtests' );
 
+# We need to check @ARGV for --home now, before initialize() and readlabels()
+# are run, otherwise we'll write files in the wrong location
+if ( @ARGV and $ARGV[0] eq '--home' ) {
```
cpeel, line 329: Assuming all args are the first one isn't very extensible (and yes I see that `runtests` above does this too). IT would be far better if we used an existing package for processing command lines, like `GetOpt::Long`. I'm not sure if this is a built-in or if it's a cpan package which might shift the scales a bit.

src/guiguts.pl
```diff
@@ -364,8 +382,8 @@
 $top->repeat( 200, sub { _updatesel($textwindow) } );
 
 # Ready to enter main loop
-unless ( -e 'header.txt' ) {
-    ::copy( 'headerdefault.txt', 'header.txt' );
+unless ( -e ::catfile( $::lglobal{homedirectory}, 'header.txt' ) ) {
```
cpeel, line 385: Rather than sprinkle the variable around it would probably be better to have a function that would return `::catfile( $::lglobal{homedirectory}, 'header.txt' )` if the homedir was set and otherwise return `::catfile( $::lglobal{guigutsdirectory}, 'header.txt' )`.

windymilla:
@srjfoo, please could you test on Mac - thanks

srjfoo:
> @srjfoo, please could you test on Mac - thanks

@windymilla, yes, I plan to 🙂  -- keeping an eye on the discussion.

src/guiguts.pl
tangledhelix: Getopt::Long has been a core module since the vey beginning of Perl 5. I can look at changing the arg processing to use it.

src/guiguts.pl
tangledhelix: I'm not sure what you mean here. In some cases you need one, and in some you need the other. If you don't use `--home` then they both point to the same path.

We could replace these uses of `::catfile()` with a wrapper function, but we'd need two of them, one that returns `guigutsdirectory` and another that returns `homedirectory` (if `--home` was used), else `guigutsdirectory`. That would be a little less verbose in the code at each occurrence maybe, but it wouldn't change that you need to know which path you want to target at each use. I'm not sure it buys us anything, honestly...

srjfoo:
I suspect that I'm just doing something wrong. I downloaded a zip of this code (I think). I created a `temp` directory in my home directory, specified that when I invoked gg with 
```
perl guiguts.pl --home ~/temp &
```
The code recognizes that `temp` is in my home directory; `header.txt`, `setting.rc` and `labels_en.rc` are all created and all located in that directory.

What confuses me is that I figured that an easy way to verify that it was using the specified files would be to have two different instances of GG in two different places with different settings, like different background colors. I've stalled out on my first copy, because even though I set a background color (also turned line numbers on), it's not being saved. What am I missing?

(In my main dev copy of GG I was able to successfully persist the line numbers and background color change.)

tangledhelix:
> What confuses me is that I figured that an easy way to verify that it was using the specified files would be to have two different instances of GG in two different places with different settings, like different background colors. I've stalled out on my first copy, because even though I set a background color (also turned line numbers on), it's not being saved. What am I missing?

I was able to reproduce this as well, I don't think you did anything wrong.

In my testing, I didn't test colors; I was using things like path to bookloupe and other utilities, which so far as I could tell were persisting fine across restarts. Maybe not all settings are initialized the same way, because setting `$activecolor` (the button highlight color) reset when I started Guiguts a second time.

This is what I saw in `~/temp/setting.rc`:

* Starting value: `$activecolor              = '#24baec';`
* After setting the button highlight color: `$activecolor              = '#8b8435e92616';`
* After quitting Guiguts: `$activecolor              = '#8b8435e92616';`
* After opening Guiguts with the same `--home` again: `$activecolor              = '#24baec';`

I haven't worked out why yet, but it did reset on application start. I did another test with `$globalviewerpath` and got this result again, which I'm pretty sure worked in previous testing. Seems I introduced a bug along the way.

I'll review this and try again.

tangledhelix:
> Could you add a routine that returns the full path of the `header.txt` file, another that returns the `setting.rc` path, and one for the language labels file? Each of those is used in several places, it would be preferable to restrict their names and path logic to fewer places.

I'll look at that. I wondered if there were more than those 3 files to worry about, so far I haven't found any (other than settings.err, which is written if the settings file is broken, can't be loaded).

windymilla:
I think those are the three that can get normally written under the release  during a run of the program. 
I think the temporary files and output logs used for things like HTML validation get written into the folder that contains the file being checked, so not under the release - this is in `ErrorCheck.pm`.
I can't see anything else other than settings.err as you said, and stuff under `tests` which no regular user would ever do.

Issue #954, opened 2022-06-25
Make it easier to transfer settings from an old release


windymilla:
Maybe it could check if the current release has a setting.rc file (early on).
If not, it could pop a dialog to browse for an old release to get settings from - cancel to avoid doing this.
Alternatively, ask first "Do you want to copy settings from an previously installed GG release?"
As it copies them across, it could make any necessary changes between releases.

tangledhelix:
Have you considered having a pairing of files, one global, one a user-override? I'll explain what I mean by that:

- Ship `setting.rc` with Guiguts, which is meant to be static, and only includes the default value for each setting
- Guiguts doesn't edit `setting.rc` when the user changes settings
- If the user changes any setting from the default, `setting-user.rc` is created and the value is stored there
- Perhaps `setting-user.rc` could be stored in the user's home directory, not in the app directory
- On startup, first load `setting.rc`, then load `setting-user.rc` - most things would be default values, but `setting-user.rc` would override some values

I am thinking ahead here to what I would desire if I continue with my own project to create a Mac app bundle. The way a Mac app bundle works is that it's a (static) bundle of files. It's arranged in a directory tree, but in a special format that the OS considers to be an "app" and treats as one object you can drag around, launch, delete... but it's not supposed to contain any changing parts, because the app must be signed. So in that model, you would always want the contents of the app directory to be static, and any user settings to be in another file. In macOS, the settings files can either live in a Library directory in $HOME or they can be unix-style dotfiles... but the important thing is they are not a part of the application as-shipped.

So, this idea would become important in a Mac app, but I can see the value to separate these two layers in all situations, on all platforms, because the current upgrade handling of setting.rc can be non-intuitive for users (IMHO).

I'd be willing to look at creating a PR to implement this, if the idea seems reasonable to you.

windymilla:
That's an interesting idea with some definite benefits. Here are a few thoughts:

1. I don't think there would be any point in having a static `setting.rc` with default values. This is because `setting.rc` is not currently shipped with the code - all the default values are in the main Perl code (variables generally need initialising anyway, and `setting.rc` is just a bit of Perl code that gets run, setting the value of some of those variables). Guiguts writes `setting.rc` during your first run of the program. So I think equivalent to what you're suggesting would be to simply write `setting.rc` outside the release, e.g. the user's home directory, instead of inside the release.
[...]

PR #959, opened 2022-07-18
Add --home to specify location for persistent data

tangledhelix:
## Changes

Adds a `--home` flag, taking a directory as an argument, which specifies a directory for storing persistent user data. At this time, three things would be stored in this directory:

 - `setting.rc`
 - `header.txt`
 - `labels_{lang}.rc`

In the case of a corrupted `setting.rc`, then this location is also where `setting.err` might be written.

This flag is optional, the default behavior is preserved.

Changes argument parsing to use Getopt::Long. Options `--home` and `--runtests` are parsed with Getopt::Long. If a filename is specified, that is still handled by reading `@ARGV`.

**Note that this means a new arg `--runtests` replaces the older form `runtests`.**

Also adds helper functions to return various paths:
 - `header.txt`
 - `headerdefault.txt`
 - `labels_{lang}.default.rc`
 - `labels_{lang}.rc`
 - `setting.rc`

Some of the helpers are to simplify things with the variable location (whether `--home` is in use or not), some are just for the sake of convenience.

## Purposes

As described in #954, one purpose is to allow an easy way for users to carry their settings from one release to the next without manually copying files around.

I am working on a macOS app bundle to simplify the install and upgrade process for Mac users. Since Mac apps must be signed, their contents can't change; the `--home` flag permits user data to live outside of the application's directory, making this possible.

## Testing

I tested these changes on macOS. Since macOS uses a similar filesystem interface, Linux probably also works, but I didn't test directly.

I have no Windows machines handy to test with, so I'd appreciate if someone could give it a try on Windows.

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -20,7 +20,8 @@ BEGIN {
       &getprojectid &setprojectid &viewprojectcomments &viewprojectdiscussion &viewprojectpage
       &scrolldismiss &updatedrecently &hidelinenumbers &restorelinenumbers &displaylinenumbers
       &enable_interrupt &disable_interrupt &set_interrupt &query_interrupt &soundbell &busy &unbusy
-      &dieerror &warnerror &infoerror &poperror &BindMouseWheel &display_manual);
+      &dieerror &warnerror &infoerror &poperror &BindMouseWheel &display_manual
+      &path_settings &path_htmlheader &path_defaulthtmlheader &path_labels &path_defaultlabels);
 
 }
 
@@ -337,6 +338,39 @@ sub dos_path {
     return Win32::GetShortPathName( $_[0] );
 }
 
+# Return paths to common files used by the application.
+sub path_settings {
+    return ::catfile( $::lglobal{homedirectory}, 'setting.rc' );
+}
+
+sub path_htmlheader {
+    return ::catfile( $::lglobal{homedirectory}, 'header.txt' );
+}
+
+sub path_defaulthtmlheader {
+    return 'headerdefault.txt';
+}
+
+sub path_labels {
+
+    # If we're using --home (a separate data directory), then we store the
+    # labels file directly there
+    if ( $::lglobal{homedirectory} ne $::lglobal{guigutsdirectory} ) {
+        return ::catfile( $::lglobal{homedirectory}, "labels_$::booklang.rc" );
+    }
+
+    # Otherwise it's stored in a subdirectory data/ from Guiguts' directory
+    return ::catfile( $::lglobal{guigutsdirectory}, 'data', "labels_$::booklang.rc" );
+}
+
+sub path_defaultlabels {
+    my $f = ::catfile( 'data', "labels_$::booklang" . "_default.rc" );
+    return $f if -e $f;
+
+    # Default to English if $::booklang has no defaults file
+    return ::catfile( 'data', 'labels_en_default.rc' );
+}
+
 # system(LIST)
 # (but slightly more robust, particularly on Windows).
 sub run {
@@ -523,10 +557,9 @@ sub deaccentdisplay {
 }
 
 sub readlabels {
-    my $labelfile        = ::catfile( 'data', "labels_$::booklang.rc" );
-    my $defaultlabelfile = ::catfile( 'data', "labels_$::booklang" . "_default.rc" );
-    $defaultlabelfile = ::catfile( 'data', 'labels_en_default.rc' ) unless ( -e $defaultlabelfile );
-    @::gcviewlang     = ();
+    my $labelfile        = path_labels();
+    my $defaultlabelfile = path_defaultlabels();
+    @::gcviewlang = ();
 
     # read the default values first, in case some are missing from the user file
     ::dofile($defaultlabelfile);
@@ -870,6 +903,11 @@ sub initialize {
     $::manualhash{'wfpop'}             = '/Tools_Menu#Word_Frequency';
     $::manualhash{'workpop'}           = '#Overview';
 
+    $::lglobal{guigutsdirectory} = ::dirname( ::rel2abs($0) )
+      unless defined $::lglobal{guigutsdirectory};
+    $::lglobal{homedirectory} = $::lglobal{guigutsdirectory}
+      unless $::lglobal{homedirectory};
+
     ::composeinitialize();
 
     ::readsettings();
@@ -1085,9 +1123,6 @@ sub initialize {
         'Wrongspaced singlequotes',
     );
     $::gsopt[$_] = $::mygcview[$_] for 0 .. $#::mygcview;    # Use default gc/bl view settings
-
-    $::lglobal{guigutsdirectory} = ::dirname( ::rel2abs($0) )
-      unless defined $::lglobal{guigutsdirectory};
 
     # Find tool locations. setdefaultpath handles differences in *nix/Windows
     # executable names and looks on the search path.
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -353,9 +353,19 @@ sub path_defaulthtmlheader {
 
 sub path_labels {
 
+    my $homedirectory    = $::lglobal{homedirectory};
+    my $guigutsdirectory = $::lglobal{guigutsdirectory};
+
+    # Windows and macOS have case-insensitive filesystems. Lowercase
+    # the paths before comparing them
+    if ( $::OS_WIN or $::OS_MAC ) {
+        $homedirectory    = lc $homedirectory;
+        $guigutsdirectory = lc $guigutsdirectory;
+    }
+
     # If we're using --home (a separate data directory), then we store the
-    # labels file directly there
-    if ( $::lglobal{homedirectory} ne $::lglobal{guigutsdirectory} ) {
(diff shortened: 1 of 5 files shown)


tangledhelix:
I fixed what I could for perltidy, but now it's hung up on what appears to be whitespace? I found one line with trailing whitespace (fixed), but now it's complaining about two locations that appear to already be correct... if someone else can spot what it's unhappy about, please let me know...

tangledhelix:
@windymilla The Actions workflow is only actually complaining about Utilities.pm, and only line 354... but there **is** a blank line after `path_labels`, and there's no trailing space on it, etc... for the life of me I can't see what it's complaining about here...

https://github.com/DistributedProofreaders/guiguts/runs/7398845301?check_suite_focus=true

tangledhelix:
I decided to take @windymilla 's suggestion and let format_files do what it wanted... but now there are many more errors, so that seems to have gone worse than where I started from 😣

windymilla:
OK, partial memory pointed me in the right direction and found the explanation and solution

The [README](https://github.com/DistributedProofreaders/guiguts/blob/master/README.md) says for developers,

> Guiguts uses [perltidy](https://metacpan.org/pod/Perl::Tidy) for consistent styling using configuration options in `.perltidyrc`. 
> You can install perltidy with `cpanm --installdeps --notest --with-develop .` including the dot at the end.

On running that, it installed perltidy version 20210111, which is the one used by the Actions, as can be seen in the log for the `perltidy --version` Action. This local command and the Action are both controlled by [cpanfile](https://github.com/DistributedProofreaders/guiguts/blob/master/cpanfile) which contains the perltidy version number.

Using that version of perltidy, all of your files check OK, except Utilities, which requires a blank line after line 354
`sub path_labels {`
and some spaces removing from the otherwise blank line 909 (which becomes 910) with the above edit), just before the `::composeinitialize();` line.

Hope that helps - if you install the 20210111 version of perltidy, you should be able to use `format_files` script to do the tidying for you in future. At some point we can pin the version of perltidy to a later release for Actions and all GG developers would need to re-run the above command to install the correct one locally.

tangledhelix:
Thanks @windymilla - worked like a charm. The CI is all green now.

srjfoo:
First round of testing (sorry to be so slow).

Downloaded code to the desktop. This time I just used the default `perl guiguts.pl &` command to open it, to see what the default does. I opened GG, typed in some text, saved, and although the text file saved, GG quit:
```
[1]+  Segmentation fault: 11  perl guiguts.pl
```
I brought GG back up, opened the file, typed in some more text, and saved successfully; no segfault. However, when I quit GG, brought it up again, and saved another new document, I got a segfault again. `setting.rc` and `header.txt` were created successfully, but `labels_en.rc` was not.

More testing mañana.

tangledhelix:
@srjfoo I can reproduce what you saw, but I have two things to add:

1. For me, `data/labels_en.rc` **was** created
2. I get that behavior also on the master branch, so whatever's going on there isn't related to this PR as far as I can tell

srjfoo:
> 

Perhaps I got the wrong file, then. The zip I downloaded is `guiguts-home_flag.zip`, and I got it by going to [tangledhelix:home_flag](https://github.com/tangledhelix/guiguts/tree/home_flag), clicking on the "Code" button and selecting "Download ZIP" (rather than doing this through my main dev install). I then `cd` into the `src` subdirectory and run guiguts from there. So, perhaps I need a change in procedure -- or I'm downloading the wrong thing?

srjfoo:
> 

Okay -- a little bit more awake now. 😁 

I have now tested with `master`, also, and confirm that I get a segfault on initial save using the default "save" on `master`, also. If, instead of a "save", I do a `save as` for that initial save, it does not segfault.

I've also confirmed that `labels_en.rc` exists -- it's just not where I expected it to be. It seems to be staying in the `data` subfolder, and overwriting the `labels_en_default.rc` file, instead of landing in the directory I invoke gg from, whereas in my previous test, when I specified a directory for the settings files, it was saved in that directory.

I just checked my master dev branch (which has _not_ been updated to the current `master`) and it also saves `labels_en.rc` in the `data` subfolder, but does not appear to overwrite the default? At least, I assume so, but I may be mistaken about that. At any rate, both `labels_en.rc` and `labels_en_default.rc` are there.

One final note: I brought up gg using my dev master and tried the save button to do the initial save. The file was created, but the program crashed, but this time, the error message was:
```
[1]+  Bus error: 10           perl guiguts.pl
```

rtonsing:
Did some testing in Windows 11, seems to work OK.

windymilla:
> One final note: I brought up gg using my dev master and tried the save button to do the initial save. The file was created, but the program crashed, but this time, the error message was:
> 
> ```
> [1]+  Bus error: 10           perl guiguts.pl
> ```
Also reproduced on Windows - have added issue #960 since it doesn't appear related to this PR

windymilla:
Initial tests on Windows 11 working for me - needs a bit more time

tangledhelix:
> I've also confirmed that `labels_en.rc` exists -- it's just not where I expected it to be. It seems to be staying in the `data` subfolder, and overwriting the `labels_en_default.rc` file, instead of landing in the directory I invoke gg from, whereas in my previous test, when I specified a directory for the settings files, it was saved in that directory.

The path difference is known / intended. When `--home` is used, I wrote all the files in the home directory without any subdirectories. But if `--home` is not used, I left the paths the way they have always been. In general, my design principle here was to not change anything unless `--home` was specified.

The only exception to that rule was that `runtests` became `--runtests` when Getopt::Long was introduced. But that seemed acceptable (perhaps @windymilla or @cpeel can tell me if it's not).

If `labels_en_default.rc` is in fact being overwritten, that is not intended. The desired behavior is:

```
if labels_{booklang}.rc does not exist:
    if labels_{booklang}_default.rc exists:
        copy labels_{booklang}_default.rc to labels_{booklang}.rc
    else
        # if no defaults for this language, then use English as global default
        copy labels_en_default.rc to labels_{booklang}.rc
    end
end
```

srjfoo:
> If `labels_en_default.rc` is in fact being overwritten, that is not intended. The desired behavior is:
> 
> ```
> if labels_{booklang}.rc does not exist:
>     if labels_{booklang}_default.rc exists:
>         copy labels_{booklang}_default.rc to labels_{booklang}.rc
>     else
>         # if no defaults for this language, then use English as global default
>         copy labels_en_default.rc to labels_{booklang}.rc
>     end
> end
> ```

My bad on that -- I must have been blind. I ran the tests again (deliberately invoking the bus error/segfault) and `labels_en_default.rc` is not actually overwritten.

Thanks for the explanation of where things live -- that makes sense.

src/lib/Guiguts/Utilities.pm
```diff
@@ -337,6 +338,39 @@ sub dos_path {
...
+
+sub path_defaulthtmlheader {
+    return 'headerdefault.txt';
+}
+
+sub path_labels {
+
+    # If we're using --home (a separate data directory), then we store the
+    # labels file directly there
+    if ( $::lglobal{homedirectory} ne $::lglobal{guigutsdirectory} ) {
```
windymilla, line 358: Although it would be a bizarre thing to do with our current plans for the use of this feature, I could envisage a future use might end up using the `--home` argument, but with the `guigutsdirectory` directory as the argument. On Windows, folders resolve OK even if the case is wrong, so passing in `\dp\guiguts` or \DP\guiguts` as the `homedirectory` refer to the same folder. Only one of these will be `eq` to the  `guigutsdirectory`, so the treatment of the labels file will be different in the two cases.
I'm not sure this matters, and may not be worth the extra code complexity of saying "On Windows, do a case-insensitive comparison". I did look at `URI::File` but although it gives a "canonical" version of a folder name, it doesn't change case. 
So, just an observation, unless you have a good idea for avoiding it simply.

tangledhelix: I see what you mean. Macs also have case-insensitive filesystems(*), so on both Mac and Windows we should compare the path without considering case (i.e. with `lc`). I'll look at it this weekend if I have time to test it.

(*) There's technically an edge case where someone could run Guiguts from a Mac case-sensitive volume, but you'd have to really go out of your way to do this and I don't think it's worth checking for it.

srjfoo: > (*) There's technically an edge case where someone could run Guiguts from a Mac case-sensitive volume, but you'd have to really go out of your way to do this and I don't think it's worth checking for it.

I agree, but it might be worth a comment line that a case-insensitive OS is assumed? Periodically, I've been tempted to do a case-sensitive install, but have always chickened out. 😁

tangledhelix: I just pushed a commit that does the case-insensitive comparison when one of (OS_WIN, OS_MAC) are true.

windymilla: That all looks good to me, thanks

windymilla (approved):
I'm happy with all of this now if you are too @srjfoo

srjfoo:
> 

Looks good to me. I went back to my original quick test of changing background color and button highlight color and show/hide line numbers, and all that seemed to be working just fine (I haven't tested out everything; just those three).

I created two different "`home`" directories, each with different settings, one all-caps, and flipped between them, invoking the all-capped directory with lower-case. Settings on bringing up GG were correct for the setting file I used.

If there's anything else I should be testing, let me know.

srjfoo (approved):

PR #961, opened 2022-07-24
Fix crash when `Save As` is triggered by a `Save`

### LINKED ISSUE #960: Crash when saving new file immediately after opening GG
**Describe the bug**
Opening GG and using Save sometimes generates a Bus error or Segmentation fault on Macs and crashes silently on Windows

**To Reproduce**
1. Open Guiguts
2. Use File--> Save
3. Type new filename, e.g. `abc.txt`

windymilla:
If the user created a new file, then tried to `Save` the file, it detected that
it hadn't been saved before, and so called the `Save As` routine.

Both the `Save` and `Save As` (and incidentally `Save As Copy`) routines
temporarily disabled the autosave feature using the Perl `local` keyword,
which stores a backup of a global variable, restoring it when the current
block is exited.

It appears that this mechanism in Perl does not work if a global variable is
saved, then in a subroutine called from within the block (i.e. while the `local`
is still in effect) the global variable is `local`ed again. Perl crashed when
the inner block was exited.

Recoded to backup to a `my` scope variable, with a few adjustments to
ensure that the restoring lines couldn't be skipped by early exit of the
routines.

Fixes #960

### DIFF
--- a/src/lib/Guiguts/FileMenu.pm
+++ b/src/lib/Guiguts/FileMenu.pm
@@ -69,7 +69,7 @@ sub file_saveas {
         -initialfile => $initialfile,
     );
 
-    if ( defined($name) and length($name) and not bad_filename_chars($name)) {
+    if ( defined($name) and length($name) and not bad_filename_chars($name) ) {
         $::top->Busy( -recurse => 1 );
         $textwindow->SaveUTF($name);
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
No crashes or error messages to report!  😁

PR #963, opened 2022-07-25
Display filename correctly when `Saving a Copy As` on Mac

windymilla:
Unlike `Save As`, `Save a Copy As` displayed the full path instead of just
the filename in the filename field of the Save dialog. This did not happen
on Windows (possibly a native Windows dialog is used which copes).
Also, on all OS, if the current file had not yet been saved, the filename
field was set to "No File Loaded" in the Save a Copy dialog only.

Much of the code in the two routines that did these operations was the
same, so the routines have been combined into one. Also, the title of the
dialog now says "Save As" or "Save a Copy As" accordingly.


srjfoo (approved):

cpeel (approved):
Nicely done!

PR #967, opened 2022-08-01
Add Spell Query tool

### LINKED ISSUE #916: Add a new spellcheck tool
On Windows in particular, the current Aspell-based spellcheck has limitations, since Windows Aspell has not been maintained for almost 20 years. 

Add a new tool (to replace the current one, or maybe retain the current one as "Old Aspell-based spellcheck") based initially on the algorithm and word lists in ppspell (used for old? PP workbench) - see [here](https://github.com/wf49670/dpwb). 

1. Spellcheck button would find all the potential spelling errors and display them in the errorcheck dialog, similar to the output of jeebies, bookloupe, etc.
2. Main dictionary would be based on the document's language. Multi-lingual spellcheck would also be available.
3. User would have a "supplementary" words file that they could add words to - they would need to copy it across when installing a new release
5. The project dictionary would function as currently
6. A goodwords list could be easily added to the project dictionary as currently.
7. In the errorcheck dialog, clicking in different ways would perform different actions - there would also need to be extra buttons across the top to do these actions, and if possible, shortcuts (like Ctrl+i) would still work:
a. Left-clicking on the misspelled word would jump to it in the file
b. Ctrl-Left would add the word to the project dictionary and remove all occurrences of it from the errors list ("Add to Proj dict")
c. Shift-Ctrl-Left would add to the supplementary words file and remove all occurrences of it from the errors list ("Add to global dict")
d. Right-click would remove the query from the errorcheck dialog ("Skip this occurrence")
e. Shift-Ctrl-Right would remove all occurrences of it from the errors list ("Skip all")

Hunspell (a more-recently-maintained tool similar to Aspell)

windymilla:
1. Spell Query tool behaves like most other tools, producing a list of queries
2. Queries in the list can be selected in any order and disposed of in a
similar way to existing spelling tool, i.e. skip this occurrence, skip all, add word
to project/global dictionary. Shortcut available by combining Shift/Ctrl with
Left/RIght clicking on list, similar to behavior for bookloupe, jeebies, etc.
3. Dictionary is SCOWL-based word list. At present, only an English word list
is included with the release. User could add any other language, simply by
putting a wordlist in the data folder, e.g. `dict_fr_default.txt` for French.

Fixes #916 
Comments on points made in Issue #916
1. Done
2. Word list is based on document language. No intention to provide (rarely used) multi-lingual spelling at the moment. Could be easily done by user concatenating the English and French word lists for example, and using that as the dictionary.
3. Instead of supplementary words file, it adds to the end of the global language word list
4. Done
5. Done
6. Done


srjfoo (approved):
Looks good as far as I can tell, but my testing skills are probably pretty basic on this sort of thing. Let me know if there's some type of exotic testing you want me to do.

src/lib/Guiguts/ErrorCheck.pm
```diff
@@ -1117,6 +1204,130 @@ sub booklouperun {
...
+            }
+        }
+
+        my $fh;
+        unless ( open $fh, "<:encoding(utf8)", $dictname ) {
+            ::warnerror("Error opening Spell Query dictionary: $dictname");
+            return 0;
+        }
+
+        while ( my $line = <$fh> ) {
```
cpeel, line 1292: This approach is not ideal for a few reasons:
1. we load the file into a hash on every spelling run
2. we need to load the file into memory at all

This may not be worth optimizing, but using something like `DB_File` and converting the default dict into a DBM file on first load would solve several of these problems and likely be just as fast with much less memory overhead.

https://www.oreilly.com/library/view/programming-the-perl/1565926994/ch02s07.html
https://www.oreilly.com/library/view/perl-cookbook/1565922433/ch14s02.html

If we used a DBM file we'd need to check against 2 hashes -- the global one and the project one -- because adding the project words to the DBM file will change the DBM file itself and that should be global.

cpeel (commented):
Not a blocker, but some ideas for us to consider.

src/lib/Guiguts/ErrorCheck.pm
windymilla: Thanks for the suggestion and info - those are valid points, and I hadn't come across DBM files before.

I did initially consider the time cost of (re-)loading the hash on every run, and decided to try the simplest option first. It seemed to run quickly enough (detail below) so I didn't revisit it. I hadn't thought about the memory aspect - the dictionary contains 120K words, which would have seemed a lot to me when first coding back in 80s, but now seems quite small. As well as the disadvantage of having separate global and project hashes that you mentioned, another is that the current global dictionary text file is easily editable with any text editor and words added by the user from within GG will appear at the end of it. Also, typically spell checking would probably be run once or twice per PPing project, certainly not frequent multiple runs.

I've just been doing a few quick tests (without any sophisticated tools) to get some info on the speed/memory implications:

1. Speed - with a typical text file (13K lines long), spell query takes less than 1 second to load the hash, perform the spell query and pop up the dialog populated with the list of queries.
2. With a very small text file (i.e. to eliminate the time actually checking spelling), it takes less than 0.5 seconds to clear the hash, recreate the hash from the dictionary file and pop the dialog.
3. For comparison, the old spell check dialog also takes about 1 second to pop up on my typical text file

4. Memory - just using the Windows Task Manager for a rough approximation, it appears that loading the dictionary into a hash and spellchecking uses about 20MB of memory. Multiple uses do not increase this.
5. For comparison, running the old spell check on the typical text file appears to also use about 20MB
6. Also for comparison with other (bloated) Windows apps I have running at the moment, the Task Manager itself uses 30MB, the Slack app is using about 400MB, as is the WhatsApp desktop app, and Chrome (with 6 tabs open) is on just over 1000MB

I don't know what you think, given that info, but it seems to me that in this case, the potential benefits of using a DBM file probably don't outweigh the disadvantages of additional code complexity and lack of user editability.

windymilla: I've had some feedback from a couple of "real" users, both of whom have requested a frequency count. The first idea that comes to mind to implement that is to use the value in the hash table: initialize it to zero, then add one each time the word is encountered. I'd have to do something different if we go for the DBM file idea.

windymilla: That idea doesn't work (obviously as soon as I started to code it!) since you are trying to keep count of how many times a word occurs that is **not** in the dictionary, so you can't store that in the dictionary hash

cpeel: Thanks for doing that testing and thinking through the pros/cons. I agree that we should go with the simple solution you have rather than adding complexity for no actual value!

cpeel (approved):

windymilla:
I've made some relatively small improvements following tester's comments, if you don't mind re-reviewing please:
1. Each queried spelling is followed by the number of times it occurs
2. If spelling occurs more times than a threshold (default 3) it is assumed to
be good
3. Improved handling of apostrophe/single quote - if word is not OK with
apostrophe/quote(s) it is checked without apostrophe/quote(s)
4. Just the queried word is highlighted, rather than from the start of the word
to the end of the line as is done with other tools.
5. Ordinals such as `1st`, `13th`, `254th` are accepted (also `sc` is OK - small caps markup)

windymilla:
@cpeel & @srjfoo - Apologies for the multiple commit efforts in this PR - let me know if you want me to do something different, like start a new PR.

cpeel (approved):
Sorry for the delay on the re-review, been a busy few days.

The code LGTM. I did not test-drive it however.

srjfoo (approved):

PR #969, opened 2022-08-04
Allow prep text files to have spaces in names

windymilla:
File names are sometimes of the form `title - 001.txt` rather than just
`001.txt`. Allow non-word characters at start of file name as long as
there are word characters just before the file extension.

Fixes issue reported by PM using Abby files.

**Edited to clarify**, this PR is to do with importing, not exporting prep text files.

### DIFF
--- a/src/lib/Guiguts/FileMenu.pm
+++ b/src/lib/Guiguts/FileMenu.pm
@@ -131,7 +131,7 @@ sub file_import_preptext {
     $::globallastpath = $directory;
 
     for my $file ( sort @files ) {
-        if ( $file =~ /^(\w+)\.txt/ ) {
+        if ( $file =~ /(\w+)\.txt$/ ) {
             $textwindow->ntinsert( 'end', ( "\n" . '-' x 5 ) );
             $textwindow->ntinsert( 'end', "File: $1.png" );
             $textwindow->ntinsert( 'end', ( '-' x 45 ) . "\n" );


srjfoo (commented):
Not entirely sure what's going on.

Rather than going looking for a TIA set that included a spacey filename in the page separator, I created my own, so I tried three options: 

1. standard numeric filename.extension
2. spacey filename where there is no space between the filename and the number
3. spacey filename where there is a space also between the filename and the number

For #2, the pages were exported, but the page separator line was included at the top of each text file. For 1 and 3, it worked as I expected, with no page separator line showing. #2 may not be an expected form, but should probably be covered?

srjfoo:
> For #2, the pages were exported, but the page separator line was included at the top of each text file. For 1 and 3, it worked as I expected, with no page separator line showing. #2 may not be an expected form, but should probably be covered?

I just ran another set of tests because there's something funky happening that I think is unrelated that I was trying to describe.[1] I'm not getting #3 to export at all with a filename like `a b c 00001.png`. The spacey filename I used on the earlier test had the same number of spaces, but more non-space characters. #2 is still exporting and still with the page separator at the top of the file.

[1: When creating the spacey filenames in the page separators, I've been using BBEdit, not GG, since I have a much better understanding of regexes in BBEdit. Which means that the `.bin` file is either now disjoint, or nonexistent if I created it from the first file using BBEdit. I need to do some more testing to see if this is what is causing the `a b c 00001.png` pattern to fail completely. Something is also, at some point, blocking the Import TIA OCR option -- I can bring up the dialog, but it can no longer detect the .gz files.]

windymilla:
Oops - I realise looking back at your comments @srjfoo and the original PR description (I'll edit it to clarify) that it is not at all clear that this PR is to do with **importing** prep text files, not exporting. It's in File/Content Providing/Import Prep Text Files. Sorry for the confusion.

No-one should be attempting to **export** prep text files with spaces in names, but the original PM's report was that sometimes their prep text files that they wanted to **import** had spaces in the names.

To test, you could take some prep text files, e.g. `001.txt`, `002.txt`, `003.txt`, and rename them something like `test 001.txt`, etc. With the old code, it would let you select the folder to import from, but nothing would import. Hopefully with the new code, it should import as normal, with separator lines saying 001, 002, 003, etc.

srjfoo:
> Oops - I realise looking back at your comments @srjfoo and the original PR description (I'll edit it to clarify) that it is not at all clear that this PR is to do with **importing** prep text files, not exporting. It's in File/Content Providing/Import Prep Text Files. Sorry for the confusion.

Aaah, no it never occurred to me that the text files' filenames would be spacey -- I assumed that (1) CP/PMs using GG know better than to put spaces in filenames, and (2) have a program that will rename sequential files (or at least remove prefixes).

> No-one should be attempting to **export** prep text files with spaces in names, but the original PM's report was that sometimes their prep text files that they wanted to **import** had spaces in the names.

😲

srjfoo (approved):
Works perfectly

cpeel (approved):

PR #970, opened 2022-08-06
Add search capability to spell query dialog

windymilla:
In a similar manner to the Word Frequency dialog, allow user to pop the
S&R dialog with the queried word in it - this allows them to review all the
occurrences (of which there will be no more than the threshold number) to
decide if it's a good/bad spelling.

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -11,6 +11,7 @@ BEGIN {
 
 my @errorchecklines;
 my %errors;
+my $APOS = "\x{2019}";
 
 # General error check window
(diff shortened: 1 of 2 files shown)


src/lib/Guiguts/ErrorCheck.pm
```diff
@@ -11,6 +11,7 @@ BEGIN {
 
 my @errorchecklines;
 my %errors;
+my $APOS = "\x{2019}";
```
cpeel, line 14: Looking at this afresh in this PR, you might add a comment that this is the right curly quote/apostrophe. Maybe it's just been a long day but it took me three shakes too long to pick up what this was.

cpeel (approved):

srjfoo (approved):

PR #972, opened 2022-08-09
Use `Command` key for `Alt` key shortcuts on Mac

windymilla:
Mac keyboards do not have an `Alt` key, but several shortcuts are set up to
use the `Alt` key. Use `Command` key (referred to internally as `Meta`) in
place of `Alt` on Macs.

Menu buttons should show Command/Alt as appropriate when
displaying shortcuts (for example `Search` menu, `Highlight Character, String or Regex`)

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -104,8 +104,10 @@ our $urlprojectpage       = 'https://www.pgdp.net/c/project.php?id=';
 our $urlprojectdiscussion = 'https://www.pgdp.net/c/tools/proofers/project_topic.php?project=';
 
 ### Application Globals
-our $activecolor      = '#24baec';    #'#f2f818';
+our $activecolor      = '#24baec';                      #'#f2f818';
 our $alpha_sort       = 'f';
+our $altkey           = $OS_MAC ? "Meta"    : "Alt";    # Alt key used in binding routine calls
+our $altkeyname       = $OS_MAC ? "Command" : "Alt";    # Name of Alt key for display to user
 our $auto_page_marks  = 1;
 our $auto_show_images = 0;
 our $autobackup       = 0;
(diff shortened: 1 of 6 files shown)


windymilla:
@srjfoo, please could you test whether you can set the Compose key to involve the Command key: Prefs, Processing, Set Compose Key.
The default key is Control-M. If you uncheck Control, and check Command, then Command+M should pop the Compose dialog instead. It maybe that Command+M is already something important that you can't override - if so, type a different letter into the Set Compose Key dialog and check Command.

src/guiguts.pl
```diff
@@ -104,8 +104,10 @@
 our $urlprojectdiscussion = 'https://www.pgdp.net/c/tools/proofers/project_topic.php?project=';
 
 ### Application Globals
-our $activecolor      = '#24baec';    #'#f2f818';
+our $activecolor      = '#24baec';                      #'#f2f818';
```
cpeel, line 107: Any idea what this color in comment is for?

cpeel (approved):
Code changes look good -- I didn't testdrive it.

srjfoo:
> @srjfoo, please could you test whether you can set the Compose key to involve the Command key: Prefs, Processing, Set Compose Key. The default key is Control-M. If you uncheck Control, and check Command, then Command+M should pop the Compose dialog instead. It maybe that Command+M is already something important that you can't override - if so, type a different letter into the Set Compose Key dialog and check Command.

🤣 -- I tried (not paying enough attention to what I was trying) to set `command`+`m` for the compose key. But it's the (relatively) global shortcut for minimize. So guess what happened to the compose dialog. Bear with me -- I think I need to look at all the shortcuts, and that will take a bit of time.

src/guiguts.pl
windymilla: No idea - maybe the previous value before a former developer changed it?
I'll remove the comment in another commit - I won't commit it yet in case @srjfoo has other suggested changes.

srjfoo (approved):

PR #976, opened 2022-08-22
Upgrade to ebookmaker 0.12

windymilla:
New ebookmaker uses Calibre instead of kindlegen to convert epubs to Kindle
formats.

Although there will be ebookmaker 0.12.1, 0.12.2, etc., probably before the next GG release, I'd like to get these code changes merged in. Upgrading to later ebookmaker versions will just be a case of editing the version in `package.sh`.

### DIFF
--- a/tools/ebookmaker/package.sh
+++ b/tools/ebookmaker/package.sh
@@ -11,7 +11,7 @@ mkdir $DEST
 cp README.md $DEST
 
 if [[ $OS == "win" ]]; then
-    VERSION=0.11.30
+    VERSION=0.12.0
     URL=https://github.com/DistributedProofreaders/ebm_builder/releases/download/v$VERSION/ebookmaker-$VERSION.zip
     curl -L -o ebookmaker.zip $URL
     unzip ebookmaker.zip -d $DEST
--- a/tools/ebookmaker/package.sh
+++ b/tools/ebookmaker/package.sh
@@ -11,7 +11,7 @@ mkdir $DEST
 cp README.md $DEST
 
 if [[ $OS == "win" ]]; then
-    VERSION=0.12.0
+    VERSION=0.12.3
     URL=https://github.com/DistributedProofreaders/ebm_builder/releases/download/v$VERSION/ebookmaker-$VERSION.zip
     curl -L -o ebookmaker.zip $URL
     unzip ebookmaker.zip -d $DEST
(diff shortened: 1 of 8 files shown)


windymilla:
This has been tested on Windows only, by me and 3 other testers. To test on another platform, you would have to install one of the `0.12.*` versions of ebookmaker locally and probably configure the path to ebookmaker using Prefs, Set File Paths

cpeel:
It looks like a few doc files need to be updated too.

`INSTALL.md` references kindlegen a couple of times. I know the new version of ebookmaker removed the use of tidy, so I think we can also remove the reference to `tidy.conf` in `tools/ebookmaker/README.md`.

cpeel (approved):
Docs look good to me 👍 

The Mac instructions probably need to be re-vetted with the new version but I'm fine handling that outside of this PR since I can't promise to get to it soon.

tools/ebookmaker/package.sh
```diff
@@ -11,7 +11,7 @@ mkdir $DEST
 cp README.md $DEST
 
 if [[ $OS == "win" ]]; then
-    VERSION=0.11.30
+    VERSION=0.12.0
```
srjfoo, line 14: Does this have to be an exact match, or can it use the `>=` form? I see that Eric has already merged 0.12.2.

(apologies for not having yet gotten the Mac testing done yet)

windymilla: No, it is an exact match, so that an exact version of ebookmaker is always included with a specific release of GG. It actually refers to a release of the ebm_builder repo, which contains a version of ebookmaker built as a Windows .exe file. At the moment those releases are named after the ebookmaker release they contain.

So, it's correct to be 0.12.0, for now which is the latest ebm_builder release, but at some point before the next GG release I will build another ebm_builder release with the latest ebookmaker - that process is pretty trivial now Eric is just making small changes to ebookmaker. Then I'll change the number in this GG file to match that release. When I build the GG Windows  release it will grab the ebookmaker exe from ebm_builder.

Sorry its a bit complicated. It was necessary due to some things in ebookmaker which need patching before the Windows exe is built. That meant it couldn't just grab straight from the ebookmaker repo, hence the need for ebm_builder. Theoretically we could name the ebm_builder release more generally, eg v12summer, and update that release with the patch versions Eric produces, eg 12.2, 12.3, etc. Then GG would refer to v12summer and so GG wouldn't need a PR if ebookmaker had a new release before GG is released. However, it would mean needing to keep track of versions carefully to avoid accidentally breaking a GG release by switching the ebookmaker release under its feet. I assume that's why @cpeel designed it as it is. It seems to work fine, but thanks for checking. Hope all of the above ramble makes sense (excuse phone typos)

srjfoo: Yep -- makes sense, thanks.

srjfoo (approved):

PR #979, opened 2022-08-26
Facilitate retention of settings between releases

windymilla:
Currently, users who want to retain settings, such as font size, have to copy
or edit the `setting.rc` file using the file from their previous release into
their new release. This also applies to customised `header.txt` files, language
label files and user dictionaries.

Building on #959, which allowed the user to specify a prefs folder with
a `--home` argument, this commit implements a default prefs folder as
follows:

1. If the --home argument is specified, it takes priority.
2. If not, then a default prefs directory is used if it already exists
3. If the default prefs directory doesn't exists, prefs are stored under the
release as was done historically.

To support the user in their one-off setup of the default prefs directory,
a new option in the Prefs->Set File Paths menu copies relevant files into
the default prefs directory, creating it if it doesn't exist.

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -3,6 +3,7 @@ use strict;
 use warnings;
 use POSIX qw /strftime /;
 use File::HomeDir qw /home/;
+use Getopt::Long;
 
 BEGIN {
     use Exporter();
@@ -408,7 +409,7 @@ sub path_defaultdict {
 #
 # Return path to default GG homedir location under user's home folder
 sub path_defaulthomedir {
-    if ($::OS_WIN) {
+    if ( $::OS_WIN or $::OS_MAC ) {
         return ::catdir( File::HomeDir::home(), "Documents", "GGprefs" );
     } else {
         return ::catdir( File::HomeDir::home(), ".GGprefs" );
@@ -3314,8 +3315,6 @@ sub BindMouseWheel {
 #
 # homedirectory must be handled before initialize(), so that
 # homedirectory and guigutsdirectory are both set correctly.
-use Getopt::Long;
-
 sub processcommandline {
 
     # Default values if not specified on command line
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -21,7 +21,8 @@ BEGIN {
       &scrolldismiss &updatedrecently &hidelinenumbers &restorelinenumbers &displaylinenumbers
       &enable_interrupt &disable_interrupt &set_interrupt &query_interrupt &soundbell &busy &unbusy
       &dieerror &warnerror &infoerror &poperror &BindMouseWheel &display_manual
-      &path_settings &path_htmlheader &path_defaulthtmlheader &path_labels &path_defaultlabels &path_userdict &path_defaultdict);
+      &path_settings &path_htmlheader &path_defaulthtmlheader &path_labels &path_defaultlabels &path_userdict &path_defaultdict
+      &processcommandline &copysettings);
 
 }
 
@@ -404,6 +405,16 @@ sub path_defaultdict {
     return ::catfile( 'data', "dict_$::booklang" . "_default.txt" );
 }
 
+#
+# Return path to default GG homedir location under user's home folder
+sub path_defaulthomedir {
+    if ($::OS_WIN) {
+        return ::catdir( File::HomeDir::home(), "Documents", "GGprefs" );
+    } else {
+        return ::catdir( File::HomeDir::home(), ".GGprefs" );
+    }
+}
+
 # system(LIST)
 # (but slightly more robust, particularly on Windows).
 sub run {
@@ -750,6 +761,13 @@ sub working {
 }
 
 sub initialize {
+
+    # Get location of guiguts.pl & make it the current directory
+    $::lglobal{guigutsdirectory} = ::dirname( ::rel2abs($0) );
+    chdir $::lglobal{guigutsdirectory};
+
+    sethomedir();
+
     $::top = ::tkinit( -title => $::window_title, );
     my $top = $::top;
     $top->minsize( 440, 90 );
@@ -790,11 +808,6 @@ sub initialize {
         -format => 'gif',
         -data   => $::icondata
     );
-    if ( $0 =~ m/\/|\\/ ) {
-        my $dir = $0;
-        $dir =~ s/(\/|\\)[^\/\\]+$/$1/;
-        chdir $dir if length $dir;
-    }
 
     # positionhash stores user's window position
     # geometryhash stores user's window position and size
@@ -935,11 +948,6 @@ sub initialize {
     $::manualhash{'versionbox'}        = '/Help_Menu#Guiguts_HELP_Menu:_Online_and_Built-in_Help';
     $::manualhash{'wfpop'}             = '/Tools_Menu#Word_Frequency';
     $::manualhash{'workpop'}           = '#Overview';
-
-    $::lglobal{guigutsdirectory} = ::dirname( ::rel2abs($0) )
-      unless defined $::lglobal{guigutsdirectory};
-    $::lglobal{homedirectory} = $::lglobal{guigutsdirectory}
-      unless $::lglobal{homedirectory};
 
     ::composeinitialize();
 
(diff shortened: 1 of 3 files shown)


windymilla:
@srjfoo and @cpeel - please could you check the suitability of the default folder location for non-Windows platforms. On Windows, it's `$HOME/Documents/GGprefs`; on other platforms, it's `$HOME/.GGprefs` (with a period). It's very easy to to change in the code of course, but the user doesn't have a choice over its location. Thanks

src/lib/Guiguts/Utilities.pm
```diff
@@ -404,6 +405,16 @@ sub path_defaultdict {
     return ::catfile( 'data', "dict_$::booklang" . "_default.txt" );
 }
 
+#
+# Return path to default GG homedir location under user's home folder
+sub path_defaulthomedir {
+    if ($::OS_WIN) {
+        return ::catdir( File::HomeDir::home(), "Documents", "GGprefs" );
+    } else {
+        return ::catdir( File::HomeDir::home(), ".GGprefs" );
```
cpeel, line 414: For Mac, it would be more user-friendly to place it in `$HOME/Documents/GGprefs` just like on Windows. dot-directories are hidden in Finder so `.GGPrefs` will make it more difficult to manage.

`.GGPrefs` is still a good location on Linux, however.

src/lib/Guiguts/Utilities.pm
```diff
@@ -3298,4 +3306,152 @@ sub BindMouseWheel {
...
 
+#
+# Process command line, which has to be done before call to initialize() in guiguts.pl
+#
+# runtests must be set before initialize(), otherwise it will load
+# setting.rc which could influence the test results.
+#
+# homedirectory must be handled before initialize(), so that
+# homedirectory and guigutsdirectory are both set correctly.
+use Getopt::Long;
```
cpeel, line 3317: While not required, it's usually a best practice to put all imports at the top of the file.

src/lib/Guiguts/Utilities.pm
srjfoo: Just out of curiosity, is there a reason for using upper case? I'm so conditioned at this point to using all lower case most of the time that upper case just looks wrong 😁 

That being said, @cpeel is right, for Mac users, unless you want to hide the directory, it's better not being a dot-directory, and `GGPrefs` is as good a name as any.

srjfoo:
As a general rule, I will likely specify my own preferred `--home` argument, but my purpose here was to test the option to let GG choose and place the homedir in a default location.

I tried it both with and without the ampersand to shove it into the background in the Terminal window and got the following responses:

>llita:src sharonjoiner$ perl guiguts.pl --home &
> [1] 27214
> llita:src sharonjoiner$ 23:41:29: Option home requires an argument
> Error in command line arguments
> perl guiguts.pl --home
> 23:41:42: Option home requires an argument
> Error in command line arguments
> [1]+  Exit 255                perl guiguts.pl --home

I tried it both ways because I wondered if the ampersand was messing things up, but that turned out not to be the case. (When I use `perl guiguts.pl --home ~/ggdata &`, it works just fine, but doesn't seem to be creating a default folder. Might it be a permissions issue?

src/lib/Guiguts/Utilities.pm
windymilla: Thanks both - certainly on Windows, mixed case is common, and Guiguts is often abbreviated to GG (not gg) - hence the name. If you feel strongly, @srjfoo, I can easily make it `GGprefs` on Windows `ggprefs` on Mac, and `.GGprefs` or `.ggprefs` on Linux. I've added another commit, but can easily add another if needed.

windymilla:
> > llita:src sharonjoiner$ perl guiguts.pl --home &
> > [1] 27214
> > llita:src sharonjoiner$ 23:41:29: Option home requires an argument
> > Error in command line arguments
> > perl guiguts.pl --home
> > 23:41:42: Option home requires an argument
> > Error in command line arguments
> > [1]+  Exit 255                perl guiguts.pl --home
> 
> I tried it both ways because I wondered if the ampersand was messing things up, but that turned out not to be the case. (When I use `perl guiguts.pl --home ~/ggdata &`, it works just fine, but doesn't seem to be creating a default folder. Might it be a permissions issue?

`--home` requires an argument. The way to use the default is to not specify the --home argument at all, i.e. `perl guiguts.pl` (optionally with `&`)

src/lib/Guiguts/Utilities.pm
srjfoo: No, I don't have any strong feelings about it -- I was just curious about why, and you answered that. And mixed case is common on macOS, too. I'd leave it as-is.

srjfoo:
> `--home` requires an argument. The way to use the default is to not specify the --home argument at all, i.e. `perl guiguts.pl` (optionally with `&`)

Then there's something that I'm obviously not understanding about getting the `GGPrefs` directory created.

    If the --home argument is specified, it takes priority.
    If not, then a default prefs directory is used if it already exists
    If the default prefs directory doesn't exists, prefs are stored under the
    release as was done historically.

Specifically, the second point -- "If not, then a default prefs directory is used if it already exists." I was expecting that GG would create this directory, with the proper permissions. Do I need to manually create it in the correct place for it to be used?

windymilla:
Sorry, maybe slight communication over the word argument. I meant if --home (with a dir name) is specified, thinking of "--home" as one of the available "command line arguments".
If you don't use that argument, it's true that it uses the default directory if it exists. It isn't intended that you create it manually though. That's what the new option in the prefs menu is for, so you can select which release has the settings, that you want to use for all future releases.
If you don't do that, then the third point applies, ie it behaves like old releases.
Hope that makes it clearer 😊

srjfoo:
Yes, I think it's a case of me misunderstanding what you're trying to say. Quoting for the sake of convenience ...

> 1. If the --home argument is specified, it takes priority.
> 2. If not, then a default prefs directory is used if it already exists
> 3. If the default prefs directory doesn't exists, prefs are stored under the release as was done historically.
>
> To support the user in their one-off setup of the default prefs directory, a new option in the Prefs->Set File Paths menu copies relevant files into the default prefs directory, creating it if it doesn't exist.

The first and third items I have no trouble with. It's the second one that I'm having trouble understanding. Perhaps because of the word "default". I think of the default as what you term the historical method in item 3.

Comparing this version with a previous version, I think I've figured out what you mean, though. It's actually under `Preferences -> File Paths`, at the same level as the `Set File Paths...` menu. `Use Settings From Another Release`?

I think what has me confused is the discussion of the best location for the `GGprefs` directory, because the `Use Settings From Another Release` dialog says "Choose top-level Guiguts release folder with settings you want to copy. Cancel now if you have unsaved changes!" (By the way, you have to make the dialog box reeeeally wide to read that whole line.)

So, success at last.

1. If the `--home` argument is specified, it takes priority.
2. If `--home <homedir>` is not specified, but preferences from another release have been specified using `Preferences -> File Paths -> Use Settings From Another Release`, a `GGprefs` directory will be created (see manual for platform-specific location of the directory) and used as a default location for customized preferences. The settings from the release specified will be copied there and used for the current version. 
3. If neither of the above, preferences will be stored under the release as was done historically.

Regarding option 2 -- does this only have to be done once? I.e. if, with future releases, GG detects the existence of the `GGpref` directory, will it be used automatically? If so, I assuming that the only way to zero things so that the user can go back to using the historical method to store prefs within the GG directory structure?

srjfoo (approved):

windymilla:
Apologies for the wrong menu option wording - you're right that it's the File Paths menu, not the Set File Paths option. Sorry for adding to the confusion!

> 2. If `--home <homedir>` is not specified, but preferences from another release have been specified using `Preferences -> File Paths -> Use Settings From Another Release`, a `GGprefs` directory will be created (see manual for platform-specific location of the directory) and used as a default location for customized preferences. The settings from the release specified will be copied there and used for the current version.

To be precise to clarify when the directory is being created...where you have "a `GGprefs` directory will be created" should be "a `GGprefs` directory will **have been** created". The creation takes place when you do  `Use Settings From Another Release`. The dir must already exist for case 2 to happen.

Also, when you say "case 2 only happens once", that's not really right. Clicking `Use Settings From Another Release` only happens once (for a regular user who isn't fiddling round with multiple releases for testing or some other nefarious reason). Case 2 happens every time they run GG from then on.

I not trying to be picky with words, just making sure you understand what's happening, and that I am aware of how the wording and the process could be misinterpreted by a user. When Charlie writes it up in the manual, I'm sure he will word it better.

srjfoo:
> Also, when you say "case 2 only happens once", that's not really right. Clicking `Use Settings From Another Release` only happens once (for a regular user who isn't fiddling round with multiple releases for testing or some other nefarious reason). Case 2 happens every time they run GG from then on.
> 
> I not trying to be picky with words, just making sure you understand what's happening, and that I am aware of how the wording and the process could be misinterpreted by a user. When Charlie writes it up in the manual, I'm sure he will word it better.

Yeah, I didn't word that as well as I should have. What I was trying to do was ask for confirmation that once a directory is created, it will be used permanently, including by any future release, as long as the directory is detected as being there, unless it's overridden (by invoking GG with `--home`) or the directory is deleted/renamed.

windymilla:
> Yeah, I didn't word that as well as I should have. What I was trying to do was ask for confirmation that once a directory is created, it will be used permanently, including by any future release, as long as the directory is detected as being there, unless it's overridden (by invoking GG with `--home`) or the directory is deleted/renamed.

That sounds completely accurate to me, thanks.

PR #981, opened 2022-08-27
Avoid using `-mustexist` for `chooseDirectory`

windymilla:
`-mustexist` argument doesn't work on Macs, so check manually instead


cpeel (approved):

srjfoo (approved):
Note that this does stop the dialog from freezing if a nonexistent directory is typed in. On Windows, a nonexistent directory alert is shown to the users. On macOS, there is not an alert, and the directory assigned is the one in which the most recent `scannos` directory resided when the dialog was opened. This is generally not the directory the user wants, but it's a Mac issue -- Mac users should just be very careful to make sure that any directory they want to set actually exists.

PR #982, opened 2022-08-27
Improve default tool locations on Macs

### LINKED ISSUE #971: Some default tool locations are no good for Mac users
**Describe the bug**
Aspell and XnView locations default to something under `Program Files` which is Windows not Mac

Improve these defaults, e.g.
XnViewMP installs as `/Applications/XnViewMP.app/Contents/MacOS/XnViewMP` on Macs
Aspell might be `/opt/homebrew/bin/aspell` - still to be checked on Intel Macs

windymilla:
XnView and Aspell had Windows-specific default install locations.
Override with better possibilities for Macs

Fixes #971

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -1177,6 +1177,7 @@ sub initialize {
 
     # If XnView or Aspell are installed under "Program Files (x86)",
     # set that as the default, otherwise use "Program Files"
+    # Suitable for Windows installation locations
     my $trypath = ::catfile( '\Program Files (x86)', 'XnView', 'xnview.exe' );
     if ( -e $trypath ) {
         $::globalviewerpath = ::setdefaultpath( $::globalviewerpath, $trypath );
@@ -1190,6 +1191,14 @@ sub initialize {
     } else {
         $::globalspellpath = ::setdefaultpath( $::globalspellpath,
             ::catfile( '\Program Files', 'Aspell', 'bin', 'aspell.exe' ) );
+    }
+
+    # Override to more likely default locations for Mac
+    if ($::OS_MAC) {
+        $::globalviewerpath = ::setdefaultpath( $::globalviewerpath,
+            ::catfile( '/Applications', 'XnViewMP.app', 'Contents', 'MacOS', 'XnViewMP' ) );
+        $::globalspellpath =
+          ::setdefaultpath( $::globalspellpath, ::catfile( '/opt', 'homebrew', 'bin', 'aspell' ) );
     }
 
     if ($::OS_MAC) {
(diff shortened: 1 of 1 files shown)


src/lib/Guiguts/Utilities.pm
```diff
@@ -1192,6 +1193,14 @@ sub initialize {
             ::catfile( '\Program Files', 'Aspell', 'bin', 'aspell.exe' ) );
     }
 
+    # Override to more likely default locations for Mac
+    if ($::OS_MAC) {
+        $::globalviewerpath = ::setdefaultpath( $::globalviewerpath,
+            ::catfile( '/Applications', 'XnViewMP.app', 'Contents', 'MacOS', 'XnViewMP' ) );
+        $::globalspellpath =
+          ::setdefaultpath( $::globalspellpath, ::catfile( '/opt', 'homebrew', 'bin', 'aspell' ) );
```
cpeel, line 1201: I'm curious where the aspell binary is for @srjfoo on her M1-based system. For my Intel-based system homebrew creates a link in `/usr/local/bin/aspell` which points to `/usr/local/Cellar/aspell/0.60.8/bin/aspell` and there is no `/opt/homebrew` directory at all. Homebrew appears to be installed in `/usr/local/Homebrew/` but `/usr/local/Homebrew/bin` doesn't include aspell.

windymilla: @srjfoo told me it as in the homebrew location. I could add another commit to check if it doesn't exist in one location I'll set the other, a bit like happens above in the windows code. That would then cope with both M1 and Intel. What do you think?

cpeel: That sounds like a great plan to me. It's frustrating that we need to dual-path so many things for Intel vs M1, but here we are.

srjfoo: From my research, the Homebrew crew made a specific point to use a different path setup on the M1 Macs from that used on the Intel Macs. Yes ... it was a bit of a surprise (I don't recall the reasoning at this point).

And yes, my M1 aspell is in `/opt/homebrew/bin/aspell`. Like you, my Intel aspell is in `/usr/local/bin/aspell`.

And yes, I like the idea of checking one place and using the other if it's not there.

windymilla:
OK - should be fine for either @cpeel or @srjfoo to check that default location for Aspell works on Intel-based Mac, now.

Bear in mind from now on when testing, that if you have created a default prefs folder, it will be used unless you either delete/rename it, or use the --home argument to set to a different folder. In particular, in this case, if you have previously set an Aspell location in your (default) prefs, you might think this latest commit is correctly (or incorrectly) finding the location, but it could be that it was just set in your prefs from previously. So really, you want to test using the default prefs that come with the release.

I wonder if, to make it easier for testing, it would be useful to have a `--nohome` argument, which forces the old behaviour, i.e. just uses the settings you have under the release, and ignores your default prefs folder or any `--home` argument. When you are testing a PR, the settings would then normally just be the defaults because you wouldn't have bothered to customise them within that PR version. It could avoid some confusion when testing and/or avoid you having to keep deleting/renaming the default folder, or keep giving new names with the `--home` argument to get a fresh start. What do you think, or have you not had enough experience of the settings system yet to express an opinion? Of course, that would go under a separate PR.

cpeel (approved):
Looks great - thanks!

srjfoo (approved):
XnViewMP and aspell correctly detected on both M1 and Intel Macs

PR #983, opened 2022-08-28
Add `--nohome` argument to override home directory usage

windymilla:
The `--nohome` argument forces GG to use the historical locations under
the release for setting.rc, header.txt, etc., ignoring the default prefs dir
or any `--home` argument (see #979).


cpeel (approved):
Nice. I see this has a very useful tool during validation and debugging. 👍

srjfoo (approved):
Many thanks!

PR #986, opened 2022-09-02
Make `<Shift>` `Next Occurrence` button reverse direction

### LINKED ISSUE #985: Make `Next Occurrence` button reversible with `<Shift>` in Stealth Scannos dialog
In the Stealth Scannos dialog (a version of the S&R dialog) the Search button reverses direction temporarily if the user Shift-clicks instead of clicking. The Next Occurrence button performs the same task and should have the same reverse feature.

windymilla:
In the Stealth Scannos dialog (a version of the S&R dialog) the Search
button reverses direction temporarily if the user Shift-clicks instead of
clicking.
The Next Occurrence button performs a similar task and should have the
same reverse feature.

Fixes #985

### DIFF
--- a/src/lib/Guiguts/SearchReplaceMenu.pm
+++ b/src/lib/Guiguts/SearchReplaceMenu.pm
@@ -1245,7 +1245,7 @@ sub searchpopup {
                 -padx   => 2,
                 -anchor => 'w'
             );
-            my $nextoccurrencebutton = $sf5->Button(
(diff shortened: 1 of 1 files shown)


windymilla:
Thanks "Past Nigel" - I'd totally forgotten that I'd previously written a routine that implements "reverse" for any search-type button, so just a call to that was needed :)

cpeel (approved):
Way to go Past Nigel!

srjfoo (approved):

PR #991, opened 2022-09-09
Fix virtual event crossover between WF and ErrorCheck

windymilla:
Apparently virtual events, e.g. `<<search>>` are "shared by all widgets of the same MainWindow" according to the Perl/Tk docs. Because both Word Frequency and ErrorCheck had a `<<search>>` event, if you popped WF first, then dismissed it, then popped ErrorCheck, e.g. to run bookloupe, when you right-clicked an entry in the list widget, it executed the search routine and popped the S&R dialog, since in the WF dialog, right-clicking was bound to `<<search>>`. However in the ErrorCheck dialog, right-clicking should `<<remove>>` the entry not do a search.

Fixed by giving the virtual event a unique name - you learn something every day :)

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -263,9 +263,9 @@ sub errorcheckpop_up {
     );
 
     # Alt/command + button 1 pops the Search dialog prepopulated with the queried word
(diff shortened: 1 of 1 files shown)


cpeel (approved):
TIL too!

srjfoo (approved):

Issue #993, opened 2022-09-09
Spell query sometimes highlights the wrong part of the line


windymilla:
**Describe the bug**
Given the line `barracks, young 'un!—in disgrace again`, spell query wants to query the word `un`.
However, it ends up reporting where the word starts as column 12, which is the `un` in `young`. This is because it's not looking for a whole word when it searches the original line for the queried string.

`$col = index( substr( $origline, $col + 1 ), $wd ) + $col + 1;` needs changing to something cleverer.

OR

Don't substitute **multiple** non-word characters for a single space, nor trim leading spaces. That way you could keep track of the column number as you process each word, rather than trying to find it later.

PR #995, opened 2022-09-10
Highlight queried spellings more accurately

windymilla:
When attempting to highlight where the queried word was on the line, the search found the first occurrence of the queried string, which might not be the queried word, e.g. given `young 'un`, it would highlight the letters `un` in the word `young`, rather than the whole word `un` that appears later.

Instead of trying to find the word on the line, new algorithm tracks the column as each word is checked. This necessitates keeping the line's alignment correct when replacing non-word characters, so each character is replaced with one space. Seems to be more intuitive than the old method, with just a minor tweak to the column when a leading apostrophe or single quote is trimmed.

The old regex for detecting and removing block markup didn't cope with the `/I[8.3,70]` case where margins are customized, so more complexity was needed. In fact removing block markup wasn't needed since it consists of a single letter, which is never flagged as a spelling query anyway. As only a small percentage of lines contain block markup, it's quicker to just spell check those, than to check every line to see it it has block markup.

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -263,9 +263,9 @@ sub errorcheckpop_up {
     );
 
     # Alt/command + button 1 pops the Search dialog prepopulated with the queried word
-    $::lglobal{errorchecklistbox}->eventAdd( '<<search>>' => "<$::altkey-ButtonRelease-1>" );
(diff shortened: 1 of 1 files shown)


windymilla:
Fixes #993

cpeel (approved):

srjfoo (approved):

PR #998, opened 2022-09-15
Update jeebies instructions

cpeel:
Update the jeebies build instructions to include moving the binary to locations outside of the Guiguts directory. This makes it such that it doesn't need to be rebuilt after a GG upgrade.

I confirmed on my Intel Mac that `mv jeebies /usr/local/bin` worked without `sudo`. @srjfoo I would appreciate you confirming that the instructions work on your Intel- and M1-based systems without `sudo` as well.

### DIFF
--- a/tools/jeebies/README.md
+++ b/tools/jeebies/README.md
@@ -22,3 +22,15 @@ file. A `Makefile` is included to make this brainless, however:
 ```
 make build
 ```
+
+### Moving to a better location
+
+Building Jeebies will create a `jeebies` executable in the `tools/jeebies`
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
Works fine copying the `jeebies` executable into `/opt/homebrew/bin` without sudo. And running `brew doctor` doesn't appear to object to it (that was my main concern).

srjfoo:
> Works fine copying the `jeebies` executable into `/opt/homebrew/bin` without sudo. And running `brew doctor` doesn't appear to object to it (that was my main concern).

Ditto my Intel Mac. (both copy without `sudo` and `brew doctor` doesn't object to it being there)

PR #1004, opened 2022-09-19
Improve Spell Query for LOTE

windymilla:
First commit

1. No longer require `default` dictionary - user can add support
for a language by creating a `user` dict in their GGprefs folder.
2. User can give list of languages (main one first, any non-word
character as delimiters) in Language setting dialog from status bar,
or File/Project menu, or field in Spell Query dialog.
3. Language list is stored in existing language variable, with access
routines to get list of languages, or get just the main language.
4. Language list is therefore stored in .bin file so specific to project.
This is better than old Spell Check multiple languages which are stored in
setting.rc, needing manual reset if user swaps projects.
5. To support French/Italian/etc, if word fails to match dictionary,
try trimming leading `l'` and `d'`

Second commit (distinct set of code changes)

Add Spell Query to the test suite. Also add some extra
words to exercise SQ at end of text check file, which necessitates minor
adjustments to some other baseline files.


windymilla:
Hopefully that's much clearer to read the code changes now - testing help notes in comment above
Thanks for review/test :)

src/lib/Guiguts/ErrorCheck.pm
```diff
@@ -1332,6 +1337,16 @@ sub booklouperun {
         return 0;
     }
 
+    #
+    # Return true if word is in dictionary: same case, lower case, or title case (e.g. LONDON matches London)
+    # Can't just do case-insensitive check because we don't want "london" to be OK.
```
cpeel, line 1342: Appreciate these comments - thank you!

src/lib/Guiguts/Utilities.pm
```diff
@@ -381,29 +381,33 @@ sub path_data {
...
+    my $lang = shift // ::main_lang();
+    return path_data( "dict_$lang" . "_user.txt" );
 }
 
 #
 # Return path to spell query default global dictionary for current language
+# Optional argument to specify language
 sub path_defaultdict {
-    return ::catfile( 'data', "dict_$::booklang" . "_default.txt" );
+    my $lang = shift // ::ain_lang();
```
cpeel, line 409: Typo of `::main_lang();` I believe

src/lib/Guiguts/Utilities.pm
```diff
@@ -381,29 +381,33 @@ sub path_data {
...
     return ::catfile( 'data', 'labels_en_default.rc' );
 }
 
 #
 # Return path to spell query user global dictionary for current language
+# Optional argument to specify language
 sub path_userdict {
-    return path_data( "dict_$::booklang" . "_user.txt" );
+    my $lang = shift // ::main_lang();
+    return path_data( "dict_$lang" . "_user.txt" );
```
cpeel, line 402: FWIW, this could be `"dict_${lang}_user.txt"` if you wanted (had to look up the perl syntax for this).

windymilla: Thanks - don't know why I didn't use that syntax - I have used it in the past.

src/lib/Guiguts/Utilities.pm
windymilla: Good spot - darned interpreted languages!

cpeel: Code review for the win!

cpeel (approved):

windymilla:
Testing notes:
I've added the relevant prefixes to  the end of this [Italian dictionary](https://github.com/DistributedProofreaders/guiguts/files/9633037/dict_it_user.zip), which you should unzip and put in your GGprefs folder. You should then be able to use Spell Query to check text like the following:
```
questo amico
quest'amico
quest' amico
quest amico
```
The first 3 should be accepted, but the 4th should query the word `quest` since it is not an Italian word.
Note that you will have to set your file's language to Italian via the status bar Lang button.

You can edit the Italian dictionary to see what other prefixes are valid, or to change them, or you could test that prefixes with typos get queried, e.g. `qxuest'amico`. You could also test capitalization variants, which should all pass for a word like `amici` that is not a proper noun requiring capitalization.

windymilla:
I hope to get the French and German dictionaries prepared for inclusion with the release. Any thoughts on whether to do that under different PR?

cpeel:
These changes look good to me. I have a slight preference for additional languages being in a separate PR.

windymilla:
Thanks - in that case, I think this PR is ready for you to test @srjfoo.
I'll put another one in once I have the fr & de files ready.

srjfoo (approved):
`perl guiguts.pl --runtests` on a Mac was not completely successful, but some of that is because of the things like the default location of jeebies in this PR being other than where I have it, and also being specified as an `exe` file. I think that's a tangential issue and we can try to figure out how to get it to run successfully all the way through on Macs later. For what it's worth, it did complete:
```
# Looks like you failed 13 tests of 76.
```

PR #1012, opened 2022-09-26
Remove jeebies' need for external `.jee` files

windymilla:
Instead of the jeebies executable loading `he.jee` and `be.jee`, adjust the build process to include the data within the source: process each `.jee` file to format it as initialization data for the `char * []` variables used in the source; include these generated `.inc` files in the C source.

There is now no need for the `.jee` files to be shipped with the Windows version, nor for any data files to be moved into a `bin` dir on other platforms. Once the executable is built by the user, they can move that to a suitable `bin` dir to make it available to all releases.


windymilla:
Added the fix for the Mac compiler warning as a separate commit. I suspect `flgets` function is now pretty much the standard `fgets` function, but I've resisted the temptation to fiddle beyond the fixes I'm trying to make. Don't know what you think about that - I don't want to break it.

windymilla:
OK - should be ready for review/testing.
The Windows executable has been recompiled under MinGW32, and has run successfully on a tester's Windows computer, giving the same number of queries.

tools/jeebies/Makefile
```diff
@@ -1,7 +1,10 @@
 all: build
 
-build: jeebies.c
+%.inc : %.jee
+	sed -e 's/^.\+$$/"&",/' $< > $@
```
srjfoo, line 4: Failed on both M1 and Intel Macs

```
sed -e 's/^.\+$/"&",/' he.jee > he.inc
sed: RE error: illegal byte sequence
make: *** [he.inc] Error 1
```

Sadly, I know no `sed`, so I'm no help in figuring out the problem.

windymilla: Hoping it was due to `he.jee` being ISO-8859-1 encoded and containing accented characters.
IMHO, those accented words are actually errors in the data file (but I'm not about to change them) - they are actually Spanish, (e.g. "he oído" means "I have heard") - there are several, where "he" is Spanish for "I have" rather than the English word.

windymilla: Should be ready for retesting - just to be sure, clean first:
```
make clean
make build
```

cpeel (approved):
This is really fantastic work, Nigel. The ratio of deletions to additions in the PR make my nerd heart happy too!

srjfoo (approved):

PR #1015, opened 2022-09-30
Use Spell Query for Word Frequency spell check

windymilla:
Only load SQ dictionaries if something has changed.
If new project loaded, or user changes language settings, need to
load SQ dictionaries, otherwise, can just use the hash already
populated.

Additional wrapper routine needed because WF's wordlist includes
hyphenated words and other pairs of words.

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -6,7 +6,7 @@ BEGIN {
     use Exporter();
     our ( @ISA, @EXPORT );
     @ISA    = qw(Exporter);
-    @EXPORT = qw(&errorcheckpop_up);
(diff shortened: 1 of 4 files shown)


windymilla:
Testing notes:
Use Tools->Word Frequency, then Check Spelling button in WF dialog.
Should display a list of dubious spellings with the number of times each occurs.
Should error like Spell Query does if the language(s) don't have dictionaries.
Previous version used Aspell - this version doesn't - you can test that by using Preferences->Set File Paths and setting the Aspell location to a non-existent name (note that will probably be reset to the correct location when you restart GG). Then Tools->Spell Check should error, but Tools->Spell Query and WF Check Spelling should not.

srjfoo (approved):

cpeel (approved):

PR #1019, opened 2022-10-06
Detect end-of-line he/be errors correctly

windymilla:
Previous PR #1012 broke the detection of he/be errors across line boundaries. The replacement of `flgets` with `fgets` didn't take account of the fact that `fgets` keeps the newline character.

Maybe I should have resisted the temptation to tidy up :)

Rather than process the line `fgets` returns, or alter the he/be processing to cope with line endings, it seems safest and simplest to go back to `flgets`

### DIFF
--- a/tools/jeebies/jeebies.c
+++ b/tools/jeebies/jeebies.c
@@ -85,7 +85,6 @@ void procfile(char *);
 char *getaword(char *, char *);
 char *getawordwithpunct(char *, char *);
 int matchword(char *, char *);
-char *flgets(char *, int, FILE *);
 void lowerit(char *);
 int gcisalpha(unsigned char);
 int gcisdigit(unsigned char);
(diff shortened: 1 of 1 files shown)


cpeel:
I'd encourage us to lean into standard functions (keep `fgets()`) and just remove the newlines ala https://stackoverflow.com/a/28462221

windymilla:
Hopefully that will be better - maybe it will even work on Macs!

srjfoo:
@windymilla, the following is a typical line pair from the output when run via the command line using the test file you provided:
```
(Bangalore), he rode that distance one day, and back\0x0D

    Line 6845 column 21 - Query phrase "Bangalore he rode" 
```
I ran it both via GG and via the command line, redirecting the output from the latter into a file. The full lines that were echoed into the file all had the `0x0D` at the end. No other lines had any undisplayable characters in them, and I used the BBEdit "Zap gremlins" function to translate the undisplayable characters into something semi-readable by humans.

When run via GG, I get the 117 results that you mentioned should be the correct count. When run via the command line, I get 118 results. These are the same results I saw with the previous version of `jeebies`, so at least that's consistent.

srjfoo (approved):
My profuse apologies for being off-base on my testing. I thought I was using the local-to-branch version of jeebies during my testing, but because during one of the previous testing sessions, I put a jeebies executable in `/opt/homebrew/bin/` what I was actually getting was that. I discovered that in my latest round of testing, actually used the correct executable, and got the expected results.

cpeel (approved):
From a git perspective, consider squashing this PR so it comes in with just the single commit message with the last change rather than the revert, un-revert, update.

PR #1020, opened 2022-10-08
Add W3C epubcheck tool to Guiguts

windymilla:
We already bundle the HTML and CSS validators. The epubcheck tool validates epub files. It is therefore slightly different to the other tools in that the user has to select a file to be checked, rather than it checking the currently loaded file.

Packaging seems to work fine when generating a release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
   English, French and German dictionaries are included with the release.
 - Word Frequency's Check Spelling button now uses Spell Query instead of
   Aspell.
+- New W3C epubcheck tool added to HTML menu. User selects the epub to be
+  checked, and results are shown in standard errorcheck dialog
 - Combining Characters now available in Compose Sequences, e.g. `+~` for 
   combining tilde above; `_~` for combining tilde below
 - Additional compose sequences added for precomposed vowels with macron and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
   English, French and German dictionaries are included with the release.
 - Word Frequency's Check Spelling button now uses Spell Query instead of
   Aspell.
-- New W3C epubcheck tool added to HTML menu. User selects the epub to be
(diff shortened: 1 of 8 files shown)


src/lib/Guiguts/Preferences.pm
```diff
@@ -785,6 +785,29 @@ sub filePathsPopup {
             -relief       => 'sunken',
             -background   => $::bkgcolor,
         )->pack( -expand => 'y', -fill => 'x' );
+        my $f7a = $::lglobal{filepathspop}->Frame->pack(
+            -side   => 'top',
+            -anchor => 'n',
+            -fill   => 'x'
+        );
+        $f7a->Label(
+            -text   => "W3C epubcheck:",
```
cpeel, line 794: Do we care that W3C uses `EPUBCheck` for the tool and not `epubcheck`? I don't really care but I want to make sure we're making an intentional decision for this capitalization. https://github.com/w3c/epubcheck

windymilla: Oops - someone in the Slack channel called it "epubcheck", then the GitHub URL had "epubcheck" and the program was called "epubcheck.jar". I ended up totally blind to the capitalization of "EPUBCheck" used in the W3C docs. I'll get that done straightaway - much better to match their name for the tool.

windymilla:
Hopefully that's better now - @srjfoo has suggested I merge after @cpeel is happy. The feature is being tested by tame testers, and works in a very similar manner to the other W3C tools, so once one of them works OK on Macs/Linux, they all should (famous last words!)

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
   English, French and German dictionaries are included with the release.
 - Word Frequency's Check Spelling button now uses Spell Query instead of
   Aspell.
+- New W3C epubcheck tool added to HTML menu. User selects the epub to be
```
cpeel, line 17: s/epubcheck/EPUBCheck/

cpeel (approved):
One minor cap difference in CHANGELOG but LGTM!

PR #1024, opened 2022-10-10
Shorten EPUBCheck error messages

### LINKED ISSUE #1023: Reduce width of EPUBCheck messages
Possible ideas:
1. Wrap at 40(?) characters (maybe try to see how wide window is) and indent wrapped lines
2. When deleting an error with right-click, also delete following indented lines?
3. Hide the query count at the top of the dialog - it's not useful for this tool
4. Replace first part of filename with `...`, e.g.
`ERROR(RSC-005): .../OEBPS/123456789_myfile-0.html.html(33,38): Error while parsing file: element "time" not allowed, etc., etc.`

windymilla:
Error messages from EPUBCheck contain the full pathname of the epub being checked, plus the internal path to the html file inside the epub. The internal html file has a long name with about 20 digits plus the name of the epub plus an index number. This can easily amount to 70+ characters before the error message even gets started.

Trim all of this down to `...-N.html` so error message can be read more easily. This contains sufficient information to find the error in the epub using Calibre editor or by unzipping it. Will be documented in the manual.

Also, hide the query count label since not needed for this check.

Fixes #1023

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -31,9 +31,12 @@ sub errorcheckpop_up {
 
     my $gcol = 0;
 
-    # Label to count number of queries
+    # Label to count number of queries (not shown for EPUBCheck)
     $::lglobal{eccountlabel} =
-      $ptopframeb->Label( -justify => 'left', )->grid( -padx => 10, -row => 0, -column => $gcol++ );
+      $errorchecktype eq 'EPUBCheck'
+      ? 0
+      : $ptopframeb->Label( -justify => 'left', )
+      ->grid( -padx => 10, -row => 0, -column => $gcol++ );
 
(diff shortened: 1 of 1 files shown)


src/lib/Guiguts/ErrorCheck.pm
```diff
@@ -500,6 +503,14 @@ sub errorcheckpop_up {
             my $freq = spellqueryfrequency($line);
             next if $freq > $::spellquerythreshold;       # If it's spelled the same way several times, it's probably not an error
             $line .= " ($freq)";
+
+            # EPUBCheck's error messages each contain the full path of the epub file, maybe plus the path
```
cpeel, line 507: I know this is here because it's before the `elsif` but the indented placement (thanks perltidy!) makes it look like it's attached to the `Spell Query` section. Consider moving this to inside the `EPUBCheck` block.

cpeel (approved):

PR #1026, opened 2022-10-13
Generate non-serialized HTML5 by default

windymilla:
Use a global flag to control whether void elements such as `<br>` are closed, i.e. `<br />`.
Default is not to close, as preferred by PG.
Global flag is saved in `setting.rc` which could be edited manually to set to true. Not available in user interface though, to discourage experimentation and confusion :)

When set to true, testsuite generates identical HTML to previous version. When set to false, only differences are void element closures.


src/guiguts.pl
```diff
@@ -150,6 +150,7 @@
 our @gsopt;
 our $highlightcolor         = '#a08dfc';
 our $history_size           = 20;
+our $html5serialization     = 0;                           # Don't serialize by default, i.e. don't close void elements like <br>
```
cpeel, line 153: I know this is terminology we got from somewhere else, but "HTML5 serialization" isn't a boolean (yes = XML, no = HTML). HTML5 documents are *always* serialized. There are 2 different ways to serialize HTML5, either HTML or XML. I would encourage us to be explicit both here and in the changelog about what we actually mean. In the code this could be as simple as making `$html5serialization` a string being either `html` or `xml`.

windymilla: What about renaming the variable xmlserialization, plus adjustment to changelog, or don't you like that?

cpeel: That's fine too!

windymilla: Thanks - done!

cpeel (approved):
Thank you for humoring my pedanticness!

windymilla:
> Thank you for humoring my pedanticness[**pedantry 😂]
I'd prefer it to be accurate too, thank you

srjfoo (approved):

PR #1031, opened 2022-10-16
Refresh SQ correctly on adding  `good_words`

### LINKED ISSUE #1030: Adding good_words file doesn't cause the SQ cache to be cleared
**Describe the bug**
If Spell Query is run, then the good_words.txt file is loaded using the button in the dialog, re-running the check doesn't take into account the new words added, presumably because the SQ cache isn't being cleared.

windymilla:
1. When `good_words.txt` was added to the project dictionary, the cached dictionaries weren't being cleared.
Overlooked when #1015 introduced caching.

2. When `good_words.txt` were added, user immediately had to re-run SQ to remove the good words from the list of queries. Now re-run SQ automatically.

Fixes #1030

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -103,9 +103,13 @@ sub errorcheckpop_up {
         $::lglobal{spellqueryballoon} = $top->Balloon() unless $::lglobal{spellqueryballoon};
         $ptopframeb->Button(
             -activebackground => $::activecolor,
-            -command          => \&::spelladdgoodwords,
-            -text             => 'Add good_words.txt',
-            -width            => 16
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes:

1. Run spell query and note one of the queried spellings, e.g. "xyz"
2. Create a file `good_words.txt` containing one word per line, one of which should be "xyz".
3. In the SQ dialog, click "Add good_words.txt"

In the main release, this has no apparent effect, even if you re-run Spell Query. You have to either quit and restart GG, or change the language and change it back to force the cached dictionaries to be cleared.
In this branch, when you add the good words, Spell Query should automatically re-run and no longer be querying "xyz".

cpeel (approved):

srjfoo (approved):

PR #1034, opened 2022-10-21
Improve display of long EPUBCheck errors

windymilla:
1. Wrap EPUBCheck errors which can be 200+ characters long
2. Provide a "Copy errors" button that copies all the errors to the copy/paste buffer, allowing user to edit them in a different editor, select bits to include in a forum question, etc.


windymilla:
Testing notes: Wrap width depends on width of dialog. If you resize the dialog and re-run EPUBCheck, it should wrap to the new size.

cpeel (approved):

srjfoo (approved):

PR #1036, opened 2022-10-23
Retrieve Nu validator by tag not release

windymilla:
Nu validator is not being released regularly, just tagged. Git clone tagged version and build `vnu.jar`

Note for Windows development: requires Mingw64 shell since Mingw32 shell cannot deal with a file created during the build process with a Unicode character in the filename. When attempting to tidy after the build, the `rm -rf` fails claiming the `docs` dir is not empty, because `rm` can't delete a file in the dir.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -17,9 +17,10 @@ Github.
 4. Commit the version update changes and tag the release
    (ie: `git tag r<version>`) and push it up to github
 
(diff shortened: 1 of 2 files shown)


windymilla:
Thanks - do you think that's OK?

cpeel (approved):
Perfect, thanks!

PR #1043, opened 2022-10-28
Avoid uninitialized variable error calling ebookmaker

windymilla:
Unlikely to happen in the wild, unless HTML file has no `<title>` field.  Caused by wrong type of "and" in #571 18 months ago.

Interesting to speculate whether specific valid locations of the `<title>` and `</title>` fields could have bitwise-anded together to give "false" and therefore not pick up the author name. But not interesting enough that I want to try to make it happen.


cpeel (approved):
Ohh... good catch, these are always tricky.

PR #1046, opened 2022-11-04
Warn user if tool fails to write error file

windymilla:
Most tools never write an empty error file - there are always some lines of output even when there are no errors.

Norton AV (and maybe others) sometimes block a tool from writing to the error file. Guiguts was giving an "All OK" message when this happened. So, check for empty output file for most tools and report to user, with hint about possible cause.

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -343,6 +343,25 @@ sub errorcheckpop_up {
         }
     }
 
+    # Zero-size error file means most tools failed badly (e.g. blocked by anti-virus software)
+    if (    -z $errname
+        and $errorchecktype ne "Spell Query"
+        and $errorchecktype ne "Load Checkfile"
+        and $errorchecktype ne "Nu HTML Check"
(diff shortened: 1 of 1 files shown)


windymilla:
@srjfoo, you probably can't test the actual bug, but please could you test that running some of the checking tools, like jeebies, validators, etc. still give errors if there's an error, or no error if there aren't any - they should never display the new error box. A Windows user with Norton AV has confirmed that the error box does get popped if Norton blocks file-writing.

cpeel (approved):

srjfoo:
> @srjfoo, you probably can't test the actual bug, but please could you test that running some of the checking tools, like jeebies, validators, etc. still give errors if there's an error, or no error if there aren't any - they should never display the new error box. A Windows user with Norton AV has confirmed that the error box does get popped if Norton blocks file-writing.

I tested jeebies and the HTML and CSS validators. All of them display some kind of text in them, whether there's an error or not (I deliberately created a short text file with no he/be instances in it). Is there anything else that you want me to test?

srjfoo (approved):

PR #1047, opened 2022-11-04
Improve "word" selection

windymilla:
As part of [#941](https://github.com/DistributedProofreaders/guiguts/issues/941) which made navigation, selecting and deleting
by word chunks more consistent, double-clicking a word selected from the previous word start to the next one (similar
 to MS Word).

This was unpopular with users, so wind this back to only select the word itself, not any trailing spaces.

Also, after word (or line) select using double/triple click, the anchor for the selection was left at the click point (ie mid-word) which led to weirdness when subsequently extending the selection using Shift+arrow keys.

Now force the anchor to the start and the cursor to the end and it behaves more logically.

### DIFF
--- a/src/lib/Guiguts/TextUnicode.pm
+++ b/src/lib/Guiguts/TextUnicode.pm
@@ -390,11 +390,11 @@ sub SelectTo {
         }
     } elsif ( $mode eq 'word' ) {
         if ( $w->compare( $cur, '<', 'anchor' ) ) {
-            $first = $w->index( $w->safeword("$cur + 1c wordstart") );
-            $last  = $w->index( $w->safeword('anchor - 1c wordend') );
+            $first = $w->index( $w->safeword( "$cur + 1c wordstart", 'strictword' ) );
+            $last  = $w->index( $w->safeword( 'anchor - 1c wordend', 'strictword' ) );
         } else {
-            $first = $w->index( $w->safeword('anchor + 1c wordstart') );
-            $last  = $w->index( $w->safeword("$cur wordend") );
+            $first = $w->index( $w->safeword( 'anchor + 1c wordstart', 'strictword' ) );
+            $last  = $w->index( $w->safeword( "$cur wordend",          'strictword' ) );
         }
     } elsif ( $mode eq 'line' ) {
         if ( $w->compare( $cur, '<', 'anchor' ) ) {
@@ -521,15 +521,17 @@ sub KeySelect {
 #
 # Note this version interprets "wordend" as the beginning of the next word,
 # not the end of the current word since this is more common in modern editors
+# This can be overridden by passing a true value for the optional "strictword" argument
 #
 # A sequence of word characters counts as a "word", as does a sequence of
 # non-word non-space characters. Apostrophes count as word characters
 #
 # Start & end of line always count as word breaks
 sub safeword {
-    my $w   = shift;
-    my $pos = shift;
-    my $new = '';
+    my $w          = shift;
+    my $pos        = shift;
+    my $strictword = shift;
+    my $new        = '';
 
     # Word characters include straight & curly apostrophes, although this means
     # that close single quotes also count as word characters.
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
Testing on M1 Mac: Double-click selects just the word without spaces. Shift-double-click, either fore or aft, extends the selection to just the word boundary.

Shift-arrow after the select behaves as I would expect; right-arrow extends the selection a character at a time, left-arrow deselects a character at a time until it gets to the beginning of the selection, and then selects a character at a time in the other direction.

Shift-control-right-arrow selects through the space following the word added to the selection -- it that the expected behavior?

Just a note -- I don't know what the behavior is on Windows or Linux, but triple-click does not select anything more than the double-click does. Should it?

windymilla:
@srjfoo The behavior you describe for double-clicking and extending via arrow keys is exactly as expected, thanks.
On Windows, triple click selects a line. NP++ does the same. If NP++ does it on Mac, but GG doesn't, I could look into it, but separately from this PR (if anyone thinks it's worth it - I never use triple click).

srjfoo:
> On Windows, triple click selects a line. NP++ does the same. If NP++ does it on Mac, but GG doesn't, I could look into it, but separately from this PR (if anyone thinks it's worth it - I never use triple click).

As far as I known NP++ is not available on Macs, so I have no way of testing that 😁 . However, triple-click in

* BBEdit, TextEdit and FF selects a whole paragraph
* LibreOffice, selects a sentence within a paragraph

Definitely doesn't need to be resolved in this PR. I mentioned it because you mentioned triple-click as though it ought to be doing something special, and I couldn't see anything special on a Mac. If it's going to do something, it should probably be consistent with what it does on Windows, regardless of what it does in other text-manipulation programs on Macs? Can we check with someone who used GG on Linux to see what it does there?

windymilla:
@srjfoo - I've looked in the Perl/Tk code and in the GG code. There is nothing obvious that would stop it working on a Mac - if the triple-click event reaches the Perl/Tk code, I think it would end up selecting a line of text. Is it possibly related to double/triple click speed as described here: https://discussions.apple.com/thread/6736381

srjfoo:
> 



> @srjfoo - I've looked in the Perl/Tk code and in the GG code. There is nothing obvious that would stop it working on a Mac - if the triple-click event reaches the Perl/Tk code, I think it would end up selecting a line of text. Is it possibly related to double/triple click speed as described here: https://discussions.apple.com/thread/6736381

@windymilla -- not that I can tell. Fwiw, the last post in that thread is nearly 5 years old. 

I use the track-pad almost exclusively (except when I need to pull out a mouse to test something like this 😁 ). The system preferences don't have an option for changing the double-click speed for trackpads. I tried changing the double-click speed for the mouse, and it made no difference.

As I mentioned earlier, LibreOffice, BBEdit, TextEdit and FF all have a working triple-click. I can add Pages, Chrome and Opera to that now. Except for LO, where the triple-click selects just a sentence, triple-click selects a paragraph, which is probably equivalent to a line in GG, when working with a DP CTF, as they are both delimited by LF or CRLF.

Hm ... might that be the issue? I.e. is the triple-click looking for CRLF, and only finding LF on a Mac?

windymilla:
I don't think it could be a line ending thing - the code does use linestart and lineend, but that's quite a low level inside Perl/Tk. Those are words used in indexes into the text, e.g. line 7, column 3 is "7.3" or the end of that line is "7.3 lineend" - if that didn't work, I think lots of other stuff would go wrong. At some point we could investigate further, probably by making edits to a version of the code on your system. I'm not going to worry for now though - no-one has noticed so far :)

srjfoo:
> I'm not going to worry for now though - no-one has noticed so far :)

Works for me. I'm game to help try to troubleshoot the issue later, but I'm not fussed, either way. There are multiple ways to select a line at a time if that's what one needs to do, and triple-clicking isn't necessarily the best way to do it.

I mentioned it only because the way it works on my Macs is inconsistent with your description (maybe should be documented?), not because I felt like it needed to be fixed immediately.

If you and @cpeel think I've done enough testing, I think we're good to go. If there's something else you want me to test, let me know.

windymilla:
I'm happy with it if @cpeel is too - thanks

cpeel (approved):
Sorry for the delay. Needed to carve out some time to review this one as it's more complex than the other recent ones.

PR #1051, opened 2022-11-06
Make `.gz` default extension for TIA OCR import

windymilla:
Having `.gz` filetypes showing by default makes it easier for the user to find the file they want. They can always change to All Files if they have a different extension for some reason.


cpeel (approved):

srjfoo (approved):

srjfoo:
@windymilla -- many thanks, again, for adding the ability for CPers to be able to use TIA's Abbyy OCR files.  🎉

PR #1052, opened 2022-11-07
Use `figure`/`figcaption` instead of `div` for illos

### LINKED ISSUE #1045: HTML Auto Illus. Search: Add option to generate HTML5 <figure> and <figcaption> tags.
Even better (perhaps a separate request), I would like to customize the replacement code. With 500 illos, it is not practical to edit each one in the menu process. I'll probably see if I can modify the code, but it would be better to have a user friendly option.

windymilla:
Now GG is producing HTML5, take advantage of the new `figure` and `figcaption` elements. One advantage should be that the caption should not get separated from the illo by a page break in epub format.

Fixes #1045

### DIFF
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -1840,7 +1840,7 @@ sub htmlimageok {
     my $alt = $::lglobal{alttext}->get;
     $alt =~ s/"/&quot;/g;
     $alt       = " alt=\"$alt\"";
-    $selection = "  <div class=\"caption\">$selection</div>\n"
+    $selection = "  <figcaption class=\"caption\">$selection</figcaption>\n"
       if $selection;
     $::lglobal{preservep} = '' unless $selection;
     my $title = $::lglobal{titltext}->get || '';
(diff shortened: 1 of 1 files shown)


rtonsing:
For `<figcaption>`, I think we can leave out the `class=\"caption\"`, since we can just override the tag in CSS. Not a deal breaker, though.

rtonsing (changes requested):
For `<figcaption>`, I think we can leave out the `class=\"caption\"`, since we can just override the tag in CSS. Not a deal breaker, though.

windymilla:
> For `<figcaption>`, I think we can leave out the `class=\"caption\"`, since we can just override the tag in CSS. Not a deal breaker, though.

I wondered about that. I thought it was perhaps more friendly to the PPer to have a class for them to modify, especially if there are different caption styles within a book and they might then have two different classes, e.g. caption1 (italic, right justified) and caption2 (bold, centered). I don't mind removing it if people feel that's best though.

rtonsing (approved):

cpeel (approved):

windymilla:
Sorry, @srjfoo, I didn't leave any testing notes - there's no rush for testing it, but for when you get onto it...

If you have the text `[Illustration: Some caption]` in a file, then use HTML->HTML Generator, then Auto Illus Search, it will find the next [Illustration] and pop a dialog for you to choose a jpeg. It will pre-fill the caption field if you have a caption in the markup already, or you can type into it. When you click OK it inserts HTML markup - with this PR it should use figure and figcaption instead of divs. You can also "Insert & Load Next", which means to add the markup for this illo and jump to the next illo in the file. It automatically chooses the next jpeg file alphabetically, which you can adjust with the Prev file/Next file buttons.

An alternative if you don't have [Illustration] is to use HTML->HTML Markup, then choose Image from the Links section. The same dialog as above pops up and you can just insert the markup at the current location.

srjfoo (approved):

PR #1059, opened 2022-12-26
Sq lang names

windymilla:
Previously any non-word character in the list of languages for a
project was used as a separator. However, some Aspell dictionaries
have hyphens, e.g. "de-1901" or underscores.
Therefore restrict the characters used as separators to spaces, commas
or plus signs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 # Changelog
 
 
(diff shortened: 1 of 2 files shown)


cpeel (approved):

srjfoo (approved):

windymilla:
That was fast!

PR #1064, opened 2022-12-30
Add optional ruler/column numbers

### LINKED ISSUE #1057: Ruler like Microsoft Word
Could we have a ruler on the bottom of the window showing a ruler like Microsoft Word?  I was thinking it could display column numbers.  Maybe make it clickable and show a line on window.  Would be helpful when working on wide tables in text mode.

Or maybe add a highlighted area to the right of the Max. Right Margin setting making it easy to see how close you are to the right margin.

windymilla:
Similar to the row numbers which can be displayed on the left of the text window, this implements a column numbers "ruler" widget across the top.

Similar configuration to the row numbers widget is available, but not exposed to the user.

Turn on/off via Preferences menu or by Shift-right-clicking the row/column display in the status bar.

Ruler is one-based rather than zero-based (as used internally), and existing Highlight Alignment Column has been adjusted to match the highlighted column number in the ruler.

Internal LineNumberText code has been simplified a little, removing historical attempt to support plain Text widget rather than the GG TextUnicode widget. Some comments also added.

Fixes #1057

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@
   e.g. '62, '60s, 1860s, 12s, 6d
 - The pphtml tool no longer reports ebookmaker control classes, e.g.
   `x-ebookmaker-drop` as undefined or unused
+- A ruler showing column numbers can be displayed at top of text window to
+  aid with text alignment; it indicates the current column. Ruler can be
+  shown/hidden using the Preferences menu, or using Shift+Right-click on
+  the Row/Col display in the Status Bar. Highlight Alignment Column now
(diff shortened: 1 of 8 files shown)


srjfoo (approved):

cpeel (approved):
In all honestly only lightly skimmed the code and did not do a thorough code review. Since Sharon approved it I know it works which is the important part.

If you would like a more in-depth review I'm happy to do that, just let me know.

windymilla:
Thanks @cpeel - I'm happy that you've cast your eyes across it.

PR #1065, opened 2022-12-30
Fix false positive in Word Frequency Emdash check

### LINKED ISSUE #1063: Guiguts 1.5.1 (newest update) Word Frequency - All Words
Always I go into 'Word Frequency'. I check 'All Words' looking for characters not suppose to be in a word. I'll check 'Emdashes' normally quite boring. In the case yesterday, I had a 'suspect word (in 'Emdashes'). 'Dining-room' (with a hyphen). This word is really correct. If I go to my .txt file, this is what I see: <sc>Act I.</sc>--Dining-room. (I am still in the 'pre-split' stage.) 'Dining-room' is preceded by a -- (hyphen hyphen) which is the problem.

From what I understand this is a bug (somewhere in GG, the algorithm(?)). A person check this out in Guiguts 1.0.25. This bug was present.

Thank you.

windymilla:
Code was attempting to build a list of "word--word", but was inadvertently including "--word" and "word--" which then led to false positives in the "suspects" check.

Fixes #1063

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
   be separated by spaces, commas, or plus signs, e.g. `en, fr`
 - `Apply Poetry Markup to Selection` could delete some characters when the 
   poem text had not been indented
+- Word Frequency could report false Emdash suspects if there was an emdash
+  before a hyphenated word.
 
 
 ## Version 1.5.1
--- a/src/lib/Guiguts/WordFrequency.pm
+++ b/src/lib/Guiguts/WordFrequency.pm
@@ -33,7 +33,7 @@ sub wordfrequencybuildwordlist {
 
         # build a list of "word--word""
         for my $word (@words) {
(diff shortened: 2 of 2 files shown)


windymilla:
Testing notes: WF, Suspects only, Emdashes is supposed to report if both "word1--word2" and "word1-word2" occur in a text, i.e. same two words separated by hyphen in one case and double hyphen (emdash) in another.
Since it was adding examples of "--word3" or "word4--" to its "double hyphen" list, it got confused later and reported them as suspects.
To test, put some examples of "abc--def" with and without examples of "abc-def" (should be reported) as well as "--xyz" and "opq--" (which should not be reported).

windymilla:
Actual case reported had `</sc>` before the emdash. WF replaces tags with a space so it triggered the above situation.

srjfoo (approved):

cpeel (approved):

PR #1069, opened 2023-01-13
Implement new PG HTML `<title>` field format

### LINKED ISSUE #1067: Adjust `<title>` field to match PG's preference
PG want to have the book title, without the author, and with ` | Project Gutenberg` at the end, e.g. 
`<title>Australia—Fortune land | Project Gutenberg</title>`

They are not concerned whether it's title- or sentence-case in the `<title>` field

windymilla:
New format is `Book title | Project Gutenberg`
Author's name is not included.

1. Change default header file to new format
2. Store author in global variable, saved in bin file
3. Continue to show author in Autogen dialog
4. Use author variable when creating epub via ebookmaker
5. Update pphtml's check of `<title>` field
6. Update changelog

Fixes #1067

### DIFF
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -1342,7 +1342,7 @@ sub html_convert_pageanchors {
 }
 
 sub html_parse_header {
-    my ( $textwindow, $headertext, $title, $author ) = @_;
+    my ( $textwindow, $headertext, $title ) = @_;
     my $selection;
     my $step;
     my $closure = voidclosure();
@@ -1397,9 +1397,7 @@ sub html_parse_header {
         close $infile;
     }
 
-    $author     =~ s/&/&amp;/g       if $author;
-    $headertext =~ s/TITLE/$title/   if $title;
-    $headertext =~ s/AUTHOR/$author/ if $author;
+    $headertext =~ s/TITLE/$title/ if $title;
     my $mainlang = ::main_lang();
     $headertext =~ s/BOOKLANG/$mainlang/g;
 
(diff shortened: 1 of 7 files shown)


windymilla:
Testing notes:
1. From a finished text file (without PG header/footer) using HTML | HTML Generator should show the title/author in the dialog if the code is able to extract them - the extraction hasn't changed
2. User can now edit the author if it's wrong and the value should be saved in the bin file.
3. Running ebookmaker should use the author field from the generator dialog/bin file.

src/lib/Guiguts/HTMLConvert.pm
```diff
@@ -1397,9 +1397,7 @@ sub html_parse_header {
         close $infile;
     }
 
-    $author     =~ s/&/&amp;/g       if $author;
```
cpeel, line 1400: Do we want to leave these others in-place for people with older updated templates? I'm not sure if it's better or worse to have older templates succeed or fail.

windymilla: I think it's better if theirs says 
`Alice in Wonderland, by AUTHOR` 
because it should alert them to the fact something hasn't worked. 
The PPer had to check it manually anyway, because the code didn't always guess at the right bit of text to use as the title and author.

windymilla:
Just PPVing a book for someone, and it created an ePub with the dummy author name I was using when testing this. Realised I'd forgotten to clear the author variable - now done.

src/lib/Guiguts/HTMLConvert.pm
cpeel: Sounds good!

cpeel (approved):

srjfoo (approved):
HTML generation dialog worked okay, but bear in mind that I still haven't gotten ebookmaker installed yet.

PR #1075, opened 2023-01-21
Fix undefined bookauthor variable

windymilla:
Introduced in #1069, new global `bookauthor` variable could end up undefined if code couldn't find an author in the text. It never used to matter because the author wasn't stored in the bin file.

### DIFF
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -1435,7 +1435,8 @@ sub html_parse_header {
 sub get_title_author {
     my $textwindow = $::textwindow;
     my $step       = 0;
-    my ( $selection, $title, $author );
+    my ( $selection, $title );
+    my $author        = '';
     my $completetitle = '';
     my $intitle       = 0;
     while (1) {    # find title


windymilla:
Testing notes: Testing notes: given a text file beginning something like that below (including blank lines at start), do an HTML Autogenerate
```




Book Title

Any text for 
a few
lines
but mustn't include
the word "b-y" (without the hyphen)
so that effectively there 
is no 
author
```

srjfoo (approved):
Is there anything that needs to be tested other than that an "author" entry isn't added to the `<title>` element when there's not one? I thought there was something about the author needing to be left out, regardless (or is that up-coming)?

cpeel (approved):

windymilla:
> Is there anything that needs to be tested other than that an "author" entry isn't added to the `<title>` element when there's not one? I thought there was something about the author needing to be left out, regardless (or is that up-coming)?

Yes, #1069 already did that (leaving out the author regardless). The header doesn't have the word "AUTHOR" (which would have bee substituted with the author if there was one) and the code (about line 1400 in HTMLConvert.pm) only substitutes the TITLE and BOOKLANG since #1069.

So, there's no need to test that in this PR, thanks

PR #1079, opened 2023-01-24
Allow user to prioritize display of footnote error types

### LINKED ISSUE #1072: Selectively prioritize Footnote Fixup color categories in "Footnote Check"
When there is more than one [possible] problem with a footnote, only one of the color codes (yellow, magenta, cyan, brown) for it is shown in the "Footnote Check" list.  For example, it seems that brown ("very long") takes priority over all of the others, so if a long footnote also has no anchor, it will be highlighted in brown but not in magenta. This makes it difficult to find the lower-priority problems, as they're not explicitly flagged.

One solution would be to be able to click on the color boxes at the top (or on new buttons associated with them) to give that category "on top" priority. Other solutions would be to hide/show the selected category in such a way that, when hidden, any lower-priority issue's color would become visible, or to only show that category's color.

Sample file available on request.

windymilla:
If a footnote has two errors, e.g. it's a duplicate and a long footnote, then only one of the errors can be shown by coloring the footnote in the Footnote Check dialog list.

This commit allows the user to click a colored label at the top to make that error type the highest priority, so all footnotes with that error will be highlighted in the appropriate color.

Also a bit of code tidying to reduce duplication and put the link between error types and colors in just one place.

Fixes #1072

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@
   match the new ruler.
 - New HTML title format: book title followed by ` | Project Gutenberg`
 - Ebookmaker version 0.12.24 included - minimum chunk size = 256KB
+- Colored labels at top of Footnote Check dialog can be clicked to make
+  that error type high priority in the list displayed below, to help with
+  the situation where footnote has more than one error
 
 ### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes: [genbugtest.txt](https://github.com/DistributedProofreaders/guiguts/files/10492392/genbugtest.txt) contains quite a few footnotes with errors. Use Tools, Footnote Fixup, then First Pass button (with or without ) to bring up Footnote Check dialog. Some footnotes have two error types (i.e. two colors) but only the latest applied one shows. Click the colored labels at the top to make that type the highest priority. One near the bottom is cyan and brown, for example.

cpeel (approved):

srjfoo (approved):

PR #1082, opened 2023-01-27
Allow SQ results to be sorted alphabetically

### LINKED ISSUE #1071: Spell Query enhancements
Since we're currently discussing Spell Query, these new options sometimes would be helpful (at least to me):

1. display in alphabetical sequence instead of in position sequence
2. display only lower-case words
3. display only upper-case words

windymilla:
When checking spellings, it can be useful to have the queries sorted alphabetically (case sensitive or insensitive) or sorted by location in the file (default)

When the sort order is changed, the list is refreshed without re-running SQ (or previously skipped spellings would re-appear)

The sort order is stored for this run of GG only, not permanently.

Fixes #1071

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Colored labels at top of Footnote Check dialog can be clicked to make
   that error type high priority in the list displayed below, to help with
   the situation where footnote has more than one error
+- Spell Query results can now be sorted alphabetically or by line number
 
 ### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes: Use Spell Query on any file. Choose sort order from the buttons at the top of the results dialog. Running SQ again should retain the same sort order. Restarting GG will revert to default order (line number).
Things to test: the same bad spelling repeated with upper/lowercase variation, on different lines and same line of file to check alphabetical and row:col sorting.

Other uses of the same dialog (e.g. Bookloupe, HTML validation, etc) should not be affected by this.

cpeel (approved):

srjfoo (approved):

PR #1089, opened 2023-02-01
Add block search options

windymilla:
Two features to improve ability to search for blocks and then do S&R operations within the block

1. Previously removed by #928 because only some block types were implemented, and it took a lot of space in the Search menu. No-one seemed to notice or care for a long time. Request for its return has been implemented - as a submenu so as not to clog Search menu - also improved to avoid code duplication, to make it cope better when it doesn't find any more blocks, and to cope better when switching from searching forwards to backwards and vice versa. All block types now implemented, ordered so that if menu is torn off, user can shrink it to just show the most important ones.

2. There is a general "stay on top" flag already, and a specific flag for the WF dialog. This implements a similar specific flag for the S/R dialog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@
   that error type high priority in the list displayed below, to help with
   the situation where footnote has more than one error
 - Spell Query results can now be sorted alphabetically or by line number
+- Find Block Markup submenu added to Search menu
+- Search/Replace dialog can be kept on top, like Word Frequency, in
+  Preferences->Appearance menu
 
 ### Bug Fixes
 
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -207,6 +207,7 @@ our $verboseerrorchecks    = 0;
 our $viscolnm              = 0;
 our $vislnnm               = 0;
 our $wfstayontop           = 0;
+our $srstayontop           = 0;
 
 # These are set to the default values in initialize()
 our $gutcommand         = '';
--- a/src/lib/Guiguts/FileMenu.pm
+++ b/src/lib/Guiguts/FileMenu.pm
@@ -904,7 +904,7 @@ EOM
             intelligentWF ignoreversions italic_char jeebiesmode lastversioncheck lastversionrun lmargin markupthreshold
             multisearchsize multiterm nobell nohighlights pagesepauto projectfileslocation notoolbar poetrylmargin projectfileslocation
             recentfile_size rmargin rmargindiff rwhyphenspace sc_char scannos_highlighted
-            searchstickyoptions spellcheckwithenchant spellquerythreshold stayontop toolside
+            searchstickyoptions spellcheckwithenchant spellquerythreshold srstayontop stayontop toolside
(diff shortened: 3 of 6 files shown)


cpeel (approved):

srjfoo:
A couple of comments.

* I really like the "find block" tearoff menu (I just wish that the find could be done without forcing the cursor to the document window; it's awfully annoying to have to click once to make the tearoff active before one can click on the find next/previous.
*  I especially like the generic "Find Next/Previous Block". This may be only something that would be useful in testing, so feel free to shoot the idea down, but would it be useful to have a "Find First Block" and "Find Last Block"?

> All block types now implemented, ordered so that if menu is torn off, user can shrink it to just show the most important ones.

Just curious about this statement. When I shrink the tear-off menu, it has no scroll-bar. `/* */` and `/# #/` are the most important, true, but without a scrollbar, or the ability to scroll even though there is no scrollbar, the PPer can't shorten the window for most of the blocks, and then scroll down if there are just one or two of the less-used blocks.

\<tb\>

Okay, having tested the "always-on-top" option, for S/R, I see that applies to my second bullet point above. Is the assumption that when a PPer finds the next block, they'll want to do something to it, so the document window should be on top?

srjfoo (approved):
Works as advertised -- my separate comments are purely questions, and not problems.

windymilla:
> A couple of comments.
> 
> * I really like the "find block" tearoff menu (I just wish that the find could be done without forcing the cursor to the document window; it's awfully annoying to have to click once to make the tearoff active before one can click on the find next/previous.

I think we discovered on a previous PR that this is a general Mac issue - if you click on a window that doesn't have focus, then your first click focuses, and you have to click again to perform the action. This is different from Windows behavior. It wouldn't be acceptable to keep the focus on the menu, since then if you find a block and want to make an edit there, you would have to click on the main window to get focus, then click where you wanted to make the edit - so 2 clicks again. In general, the S&R features put the focus in the main window at the location searched for.

> * I especially like the generic "Find Next/Previous Block". This may be only something that would be useful in testing, so feel free to shoot the idea down, but would it be useful to have a "Find First Block" and "Find Last Block"?

It wouldn't be used much in PPing, I don't think (though I can see it in testing) and it's the equivalent of "go to start" (Windows shortcut Ctrl+Home) then Find Next Block, and similar for the Find Last Block.


> > All block types now implemented, ordered so that if menu is torn off, user can shrink it to just show the most important ones.
> 
> Just curious about this statement. When I shrink the tear-off menu, it has no scroll-bar. `/* */` and `/# #/` are the most important, true, but without a scrollbar, or the ability to scroll even though there is no scrollbar, the PPer can't shorten the window for most of the blocks, and then scroll down if there are just one or two of the less-used blocks.

For many projects (and all at DPC) those most common are the only ones needed, so there would be no need to access the lower ones. If the user wants to access them, then would need to stretch the menu bigger again. I just thought it was worth mentioning, since it would save screen space, and I didn't even know until recently that you could shrink the tear-off.


> Okay, having tested the "always-on-top" option, for S/R, I see that applies to my second bullet point above. Is the assumption that when a PPer finds the next block, they'll want to do something to it, so the document window should be on top?

Yes :)

PR #1092, opened 2023-02-02
Fix SQ sorting bug

windymilla:
If Spell Query errors were right clicked (to skip/remove from list) then using the new sorting options made them reappear. Also you could then get error messages if you clicked on one of those errors.

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -644,7 +644,7 @@ sub errorcheckpop_up {
 sub errsortfunc {
 
     # Keep first line first, and last line last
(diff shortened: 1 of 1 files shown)


cpeel (approved):

windymilla:
Bug spotted by @srjfoo - "Beginning check" and "Check is complete" messages have been lost for SQ as a result of this commit. Will upload another commit soon

windymilla:
Commit should fix the situation for SQ, and warn future developers of the limitations of the sorting/refreshing code

srjfoo (approved):

cpeel (approved):
LGTM!

PR #1096, opened 2023-02-11
Implement Quick Search feature

### LINKED ISSUE #1088: Additional Quick Search feature
It could be helpful to have an alternative Search string live, e.g. when searching for the start of a block (or HTML element) and then wanting to do some S&R work in that block/element, then move on to the next one.

Options:
1. Text field in the status bar, with a ">" (better symbol, like magnifying glass) button (to the left of the field) to click for searching (or maybe "Enter" when focus is in the text field)
2. Mini popup dialog that would always stay on top (user could drag it over the bottom right of the status bar) containing a search field and minimal other buttons
3. Option in the existing S&R dialog (which now has stay-on-top functionality) to display an additional Search box - consider if it needs its own regex/case/word/start-at-beginning check boxes)

Features to consider:
1. Shift-click the search button for reverse
2. To reduce screen space used, right-click search button (or text field? clash with "paste") for options like regex/case/etc
3. Probably no search history (but maybe?)
4. If in status bar, might still want to "float" it to drag to somewhere on screen, then re-dock it
5. Crazy idea - if you float/tear-off from the one in the status bar, you get a new instance, so theoretically you could do that twice and have two more search boxes. 
6. Same idea, maybe slightly less crazy would be to do this in the "Option in existing S&R dialog" method - like add/remove "Multi" replacements, just have add/remove "Multi" search boxes.
7. Keyboard shortcuts for goto/pop quick search and repeat quick search

windymilla:
1. Additional mini search dialog that works independently of the main S/R dialog
2. Accessible through Search menu and Shift+Ctrl+f shortcut

Fixes #1088

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Next version
+
+- `Quick Search` in Search menu (or Shift-Ctrl-f) pops a mini-dialog with
+  basic search options and a history button. Clicking or Shift-clicking Search
+  button, or typing Return/Shift-Return, searches forwards or backwards.
+
 
 ## Version 1.5.2
 
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -235,6 +235,7 @@ our @recentfile;
 @recentfile = ('README.md');
 our @replace_history;
 our @search_history;
+our @quicksearch_history;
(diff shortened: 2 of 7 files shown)


cpeel (approved):
Code LGTM but I defer to @srjfoo for verifying the functionality.

srjfoo (approved):

PR #1097, opened 2023-02-12
Show matching tags, and detect unclosed tags

windymilla:
1. User can select quote/bracket/start of tag, or place cursor
adjacent (code prefers before) to or inside tag
2. Menu option or Control+square-bracket finds matching one,
or warns if not found

Also

3. New check, like Tidy, Jeebies, etc., accessible in the HTML menu
4. Lists any HTML opening tags that don't have a matching close
and vice versa
5. Ignores self-closing (void) tags.

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -16,7 +16,7 @@ my $BEGMSG = "Beginning check:";
 my $ENDMSG = "Check is complete:";
 
 # General error check window
-# Handles Bookloupe, Jeebies, HTML & CSS Validate, Tidy, Link Check
+# Handles Bookloupe, Jeebies, HTML & CSS Validate, Tidy, Unclosed Tag Check, Link Check
 # pphtml, pptxt, ppvimage, Spell Query, EPUBCheck and Load External Checkfile.
 sub errorcheckpop_up {
     my ( $textwindow, $top, $errorchecktype ) = @_;
@@ -365,7 +365,8 @@ sub errorcheckpop_up {
         and $errorchecktype ne "Spell Query"
         and $errorchecktype ne "Load Checkfile"
         and $errorchecktype ne "Nu HTML Check"
-        and $errorchecktype ne "Nu XHTML Check" ) {
+        and $errorchecktype ne "Nu XHTML Check"
+        and $errorchecktype ne "HTML Tags" ) {
         unlink $errname;
         my $dialog = $top->Dialog(
             -text    => 'Error file was empty - maybe blocked by anti-virus software?',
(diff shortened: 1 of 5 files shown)


windymilla:
Testing notes:
In HTML file, put cursor on or adjacent to HTML tag, opening or closing, and press Ctrl-[
The tag and it's pair should be highlighted - if the match is off screen, it should scroll to bring it on screen.
In text file, do the same with curly quotes or brackets (round, square or braces).
It should be able to cope with nested, e.g. 
`<div><div><span>asdasd</span></div></div>`
or
`{{([[asd]])}}`

For the report, have an HTML file, and just run the report from the HTML menu. You should be able to click on errors to jump to the source of the error. If it runs successfully, you can introduce errors, e.g. by changing `<div>` to `<notdiv>` or by deleting an open or close tag

windymilla:
OK - I think that should be not only ready for @srjfoo functionality check, but also @cpeel code check

srjfoo (approved):

src/lib/Guiguts/ErrorCheck.pm
```diff
@@ -1666,6 +1669,46 @@ sub booklouperun {
 
 }    # end of variable-enclosing block
 
+#
+# Check that all relevant opening HTML tags have a matching close tag
+# and vice versa
+sub htmltagsrun {
+    my $errname    = shift;
+    my $textwindow = $::textwindow;
+    my $TAGCH      = "[a-z0-9]";      # Permissible characters in HTML tag name
```
cpeel, line 1678: Do we care that technically it's `[a-zA-Z][a-zA-Z0-9]`? That is, all HTML tags start with a letter and that they can be both upper and lowercase?

src/lib/Guiguts/Highlight.pm
```diff
@@ -8,9 +8,12 @@ BEGIN {
     @ISA = qw(Exporter);
     @EXPORT =
       qw(&scannosfile &hilite &hiliteremove &hilitesinglequotes &hilitedoublequotes &hilitepopup &highlight_scannos
-      &highlight_quotbrac &hilite_alignment_start &hilite_alignment_stop &hilite_alignment_toggle);
+      &highlight_quotbrac &hilite_alignment_start &hilite_alignment_stop &hilite_alignment_toggle
+      &hilitematch &hilitematchfind &hilitematchtag &hilitematchvoid);
 }
 
+my $TAGCH = "[a-z0-9]";    # Permissible characters in HTML tag name
```
cpeel, line 15: Ditto tag name restrictions.

cpeel (approved):

src/lib/Guiguts/ErrorCheck.pm
windymilla: Thanks - I think it's adequate for this purpose. I'm happy only supporting lowercase - I've never seen uppercase used in our books, though I know it is not illegal. We don't need the further complication of matching `<div>` with `</DIV>`. And if the tag is illegal, e.g. starting with a digit, it will fail the HTML validation.

Issue #1087, opened 2023-01-31
Remove calls to `update_indicators` from almost everywhere


windymilla:
Since `update_indicators` is called from `_updatesel` which is called every 0.2 seconds, it's not necessary for `update_indicators` to also be called explicitly every time something happens. There are about 40 calls to it currently.

PR #1101, opened 2023-02-15
Remove superfluous calls to `update_indicators`

windymilla:
Since `update_indicators` is called from `_updatesel` which is called on a 0.2 second timer, it's not necessary to call `update_indicators` via `SetGUICallbacks`, nor in various other places through the code.


windymilla:
Fixes #1087 

Testing: Check the status bar gets updated when you move around the file, edit it, load a new file, etc.

cpeel (approved):
Nice cleanup!

srjfoo (approved):

PR #1109, opened 2023-02-21
Add unmatched item error checks

### LINKED ISSUE #1100: Add "Unmatched double quote" report
Similar to the unclosed tag (element) report, produce a list of unclosed curly double quotes.

### LINKED ISSUE #1105: Add "unmatched block markup" report

### LINKED ISSUE #1106: Add "unmatched tag" report for pre-split text file
Essentially the same as Unclosed HTML tag report, but would be run to detect unclosed `b, i, f, g, sc` markup

### LINKED ISSUE #1107: Add "unmatched brackets" check

windymilla:
1. Add new menu options, including necessary reassignment of some keys used for keyboard navigation.
`Unclosed Tag Check` in HTML menu changed to `Unmatched Tag Check` since the tag is unmatched rather than unclosed. Quote check is with other curly quote options in Txt menu. Tag, Bracket and Block checks are in Tools menu.
2. Convert existing routine for checking tags to a general routine for checking matching items given regex and reference to match routine. Call it for each check type.
3.  Use existing global constant variable for possible block types
4. Updated changelog

Fixes #1100
Fixes #1105
Fixes #1106
Fixes #1107

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,13 @@
 
 - `Quick Search` in Search menu (or Shift-Ctrl-f) pops a mini-dialog with
   basic search options and a history button. Clicking or Shift-clicking Search
-  button, or typing Return/Shift-Return, searches forwards or backwards.
+  button, or typing Return/Shift-Return, searches forwards or backwards
 - `Find Match` in the Search menu (or `Ctrl-[`) finds the matching bracket,
-  quote, block markup, or HTML tag to the one currently selected or indicated
-  by the cursor.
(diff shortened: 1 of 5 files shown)


windymilla:
Testing notes: Should be fairly self-explanatory. These are error-check reports corresponding to the different types of matching item that the `Ctrl+[` keystroke finds. They can take a few seconds, particularly where there is an unmatched "open" at the start of a large file, then many matched open-close pairs through the rest of the file, since it needs to track through all of these to discover if there will be a matching "close" near the end.

srjfoo (approved):

cpeel (approved):
Nice abstraction!

PR #1111, opened 2023-02-23
Stop page markers within markup breaking rewrapping

### LINKED ISSUE #1110: One "R/" not recognized
When the attached file is rewrapped using "Rewrap All", the "R/" at the bottom of page 48 is not recognized, so that closing tag, the [Illustration tag below it, and "Miss Cavell" are moved to the right during rewrap.  The other 149 occurrences of this tag seem to be processed correctly.  This happens with this instance of /R...R/ in all versions of Guiguts starting with 1.5.0.

I tried deleting text above or below it, and copy/pasting the apparently identical three lines from the bottom of page 14 to this location at the bottom of page 48, but none of those made any difference.

[rewrapbug.zip](https://github.com/DistributedProofreaders/guiguts/files/10798863/rewrapbug.zip)

windymilla:
Temporary characters are inserted where page markers occur, so that after rewrapping, the page makers can be moved to the same location in the text that they were previously. This has been the case for many years. It was therefore necessary for many regexps used within rewrapping to allow for the temporary character appearing at the start of end of block markup lines.

Under two circumstances, these characters caused problems for rewrapping:

1. If, through manual editing of markup, e.g. when changing markup from the rounds to specific PPing markup, a page marker could end up in the middle of the markup, e.g. `/<123.png>*`. The regexps used didn't allow for this, so markup was missed, causing incorrect rewrapping.

2. Even when a page marker was within text as normal, the temporary character was being counted as part of the line length, and could cause center block or right-align block operations to be slightly misaligned. This has never been reported, since mis-centering by 1 character is unlikely to be noticed, and right-aligning to column 71 instead of 72 is also easily missed, unless two right-align blocks occur close to one another.

Fixes #1110

### DIFF
--- a/src/lib/Guiguts/SelectionMenu.pm
+++ b/src/lib/Guiguts/SelectionMenu.pm
@@ -18,8 +18,10 @@ my $TEMPPAGEMARK = "\x7f";
 sub wrapper {
     my ( $leftmargin, $firstmargin, $rightmargin, $paragraph, $rwhyphenspace ) = @_;
     return $paragraph
-      if ( $paragraph =~ m|^$TEMPPAGEMARK*[$::allblocktypes]/$TEMPPAGEMARK*\n$TEMPPAGEMARK*$|
-        || $paragraph =~ m|^$TEMPPAGEMARK*/[$::allblocktypes]$TEMPPAGEMARK*\n$TEMPPAGEMARK*$|
+      if ( $paragraph =~
+        m|^$TEMPPAGEMARK*[$::allblocktypes]$TEMPPAGEMARK*/$TEMPPAGEMARK*\n$TEMPPAGEMARK*$|
+        || $paragraph =~
+        m|^$TEMPPAGEMARK*/$TEMPPAGEMARK*[$::allblocktypes]$TEMPPAGEMARK*\n$TEMPPAGEMARK*$|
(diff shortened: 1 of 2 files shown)


windymilla:
[Test file with comments](https://github.com/DistributedProofreaders/guiguts/files/10815829/rewraperrors.txt)
Use Tools->Rewrap All
If you want to try moving page markers around to do similar tests and/or move more/less page markers into the danger zones, then right-click in status bar where it shows `Img: 048` for example, or use File->Project->Display/Adjust Page Markers.

srjfoo (approved):

src/lib/Guiguts/SelectionMenu.pm
```diff
@@ -515,6 +535,17 @@ sub selectrewrap {
     $::blockwrap = 0;    # In case of markup error, don't want to leave blockwrap variable set
 }
 
+#
+# Create a copy of a paragraph without temporary pagemarker characters for ease of
+# regex comparisons. Page markers can otherwise optionally appear anywhere in
+# "/#[4.8,24]" which means the regex to match gets very long and unreadable.
+# Note: do not use this copy for wrapping, just comparing
+sub removetpms {
```
cpeel, line 543: Great use of this new function to simplify a lot of code 🎉

cpeel (approved):

PR #1112, opened 2023-02-24
Fail more gracefully if Aspell not installed

windymilla:
Now that Aspell is not required (since PPer can use Spell Query tool instead), using Help->Software Versions should not give an error about Aspell not being installed.

Also, a few places needed to extend the check that the Aspell path variable is set, to also check the executable actually exists.

### DIFF
--- a/src/lib/Guiguts/MultiLingual.pm
+++ b/src/lib/Guiguts/MultiLingual.pm
@@ -36,8 +36,8 @@ sub spellmultiplelanguages {
     ::hidepagenums();
 
     # find Aspell and base language if necessary
-    ::spelloptions() unless $::globalspellpath;
-    return           unless $::globalspellpath;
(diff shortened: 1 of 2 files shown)


windymilla:
If you have Aspell installed, the easiest way to check is to either rename the Aspell install directory or executable, or change the Aspell directory setting under Prefs, Set File Paths. The error message would occur when using Help, Software Versions.

cpeel (approved):

srjfoo (approved):
Making sure that I understand what I'm supposed to be seeing. Since you said
>The error message would occur when using Help, Software Versions.

and that's what I got, I'm going to assume that's the extent of what you were expecting.

No error generated in the Terminal window (unlike the results using 1.5.2 as a control 😁 ). 

In addition, at least on a Mac, I get both the spell check window and a dialog to choose the location of the aspell executable if I try to run spell check in the `aspell-version` branch.

Issue #1114, opened 2023-02-26
Word Frequency mishandles combining characters


G4OEU:
**Describe the bug**
**MiXeD CasE**, **Check Accents**, and **Check Spelling** buttons in Word Frequency (only those buttons were tested) will, in one case, displace the combining character when the word containing it in the report is double-clicked and the text with that the word is displayed.  In the other case, where there is no precomposed Unicode equivalent for the combined character, the word in which the combined character is contained is truncated in the report that is produced by each of those three buttons above.

See below for examples of this.

**To Reproduce**
1. Open Guiguts with 
[WF-error-example.txt](https://github.com/DistributedProofreaders/guiguts/files/10833912/WF-error-example.txt) 2. Open the Tools>Word Frequency menu and click the MiXeD CasE button. Scroll down the report to the line with 'Shui-Hsien-Tszu' (no breve on top of the last 'u'). Double-click that line and you will see that the breve is shifted in the display (but missing in the report). This is the UN-NORMALISED text.
3. Double-click on the next line in the report which contains 'Shui-Hsien-Tszŭ' (breve on top of the last 'u'). The precomposed character that has replaced the combined character displays correctly in the second paragraph. This is the NORMALISED text.
4. Next, go back up the list to the line that says '2    en-Hou'. Double-click then double-click again. An instance of the name 'Ti̔en-Hou' is displayed in the third and fourth paragraphs. Note how the MiXeD CasE report only lists the part of the word after the combined character. In some other WF reports, the part of the word _before_ the combined character might be displayed. The truncated word in the report is the same, whether the original word (and combined character) is normalised or not.
5. The combining characters used are COMBINING BREVE (U+0306, 774 Dec) in the first two paragraphs, and COMBINING REVERSED COMMA ABOVE (U+0314, Dec 788) in the second two paragraphs.
6. It turns out there is a precomposed Unicode character for u+breve which I only later discovered. Don't know whether this is relevant to the fact that i+reversed comma above does not have a Unicode precomposed equivalent.

**Expected behavior**
(1) That Word Frequency should display combined characters as if they have been normalised when a report entry is double-clicked; 
(2)  Display words with combined characters correctly (i.e. not truncated) in the reports produced by the WF buttons.

**Environment**
* Guiguts version: From at least 1.4.0 to 1.5.3-test3. Likely present in 1.0.25 also.
* OS and version: Windows 10
* Perl version: 
* Perl distribution: 

**Screenshots**

**Additional context**

[...]

PR #1115, opened 2023-02-27
Handle words with combining chars in WF

windymilla:
Previously, combining characters were not considered part of a word, so words containing them were split into two at the location of the combining character.

By including combining characters in the permitted "word" characters, these words are now treated the same as other words.

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -596,6 +596,7 @@ sub escape_regexmetacharacters {
 
 sub deaccentsort {
     my $phrase = shift;
+    $phrase =~ s/\p{Mark}//g;                              # First remove any combining marks from phrase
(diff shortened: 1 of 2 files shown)


windymilla:
Fixes #1114 
The above issue is a good example of how WF used to fail.
Alternatively, add a combining character to a character within a word, e.g. by bringing up the Combining Diacritical Marks block from the Unicode menu. Instead of WF splitting that word into two at that point, it should consider it all one word, including the mark.
Note that Check Accents can now warn about the difference between `Ti̔en-Hou` and `Tien-Hou` (first one has accent over the `i`) if both exist in the file, rather than only for pre-composed accented characters.

cpeel (approved):

srjfoo (approved):
NTS: After you add stuff that you're using as test data REMEMBER to re-run WF before clicking on the button. "Check Accents" doesn't know about the new words if you haven't run WF after you add them. 🤦‍♀️

Issue #1113, opened 2023-02-25
"Unmatched Brackets Check" overlooks some missing brackets


charliehoward4dp:
(This is a contrived example....) In the attached file, there's a missing closing parenthesis, then, much later, a missing opening parenthesis. In between are matched parentheses.  The new "Unmatched Brackets Check" thinks the early opening parenthesis matches the later closing parenthesis and reports no errors, which is incorrect.

This happens with other brackets, as well as with "Unmatched Block Markups." (No sample provided for Blocks, but it's very easy to reproduce.)

May be an unintended consequence of supporting nesting. "No good deed...."

The existing "Check Orphaned Brackets" on the Tools Menu finds all of them.

[unmatchedbracketsbug.zip](https://github.com/DistributedProofreaders/guiguts/files/10830277/unmatchedbracketsbug.zip)

windymilla:
Are you saying that you don't want nesting to be allowed? 
Those parentheses do all have matches if you're permitting nesting.

So, are the following examples unmatched? Or more pertinently, do you want the checks to report them as errors?

`(Some (nested) brackets)`
or
```
/#
/#
Quote within quote
#/
Outer text
#/
```
or
```
<div class="poetry">
<div class="verse">
Line of poem
</div>
</div>
```
or
`<i>Some <i>italic</i> text</i>`

Or do you want different nesting rules for different types of check? Hopefully, if I know what you want, I'll be able to implement it.

charliehoward4dp:
I'm now thinking that there isn't one right answer to this, and that our best solution comes in two parts:

1. Leave the new "Unmatched" checks as they are, with their much-needed nesting support.  Keep the existing "Check Orphaned Brackets", which finds consecutive identical tags.

2. In the documentation for each of these, explain their advantages and limitations.

windymilla:
I've continued to ponder the last line of my previous comment, which is whether all the checks should have the same nesting rules. They currently do because I re-use the same code for them all. 
However,
1. I don't think `<i>Some <i>italic</i> text</i>` is valid in a pre-split file (the Tools, Unmatched Tags entry).
2. More often than not, nested brackets would indicate a potential error that the PPer should check manually. 
3. Some block markup is not permitted to be nested. Block quotes are, but you can't put `/R` inside `/R`, for example.
[...]

PR #1116, opened 2023-02-27
Adjust which match checks permit nesting

windymilla:
The Unmatched item checks now report when some items are nested. "Nested" means contained within an item of the identical type, e.g. `/*` within `/*` would be reported, but not `/R` within `/*`. Although this could lead to occasional false positives, this is better than failing to detect badly formed items.
Specifically
1. Brackets may not be nested within their own type.
2. DP tags may not be nested
3. Only `/#` block markup may be nested
4. All non-void HTML tags may be nested. It may not be valid HTML but that is not the purpose of the check. Primarily it is to detect `<p>`, `<span>`, `<div>`, etc. that are not closed/nested correctly.
5. The tag check is now split for DP and HTML tags to be treated differently.


windymilla:
Fixes #1113

cpeel (approved):

srjfoo (approved):

PR #1148, opened 2023-03-10
Remove unused rewrap parameters and code

windymilla:
There were already flags, fixes and tweaks in the rewrap code
to cope with what is wanted for Guiguts. Quite a lot of code
could never be executed. It's highly unlikely the GG use of
rewrap will be changed in the near future to take advantage
of those features, and if it is, an old version of ReflowGG can
always be dug up.

Rewrapping appears to behave identically to previously.

Second commit is because that feature is slightly less clear cut,
so keeping it separate would make it easier to restore in future
should that every be required.

### DIFF
--- a/src/lib/Guiguts/ReflowGG.pm
+++ b/src/lib/Guiguts/ReflowGG.pm
@@ -39,11 +39,9 @@ BEGIN {
 
 # This is the perl version of the C function reflow_trial
 
-sub reflow_trial($$$$$$$$$$) {
-    my (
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes: Try rewrapping files (including with various block markups) and check they behave identically to previous releases.

Theoretically, it could be a fraction quicker, but I haven't noticed any speed improvement.

cpeel (approved):
Great idea on doing this as 2-commits.

srjfoo (approved):

PR #1149, opened 2023-03-11
Make SQ, WF, S/R cope better with combining characters

### LINKED ISSUE #1118: Review use of `\p{Alnum}` to see if `\p{Mark}` should also be included
`Alnum` may be being used to find "words", in which case `Mark` should be included too to cover combining characters.

### LINKED ISSUE #1122: Just as for WF, both Spell Query and Spell Check do not consider as part of a word, composed characters that can't be normalised
**Describe the bug**
I'm seeing in both SQ and SC exactly the same problem with a composed character as I reported in Word Frequency, namely that the letter 'i' plus a 'reverse comma above' (which can't be normalised) is not recognised as part of a word. I note you may already be dealing with this sort of problem as part of issue #1118?

Bookloupe>Query word exhibits similar behaviour too. See comment below.

**To Reproduce**
1. Open Guiguts with 
[SQ-error-example.txt](https://github.com/DistributedProofreaders/guiguts/files/10861306/SQ-error-example.txt)
2. Open the Tools>Spell Check menu 
3. See that the word 'Pi̔ngs' is not treated as a whole word. As a BTW, neither is the word 'Ta̔i' but it appears that happens to be broken in to parts that are 'words' recognised by Spell Check.
4. Repeat steps 2) & 3) but instead using Tools>Spell Query

**Expected behavior**
As for your fix to Word Frequency

**Environment**
* Guiguts version: 1.5.3-test3
* OS and version: Windows 10
* Perl version: 
* Perl distribution: 

**Screenshots**

**Additional context**
The following comments are not context as such, but reflections/questions on and about the behaviour of Spell Query (SQ).
1. I'm surprised that Spell Check (SC) can process text that has not been saved to a file but that Spell Query requires the text to be saved to a file first.
2. I may have encountered a GG error when I first ran SQ without first saving the text to a file. I have not been able to reproduce the error but it happened while you have the "No Filename" popup open telling you that "File has not been saved." There was also a blue popup open saying "Working...\nSpell Query\nPLease wait". The error message I saw said "12:43:17: Use of uninitialized value $msg in concatenation (.) or string at C:/Users/nqgc/Desktop/DP/Tools/guiguts-1.5.3-test3/lib/Guiguts/Utilities.pm line 774."
3. I thought that SQ used a more comprehensive English dictionary so was surprised to see the word "unimportance" flagged while SC does not flag it. This is a minor point and I may have misunderstood the dictionary issues.

windymilla:
Several places in the code (now only some of the scanno highlighting code) considered a word to be made up of alphanumeric characters. The problem is, now that PPers use combining characters more e.g. to add unusual accents to a variety of letters, these combining characters were being treated as word-breaking characters.

Spell Check can't easily be fixed, since it relies on Aspell which, on Windows at least, scarcely copes with Unicode, never mind combining characters. However, this commit fixes Spell Query, and makes WF search and normal Search/Replace (with the whole word option checked) do a better job of finding whole words.

There's also a minor tidy-up in Multi-language Spellcheck code. It used to copy the whole file out to a temporary file a line at a time, then read it back in a line at a time to process the words in it. Now it just works directly from the loaded file.

Fixes #1118
Fixes #1122

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -1472,7 +1472,7 @@ sub booklouperun {
             my $nextcol = 0;
 
             # Replace all non-alphanumeric (but not apostrophes) with space
-            $line =~ s/[^[:alnum:]'$APOS]/ /g;
+            $line =~ s/[^\p{Alnum}\p{Mark}'$APOS]/ /g;
 
             # Check each word individually
             my @words = split( /\s/, $line );    # Split on single spaces to aid column counting
--- a/src/lib/Guiguts/MultiLingual.pm
+++ b/src/lib/Guiguts/MultiLingual.pm
@@ -528,62 +528,24 @@ sub createseenwordslang {
 # Build basic list of distinct words
 sub buildwordlist {
     my $textwindow = shift;
-    my ( @words, $match, @savesets );
-    my $index = '1.0';
-    my $wc    = 0;
-    my $end   = $textwindow->index('end');
+    my @words;
+    my $index          = '1.0';
+    my $totalwordcount = 0;
     %distinctwords = ();
-    my $filename = $textwindow->FileName;
-    unless ($filename) {
-        $filename = 'tempfile.tmp';
-        open( my $file, ">", "$filename" );
-        my ($lines) = $textwindow->index('end - 1 chars') =~ /^(\d+)\./;
-        while ( $textwindow->compare( $index, '<', 'end' ) ) {
-            my $end  = $textwindow->index("$index  lineend +1c");
-            my $line = $textwindow->get( $index, $end );
-            print $file $line;
-            $index = $end;
-        }
-    }
-    ::savefile()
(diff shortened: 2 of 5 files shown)


windymilla:
Testing notes: [SQ-error-example.txt](https://github.com/DistributedProofreaders/guiguts/files/10949283/SQ-error-example.txt) contains some words with combining characters, e.g. `Ta̔i`
Spell Query now considers this a single word
Word Frequency also recognises it as a word, and if double-clicked in the word list, jumps to the right place.
The word `Ta` also exists in the file. Previous versions would fail to search correctly for `Ta`, finding the `Ta` at the start of `Ta̔i`, because the accent over the `a` was considered the end of a word.
Search/Replace with the Whole Word option checked can also be used to manually check for whole words, e.g. now it will find `Ta` once, and the Count button will say one occurrence, where previously it would find 3.
Multi-language spell check is also improved, but still relies on Aspell, so not worth testing.

cpeel (approved):

srjfoo (approved):

PR #1151, opened 2023-03-11
Use button/label font for toolbar

### LINKED ISSUE #1119: The "menus, buttons, labels" font is not used for the toolbar
If you change the font for menus, buttons, etc., from the system one to a custom one, the toolbar doesn't change because the font is hardwired to "Times 10 bold"

Where `$::lglobal{toolfont}` is currently used, change it to `'global'`. If there are complaints about it not being bold, could create a bold version of global to use.

While you're at it, have a look at the font used for "Working" dialog (hardwired to Helvetica 20 bold). Maybe it can be a larger version of the global font (e.g. add 10 to font size?)

windymilla:
Buttons & labels use either the system default font, or one that the user configures. The same font is now being used for the toolbar for consistency.

Also, the large bold font used in the "Working" popup is created as a named font (like the other fonts used) rather than declared ad hoc in arguments to the Label creation sub. This means it can be used of the same type of font is needed elsewhere without duplicating the definition.

Fixes #1119

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -786,7 +786,7 @@ sub working {
             $::lglobal{workpop}->title('Working.....');
             $::lglobal{worklabel} = $::lglobal{workpop}->Label(
                 -text       => "\n\n\nWorking....\n$msg\nPlease wait.\n\n\n",
-                -font       => '{helvetica} 20 bold',
+                -font       => 'largebold',
                 -background => $::activecolor,
             )->pack;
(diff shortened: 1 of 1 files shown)


windymilla:
Toolbar (can turn it on via Prefs->Toolbar menu) should now use the same font at labels and buttons throughout the interface (except the top level menus, at least on Windows, which are under control of the OS window manager). You can configure this font to either use the system default or define your own font via the Prefs->Appearance->Set Fonts button.

cpeel (approved):

srjfoo (approved):

PR #1152, opened 2023-03-11
Fix `Check for Updates`

### LINKED ISSUE #1120: Check for updates doesn't work
**Describe the bug**
I don't think the Github releases page format is what the code is expecting.
Change regex to match the response content to `/(\d+)\.(\d+)\.(\d+)<\/h2>/`
or something similar.
BTW, to print the response, use `$response->decoded_content`

Maybe add `if ($response->is_success) {` and error message if not OK

windymilla:
Format of Github releases page has changed, so previous regexp no longer worked.

Fixes #1120

### DIFF
--- a/src/lib/Guiguts/Utilities.pm
+++ b/src/lib/Guiguts/Utilities.pm
@@ -2189,25 +2189,49 @@ END
 }
 
 #
-# Check what is the most recent version of GG online
+# Check and return most recent version of GG available at Github
+# Assumes most recent production release is the first on the
+# "releases" page, with format "M.m.r</h2>", being Major number,
+# minor number, revision number in an h2 heading.
+# If release has a further suffix, like "-beta", we don't want to
+# offer it to user via this mechanism. They would access it
+# independently of GG via specific forum post or PM which would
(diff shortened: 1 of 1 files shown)


cpeel (approved):
Oh yes, this is so much more robust!

srjfoo (approved):

windymilla:
@cpeel I've requested another review if that's OK. @srjfoo discovered that the code didn't work on the Mac and a bit of online searching revealed a installing `LWP::UserAgent` on Macs has issues (depending on something else that requires other stuff and fails by default, I think unless you manually compile). Anyway, the suggestion was to use `curl` instead, so I've implemented that just for Macs. We wanted to run it by you again though, in case using `curl` is a bad idea. Thanks

srjfoo:
Many kudos to @windymilla for figuring out what the problem was. 👍

cpeel:
`curl` is a perfectly good thing to use on Macs. Bummer about `LWP::UserAgent` not working out of the box, but this is a perfectly good workaround.

cpeel (approved):

PR #1156, opened 2023-03-13
Allow spaces within block wrap parameters

### LINKED ISSUE #1132: Bug in Block Quote wrapping when specified Indent and Margins have a space between left & right margin specs
**Describe the bug**
/#[4.0, 72] (note space) causes wrapping to behaves differently to /#[4.0,72] (no space within margins specification)

**To Reproduce**
1. Open Guiguts with 
[rewrap-bug.txt](https://github.com/DistributedProofreaders/guiguts/files/10899530/rewrap-bug.txt)
2. Open the Tools menu.
3. Either do Rewrap All _or_ highlight block, including markers, and do Rewrap Selection.
4. See the way the block has been wrapped to include the margin specifications.

**Expected behavior**
It should not matter whether there are spaces between left & right margin specifications.

**Environment**
* Guiguts version: 1.5.3 but also present in 1.4.0 (so long-standing issue)
* OS and version: Windows 10
* Perl version: 
* Perl distribution: 

**Screenshots**

**Additional context**
I wouldn't normally have a space within the /#[4.0,72] specification but did not initially twig that that was the cause of the incorrect wrapping. A space or spaces there should not cause problems if specifications within the square brackets are parsed properly. Software should not take users by surprise like that!

windymilla:
User can specify margins, e.g. `/#[12.4,50]`
Spaces are now permitted at start and around period and comma, e.g. `/#[ 12 . 4 , 50 ]`

Fixes #1132

### DIFF
--- a/src/lib/Guiguts/SelectionMenu.pm
+++ b/src/lib/Guiguts/SelectionMenu.pm
@@ -38,7 +38,7 @@ sub knuth_wrapper {
 
     # if open rewrap markup, remove, then prepend once rewrapped
     if ( $paragraph =~
-        s|^($TEMPPAGEMARK*/$TEMPPAGEMARK*[$blockwraptypes]$TEMPPAGEMARK*\[$TEMPPAGEMARK*[0-9.,$TEMPPAGEMARK*]+$TEMPPAGEMARK*\]$TEMPPAGEMARK*)||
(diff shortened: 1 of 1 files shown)


windymilla:
Testing should be fairly straightforward: use `/#` markup, and add `[20]` or `[20, 40]` or `[20.6, 40] after it. Spaces are now permitted within the square brackets (but not before, e.g. `/#   [30]` won't work.

cpeel (approved):

srjfoo (approved):

PR #1158, opened 2023-03-14
X11 setup for macOS

srjfoo:
Add instructions for adjusting X11 preferences. These preferences will allow (1) Greek to be typed directly into GG, and (2) commands to be passed through directly from popup menus with just one click.


windymilla (approved):
I think that looks good - there's no need for anything more technical. If users end up not liking those Prefs, they can always undo them of course.

cpeel (approved):
Fantastic - thanks for figuring these out and updating the docs!

PR #1159, opened 2023-03-14
Accept wider range of page separator lines

windymilla:
When a CTF is first loaded into Guiguts, it detects the page separator lines and inserts page marks.
Other sources, e.g. gImageReader OCR files, also have page separator lines.

Widen the accepted format of page separator lines to include anything of the form `...File:...<page>.png...` where `...` are any other characters, and `<page>` is the basename of the scan file. This will accept CTFs and gImageReader OCR files, but is sufficiently tight that it will not accidentally match a line of text in one of our books.

### DIFF
--- a/src/lib/Guiguts/FileMenu.pm
+++ b/src/lib/Guiguts/FileMenu.pm
@@ -456,23 +456,35 @@ sub file_mark_pages {
     my $textwindow = $::textwindow;
     $top->Busy( -recurse => 1 );
     ::hidepagenums();
-    my ( $line, $index, $page, $rnd1, $rnd2, $pagemark );
-    $::searchstartindex = '1.0';
-    $::searchendindex   = '1.0';
-    while ($::searchstartindex) {
-        $::searchstartindex =
-          $textwindow->search( '-nocase', '-regexp', '--', '-*\s?File:\s?(\S+)\.(png|jpg)---.*$',
-            $::searchendindex, 'end' );
-        last unless $::searchstartindex;
-        my ( $row, $col ) = split /\./, $::searchstartindex;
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: Open a CTF, and the page separators should be recognised - look in the Img label in the status bar as you scroll through the file. This should work in any release.
The new bit is that the same should happen when you load this file: [ocred.txt](https://github.com/DistributedProofreaders/guiguts/files/10973388/ocred.txt)

windymilla:
@srjfoo found a problem, so this will need another commit, maybe tomorrow

windymilla:
OK, I think that commit will be better. It now accepts a wider range of page separator lines, but if the format is not the standard one (or close enough that other code will accept it) it rewrites the line in standard format. 
Testing notes are the same as before - also try exporting as you did previously and check it doesn't include the page separator line at the top of the exported files.
Thanks

srjfoo (approved):

srjfoo:
> OK, I think that commit will be better. It now accepts a wider range of page separator lines, but if the format is not the standard one (or close enough that other code will accept it) it rewrites the line in standard format. Testing notes are the same as before - also try exporting as you did previously and check it doesn't include the page separator line at the top of the exported files. Thanks

When the test file with non-standard page separators was opened, I didn't even see the non-standard page separators; the page separators were standardized upon import. It just occurred to me to ask if that was the effect that you wanted.

windymilla:
Yes - I think that's safer than having an explicit menu button that people have to use to convert the page separators to the standard form. There is some precedent, e.g. automatic correction of Fotonote to Footnote, etc.

cpeel (approved):

PR #1162, opened 2023-03-17
HTML-convert `<`, `>` that can't be part of a tag

### LINKED ISSUE #1135: HTML convert doesn't convert all lt and gt symbols
**Describe the bug**
Using autogenerate to convert `1>2` or `a<b` leaves the symbols in the HTML, potentially causing validation errors later.

Note to self: rather than making the current regexps even more comples, it might be easier to convert all angle brackets that **are** part of an HTML tag to other characters temporarily, then convert all remaining lt/gt, then restore the tag-delimiting ones.

windymilla:
When a text file is converted to HTML, angle brackets that are part of the text (not tags) should be converted to HTML entities. Previous code didn't convert text like `1<2` or `a>b` Current code converts anything except a tag consisting of `<` followed by any number of characters except angle brackets then a closing `>`.  This is broader than tags that will typically be found, but safer. It is better not to convert something that should be (since the HTML validator should catch it) than to convert something that shouldn't (since that may not trigger the validator and could mean HTML tag code will appear in the text.)

Fixes #1135

### DIFF
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -80,15 +80,34 @@ sub html_convert_simple_tag {
 
 #
 # Convert ampersands, greater than and less than symbols to HTML code
-# Regexps needed for gt and lt to avoid converting those that are part of HTML tags
-# Note that some gt/lt will not be converted, e.g. 2>1 or a<b (likely causing validation error)
+# Called at start of HTML conversion. The "<" or ">" symbols may be used in
+# DP tags, e.g. </sc>. Normally there should not be full HTML tags, but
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: [ampersands.txt](https://github.com/DistributedProofreaders/guiguts/files/11001089/ampersands.txt) has some angle brackets in. Those that look vaguely like tags shouldn't be converts to HTML entities when you use HTML Autogenerate on the file.

srjfoo (approved):
I don't say it often enough, but I really appreciate the testing notes and the test files for those changes that need a test file!

cpeel (approved):

PR #1165, opened 2023-03-19
Remove unnecessary tidy up after HTML generation

windymilla:
Old check for `<br>` or `<br/>` or `<br />` is no longer needed, IMHO


windymilla:
Testing notes: Convert every possible text file to HTML and see if the results are the same with and without this code :)
Seriously, I don't think the HTML generation code will add spurious `<br>` lines immediately after a heading. I haven't seen it, and don't know what would generate it.

cpeel (approved):

srjfoo (approved):

PR #1168, opened 2023-03-20
Add `Convert DP diacriticals` feature

### LINKED ISSUE #1154: Convert DP-notation of diacriticals
This isn't a suggestion, just a suggestion of something to consider elevating to a suggestion.  The Proofreading Guidelines show notation to be used to represent diacriticals such as Ṣ, which should be typed in as [S.], or ō, which should be typed in as [=o]. Ideally, the Project Managers of projects containing diacriticals will include the Latin Extended A, Latin Extended Additional, and/or other Unicode blocks appropriate to the project, in which case the Proofreading Interface will automagically convert the typed notation to the actual Unicode character.

However, when the PM hasn't included the relevant Unicode Blocks, the project may arrive on the PP'ers desk with a lot of DP-notation.  In such cases, it would be helpful to have a Tools>Character Tools option similar to "Normalize Selected Characters", i.e., "Convert DP-style Diacriticals".

The project I'm just starting has 524 of them.  I'd hoped that converting the DP-notation to combining-character diacriticals and then using "Normalize Selected Characters" would work, but "Normalize" didn't recognize those diacriticals.

This situation may be far too rare to merit an enhancement to Guiguts, and a series of S&R's should successfully convert what's in this current project, so please don't elevate this to an actual "Feature request" unless you think it'll be used sufficiently often to justify the implementation effort.

If you are looking for a suitable text file to use for testing, see the one in [THIS "City of Jerusalem"](https://www.pgdp.net/c/project.php?id=projectID6070a15488753) DP project.

windymilla:
Within selected range(s), or whole file, convert DP diacritical markup, e.g. `[=a]` or `[a.]` to the Unicode character(s) it represents. If a single normalized character exists, uses that, otherwise uses combining character.

Also, minor bug fix to similar code in fraction conversion: as the conversions are made, if there is more than one on a line, the "end" index becomes stale, since the replacement text is a different length to the original. Using a mark in place of an index fixes this.

Fixes #1154

### DIFF
--- a/src/lib/Guiguts/CharacterTools.pm
+++ b/src/lib/Guiguts/CharacterTools.pm
@@ -1206,19 +1206,22 @@ sub fractionconvert {
 # Only attempts to convert if diacritical is applied to an alphanumeric character
 sub convertdiacriticalmarks {
     my $textwindow = $::textwindow;
-    my %diacabv    = (                # Hash of combining diacriticals above and below, indexed by DP markup character
-        '=' => [ "\x{0304}", "\x{0331}" ],
-        ':' => [ "\x{0308}", "\x{0324}" ],
-        '.' => [ "\x{0307}", "\x{0323}" ],
-        '`' => [ "\x{0300}", "\x{0316}" ],
-        "'" => [ "\x{0301}", "\x{0317}" ],
-        '^' => [ "\x{0302}", "\x{032D}" ],
-        'v' => [ "\x{030C}", "\x{032C}" ],
-        ')' => [ "\x{0306}", "\x{032E}" ],
-        '(' => [ "\x{0311}", "\x{032F}" ],
-        '~' => [ "\x{0303}", "\x{0330}" ],
-        ',' => [ "",         "\x{0327}" ],    # Empty string ensures conversion won't happen below
-        '*' => [ "\x{030A}", "\x{0325}" ],
+
+    # Hash of combining diacriticals above and below, indexed by DP markup character, taken from
+    # https://github.com/DistributedProofreaders/dproofreaders/blob/master/tools/proofers/process_diacritcal_markup.js
+    my %diacabv = (
+        '=' => [ "\x{0304}", "\x{0331}" ],    # macron
+        ':' => [ "\x{0308}", "\x{0324}" ],    # diaeresis
+        '.' => [ "\x{0307}", "\x{0323}" ],    # dot
+        '`' => [ "\x{0300}", "\x{0316}" ],    # grave
+        "'" => [ "\x{0301}", "\x{0317}" ],    # acute
+        '^' => [ "\x{0302}", "\x{032D}" ],    # circumflex
+        ')' => [ "\x{0306}", "\x{032E}" ],    # breve
+        '~' => [ "\x{0303}", "\x{0330}" ],    # tilde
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes: The text below might be useful- note that there is no "cedilla above" conversion, so it should remain unchanged.
Also, markup with non-supported diacritical character, e.g. `[%a]`, should not change
Also, any other alphanumeric should work in place of `a`, e.g. `[=α]` (Greek alpha)

```
=[=a][a=]
:[:a][a:]
.[.a][a.]
`[`a][a`]
'['a][a']
^[^a][a^]
v[va][av]
)[)a][a)]
([(a][a(]
~[~a][a~]
,[,a][a,]
*[*a][a*]
```

src/lib/Guiguts/CharacterTools.pm
```diff
@@ -1200,6 +1201,65 @@ sub fractionconvert {
     $textwindow->addGlobEnd;
 }
 
+#
+# Convert common DP diacritical markup into normalized Unicode characters
+# Only attempts to convert if diacritical is applied to an alphanumeric character
+sub convertdiacriticalmarks {
```
cpeel, line 1207: Consider adding a link to the DP code that defines the set used there.
https://github.com/DistributedProofreaders/dproofreaders/blob/master/tools/proofers/process_diacritcal_markup.js

windymilla: Thanks - I've taken the var definitions from there and Perl-ified them, so the order of diacriticals in GG's hash definition now match the DP code, just in case another markup is ever added to the DP code.

srjfoo (approved):
Is it worth documenting somewhere (maybe in the GG manual) that how converted characters display is very font dependent?

When tested this, I had to remind myself that the only ones that didn't get converted were the ones that still had brackets around them -- everything else actually did get converted, even if it seemed to display as two separate characters beside each other. Helvetica is a good font for checking this out, even though it's a proportional font. Courier New (the default font in GG) looks pretty miserable for most of the combinations that don't have a normalized version.

windymilla:
> Is it worth documenting somewhere (maybe in the GG manual) that how converted characters display is very font dependent?
Possibly, although tbh, the DP mono font is good at displaying the combining characters superimposed, then splitting them as the cursor gets to them

srjfoo:
> > Is it worth documenting somewhere (maybe in the GG manual) that how converted characters display is very font dependent?
> 
> Possibly, although tbh, the DP mono font is good at displaying the combining characters superimposed, then splitting them as the cursor gets to them. Courier is the default font because user might not have the DP font, but they are told to use it.

On my Mac:
DPSM:
image
vs.
Helvetica:
image

Yes, the cursor can tell where there are two characters displayed as one -- that's not so much the issue. Anyway -- just making a note of it, because it may cause concerns for some PPers.

windymilla:
This is DPSM on Windows. I'll mention it to Charlie though, even if it's only for Mac users, it's worth putting in
image

srjfoo:
Update -- @windymilla asked if this was specific to GG, and I checked it out in BBEdit, LibreOffice, Pages and TextEdit, all of which appear to match the Windows GG display. I think a warning to Mac users should be sufficient.

PR #1169, opened 2023-03-20
Allow `/r[nn]` to set a right margin

### LINKED ISSUE #1145: Allow /RR/ markers to specify a new right margin
/R[_right_] would specify a new right margin _right_ to which blocks of text are pushed up against when shifted right. It would override the default right margin for that /RR/ block only. The default left margin would be used in any calculation when deciding if a block can be shifted left in order to align its longest line against the new right margin.

I had previously proposed /R[_right indent_] which would mean indent from the right margin by _indent_ spaces when shifting a block of text right. However this is not consistent with the way in which indents are specified using new margins on /# markers. The proposal above is now consistent with that.

**Additional context**
This change could apply to Text Files only as is the case when specifying indents via new margins on /# markers. However there is no obvious reason why it cannot apply to the HTML File too. In the latter case the indent from the right margin could, say, use 0.5em as the unit of padding.

windymilla:
When right-aligning a block, allow the user to specify a right margin for any block, rather than the default right or blockquote right margins.

When generating HTML, the right margin on the spans is adjusted by the amount necessary to allow 0.5em per character from the default right or blockquote right margins.

It's not appropriate to allow setting of a left margin nor a first-line left margin on a right-aligned block as is used on `/#` blocks

Fixes #1145

### DIFF
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -304,13 +304,14 @@ sub html_convert_body {
     my @last5            = [ '1', '1', '1', '1', '1', '1' ];
     my $step             = 1;
     $thisblockend = $textwindow->index('end');
-    my $blkcenter = 0;
-    my $blkright  = 0;
-    my $blkrstart = 0;               # Value of $step when a block right starts
-    my @blkrlens  = ();
-    my $closure   = voidclosure();
+    my $blkcenter      = 0;
+    my $blkright       = 0;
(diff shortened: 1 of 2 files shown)


windymilla:
Testing: Try this adaptation of the previous right-wrap file: [RR-rewrap-bug.txt](https://github.com/DistributedProofreaders/guiguts/files/11023134/RR-rewrap-bug.txt)

cpeel (approved):

srjfoo (approved):

PR #1171, opened 2023-03-21
Three minor bugs

windymilla:
1. Spaces are displayed in the WF Character Count list as `*space*`.
The number of occurrences was set to undef if the file contains
no spaces because of a missing check when space display was
handled specially. Check was there for non-breaking spaces,
but not regular ones. This caused an error when displaying the
character counts.
Also fixed similar (but more minor) error if file contains no
newline characters. Character count displayed a line saying
there were no newlines, instead of omitting mention of
newlines as happens with other characters that aren't in
the file.

2. Similar to loading text/HTML file, don't use `encoding(utf8)`
on file open of global dictionary, because if it isn't utf8, bad
things will happen.

3. If file has a `.bin` file, it may have language(s) set in it.
If it does not, the language(s) remained as for the previously
loaded file.

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -1627,7 +1627,7 @@ sub booklouperun {
     sub spellqueryloadglobaldict {
         my $dictname = shift;
         my $fh;
-        unless ( open $fh, "<:encoding(utf8)", $dictname ) {
+        unless ( open $fh, "<", $dictname ) {
             ::warnerror("Error opening Spell Query dictionary: $dictname");
             return 0;
         }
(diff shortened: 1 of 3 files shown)


cpeel (approved):

windymilla:
Testing notes:
1. Create a file with no spaces. Run Tools->WF, then Character Cnt. Also try a file with no newlines.
2. [testfiles.zip](https://github.com/DistributedProofreaders/guiguts/files/11033429/testfiles.zip) contains a text file, `bug1-it.txt` and two dictionaries. Put the dictionaries in your GGPrefs folder. `dict_ita_user` is ISO-8859, and `dict_itu_user` is utf8 encoded. To use them, load the text file, set the language (in status bar) to `ita` or `itu` and run Spell Query. Without this PR, `ita` will give lots of errors. With, it shouldn't, and the 3 italian words near the top should not be flagged as misspelled.
3. Load a file and set its language to anything you like, then load a file that doesn't have a .bin file - without this PR, the language will transfer to the new file. It shouldn't - the default is English for a new file until told otherwise, as happens if you quit the program, restart and then load your new file.

srjfoo (approved):

PR #1174, opened 2023-03-23
Quicksearch enhancements

### LINKED ISSUE #1130: Enhancements to "Quick Search"
Two additional features to consider for "Quick Search":
1. If something is user-highlighted (in blue) in the main window when shift+ctrl+f is pressed , place what's highlighted in the QS search bar. If nothing is highlighted, keep whatever's there.  Objection user may want to re-use what's already there. However, QS History will list that.

2. Add a "Count" button to the Quick Search options. Useful when resolving remaining split words -* and the main S&R dialog is already being used to find those splits.  To avoid making the QS dialog wider, it's tempting to suggest a very narrow button to the left of the "Search" button, with the # sign as its caption; but that may not have a clear meaning.

windymilla:
Fixes #1130 with two small features:
1. If text is selected when QS is popped, the selected text appears as the search string (like main S/R dialog)
2. Small `#` button to count occurrences of search text in the selected region, or the whole file if nothing is selected.

### DIFF
--- a/src/lib/Guiguts/SearchReplaceMenu.pm
+++ b/src/lib/Guiguts/SearchReplaceMenu.pm
@@ -2272,11 +2272,18 @@ sub quicksearchpopup {
         -height => 15,
     )->pack( -side => 'left', -anchor => 'nw' );
     $::lglobal{statussearchtext} = '' unless defined $::lglobal{statussearchtext};
+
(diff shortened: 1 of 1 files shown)


cpeel (approved):
Really appreciate the great comments that make the code easier to understand!

srjfoo (approved):

PR #1175, opened 2023-03-23
Don't save file on some WF operations

windymilla:
Without warning, and ignoring autosave settings, some WF operations, including Character Count, silently saved the file, then read it all back in again.

Instead, just get the text from the currently loaded version of the file.

Also, tidy the way Character Count displayed some whitespace characters as, e.g. `*space*`. Spaces and nbsp were handled one way, newlines another way, tabs another way. Now all handled the same, and easily expandable when someone requests another non-visible character, e.g. thin-space, to get the same treatment.

### DIFF
--- a/src/lib/Guiguts/WordFrequency.pm
+++ b/src/lib/Guiguts/WordFrequency.pm
@@ -444,7 +444,7 @@ sub bangmark {
     $::lglobal{wclistbox}->delete( '0', 'end' );
     $::lglobal{wclistbox}->insert( 'end', 'Please wait, building list....' );
     $::lglobal{wclistbox}->update;
-    my $wholefile = slurpfile();
+    my $wholefile = $::textwindow->get( '1.0', 'end' );
 
     while (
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: Edit a file, don't save it, then run Character Counts within Tools->Word Frequency.
Previous versions would silently save your file.
The number of newlines reported may be one different to previously. The current version is as correct as the previous :)

srjfoo (approved):

src/lib/Guiguts/WordFrequency.pm
```diff
@@ -920,47 +920,37 @@ sub charsortcheck {
...
-        my @words = split( //, $line );
-        foreach (@words) {
-            $chars{$_}++;
-            $wordw++;
-        }
-        $index++;
-        $index .= '.0';
+    my $wholefile = $::textwindow->get( '1.0', 'end' );
+    $wholefile = lc($wholefile) if $::lglobal{wf_ignore_case};
+    my @words = split( //, $wholefile );    # Split into individual characters
```
cpeel, line 929: I can't remember - how does this work with Unicode characters? Is this a byte-level separation or a Unicode codepoint separation? Ad what does this do for combining characters?

PR #1176, opened 2023-03-24
Make ruler cope with long lines

windymilla:
The ruler was built from a 190 character wide string. If the line length was longer than that, it could attempt an out of range `substr` call.

Instead, just build a 190 character long ruler that shows the correct range of columns.

Tested beyond a million character length line OK.

### DIFF
--- a/src/lib/Guiguts/LineNumberText.pm
+++ b/src/lib/Guiguts/LineNumberText.pm
@@ -359,13 +359,19 @@ sub _lincolupdate {
     # Now insert the column numbers
     if ( $w->{'ctext'}->ismapped ) {
 
-        # Note that on Mac, ROText widget appears to have a width limit of 192.
-        # If the string below is more than that, its first characters don't get displayed
-        my $ruler = substr(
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: @srjfoo - you broke it - you test it! :)
I think it would have broken when the leftmost column number was higher than 190, so the overall maximum line length previously would be 190 plus your screen width.
I've tried a million character line and the ruler seems to display correctly, although the letters within the line of text itself didn't always display properly at that line length - I'm not worried about that though.

srjfoo (approved):

srjfoo:
> The optimizer in me wanted to figure out how to make this more efficient (because that's just how my brain works) but I don't think it needs to be more efficient.
> 
> If this turns out, for some odd reason, to need to be faster I'm happy to help problem solve that.

In testing, I didn't see any delay on the growth of the ruler as the line length grew, so if that's the measure of whether the optimization is needed, probably not. (That being said, I admire efficient code :D)

PR #1178, opened 2023-03-28
Avoid error converting fractions in whole file

windymilla:
The mark for the end of the section where fractions were to be converted was not set if user did not make a selection, i.e. to convert whole file.

Also remove unused variable


windymilla:
Testing notes: Add some fractions to a file, e.g. 1/2, 4-1/4, 27/45, then use Tools, Convert Fractions and any of the conversion methods. You can also select a region and convert fractions within the selection. The error used to occur when you didn't selected a region, so attempting to convert the whole file.

Caused by "minor fix" during #1168

PR #1182, opened 2023-04-04
Make pphtml report fewer CSS false positives

windymilla:
1. Some HTML elements had special treatment to avoid their CSS being reported as unused. Extend this list to include `a`, `img` and `hr`
2. Previously a selector had to have `{` following it on the same line to be recognised - also permit it to be on its own on the following line
3. Don't report `@media` query types as unused classes

### DIFF
--- a/src/lib/ppvchecks/pphtml.pl
+++ b/src/lib/ppvchecks/pphtml.pl
@@ -214,32 +214,19 @@ sub runProgram {
             }
             $intextbody = 1;
 
-            # special case <h?>
-            if (/<(h\d)/) {
-                my $h = $1;
-
-                #        $h =~ s/<(h\d).*$/$1/;
-                $classes_used{$h} += 1;
-                $classes_line{$h} = $count unless exists $classes_line{$h};
+            # special code to also allow storage of some HTML elements in list of classes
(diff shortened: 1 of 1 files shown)


windymilla:
Sorry about the second pphtml PR in quick succession. Additional false positives were reported by the user.
Testing notes: [This test file](https://github.com/DistributedProofreaders/guiguts/files/11151952/the-collectors-whatnot.zip) or edited versions of it should suffice. Just run HTML->pphtml, and see the list of unused/undefined classes.
Note that the wording "unused/undefined" as used by pphtml could be misleading, since you never really "define" a class in CSS. You either label an element with the class in the HTML, or you select using the class in CSS. "Unused" means that the class name is never used to label an element. "Undefined" means the class is never used to select in the CSS. Also note that since the CSS/HTML is not parsed properly, just checked with a few regexes, there are still going to be many ways for uses to slip through the net, either leading to false positives (e.g. with a comment in the CSS) or false negatives (e.g. `p.xyz` used as a selector, but `img class="xyz"` used in HTML).

cpeel (approved):

srjfoo (approved):

PR #1190, opened 2023-04-19
Fix S/R in selected range end-point failure

### LINKED ISSUE #1189: Replace-all over a range of lines doesn't process the entire range
Here is an old bug that has resurfaced in GG 1.6.0; it doesn't happen in GG 1.5.3. It's easily reproducible using the attached file and the regex below.

Go to bookmark 1 on page 21. The input to AutoTable was improperly formatted (by me .. didn't leave 2 spaces between columns), so instead of a right column and a left column, there's just one column.  The regex below should split each `<td>` into two `<td>s` ... but it stops before reaching the end of the selected range.

Here is the Search argument: `"tdr">(\d) (.+?)</td>`
Here is the Replace argument: `"tdr">$1</td>\n<td class="tdl">$2</td>`

1. Select the table. It begins on line 1856 and ends on line 1977.
2. click Rpl All

The last ten rows of the table are not replaced.  If the entire table is selected again and Rpl All is clicked again, 8 of the 10 are processed, still leaving the last two unprocessed.

If R&S is done manually, it works fine. if nothing is selected, but the cursor is positioned at the top of the file and Rpl All is clicked, it works fine.  So, it only fails when a somewhat long range is selected.

[searchreplacebug.zip](https://github.com/DistributedProofreaders/guiguts/files/11240800/searchreplacebug.zip)

windymilla:
When using Replace All in a selected range, if the replacement adds a new line, the replacing stopped part way through the range.

This was caused by the end point being stored as a line number. The extra line in the replace string caused the actual line number of the intended end point to increase, so the replacement stopped early. Fixed by using a mark instead of the line number to note the end point.

Fixes #1189

### DIFF
--- a/src/lib/Guiguts/SearchReplaceMenu.pm
+++ b/src/lib/Guiguts/SearchReplaceMenu.pm
@@ -59,8 +59,10 @@ sub searchtext {
 
     # this is starting a search within a selection
     if ( $range_total > 0 ) {
-        $end                        = pop(@ranges);
-        $start                      = pop(@ranges);
+        $end   = pop(@ranges);
+        $start = pop(@ranges);
+        $textwindow->markSet( 'searchendmark', $end );      # Mark end of selection
+        $end                        = 'searchendmark';      # Use mark instead of line number in case replace adds/deletes lines
         $::lglobal{selectionsearch} = $end;
         $::searchstartindex         = $end if $::sopt[2];
 


windymilla:
The linked issue has a good test file, but any replacement string which adds a new line would show the same effect, effectively reducing the selected range by one line for every replacement made.

Guiguts S/R idiosyncrasy: Enter the search and replace strings in the S/R dialog before going back to the main window and selecting the intended range, then back to the S/R dialog again to do Replace All.

cpeel (approved):

srjfoo (approved):
NTS: Remember to check the "regex" box. 😁

Issue #1186, opened 2023-04-14
Illustration Fixup: Auto Move Illustrations to paragraph breaks


charliehoward4dp:
[Illustration] tags, with or without a preceding asterisk, often occur mid-paragraph, either because there were no paragraph breaks on the same page or the illustrations are full-page.  Best Practices encourages post-processors to move illustrations so that they are between paragraphs. This is not difficult, but it can be tedious, particularly when there are a lot of illustrations in a book.

A tool to do these moves automatically could be a real time-saver in projects containing a lot of illustrations.  Here are some thoughts about possible rules for doing this:

1. if the _next_ page is a [Blank Page] (or two consecutive page separators or internal page labels), try to move the illustration up to the last paragraph break on the page preceding it. If successful, this probably will place the illustration on the page referenced in the List of Illustrations.
2. if the _preceding_ page is a [Blank Page] etc., try to move the illustration down to the first paragraph break on the page following it. (Same LoI comment.)
3. if neither of the first two conditions exists, find the closest paragraph break in either direction and move the illustration to it.
4. in all cases, be sure not to move an illustration to a position within another illustration, past another illustration, past a block marker (except a Block Quote, within which it should be acceptable), within a footnote, or within a Chapter heading.

If implemented, it could be "Illustration Fixup" on the Tools Menu.

I know this is feasible, as I wrote a Python script to do most of it in 2018 (see [THIS](https://www.pgdp.net/phpBB3/viewtopic.php?f=13&t=65851) post in the DP Forums).  The results generally are very good but not always perfect, and AFAIK, I remain the only person using it.  It handles Sidenotes and Footnotes as well as Illustrations, and originally was written mostly to handle a six-volume series with _copious_ Sidenotes.  (It's only at Github now; the other website no longer exists.)

charliehoward4dp:
Attached are three zips, each one containing two files: the original projectxxx.txt as received from the Rounds, and the projectxxx_moved.txt, which is the output of the Python script.

The procedure worked very well for example1 and example2, but not too well for example3. Most of the unmoved illustrations in example3 probably could have been moved if the script was a bit more sophisticated.

The script does not flag what it couldn't move, but it does add a summary of results at the very top of the output file (as you will see when looking at them). It's necessary for the user to check the results and correct/complete any errors or omissions.

Thank you for even considering to add this feature, and please don't hesitate to ask for more information.

[example1.zip]
(https://github.com/DistributedProofreaders/guiguts/files/1129
[...]

Issue #1195, opened 2023-05-12
Provide Sidenote Fixup similar to new Illo Fixup


windymilla:
Create Sidenote Fixup submenu and put existing feature in that as "Automatic".
New feature will be "Manual"

Issue #1194, opened 2023-05-10
Recursive Busy doesn't seem to always Unbusy properly


windymilla:
**Describe the bug**
Where Busy is called in ErrorCheck, the Unbusy doesn't seem to unbusy fully, in that if you Ctrl-Click an Illo Check error having previous caused the bell to ring (e.g. clicking while it's actually busy, or trying to move an unclosed illo) then all further Ctrl-clicks cause the bell to ring in `Widget::_busy` 

Consider removing the `recurse=>1` from the Busy call, unless there's something wrong in the logic somewhere.

windymilla:
See #1211 - not to do with "unbusying", but due to "fake" shift/control key presses being generated when focus switches to text window while shift/control held down.

PR #1196, opened 2023-05-12
Add Illustration Fixup feature to Tools menu

windymilla:
Check and report on all illustration markup in downloaded CTF. Allow user to move illos up/down a paragraph, either to ensure illo is not mid-paragraph or because illo fits better at the new location.

Ctrl-click to move illo down. Ctrl-shift-click to move illo up. Once an illo is moved, Ctrl+shift+j jumps between original and new locations for illo.  The "original" location is only stored for the most recently moved illo, so if you move one illo, then a different one, then go back to move the first again, the "original" position of the first is where you left it after its previous move, not its position when you first loaded the file. This is implemented with a general "system bookmark" feature, which could be useful in other places. Menu entry corresponding to Shift+Ctrl+j is in the Bookmarks menu.

Undo/redo have had to have a slight boost, so that undoing an illo move can refresh the list.

### DIFF
--- a/src/lib/Guiguts/MenuStructure.pm
+++ b/src/lib/Guiguts/MenuStructure.pm
@@ -480,28 +480,28 @@ sub menu_tools {
         ],
         [
             'command',
-            'Unmatched DP Ta~g Check',
+            'Unmatched DP Ta~g Check...',
             -command => sub {
                 ::errorcheckpop_up( $textwindow, $top, 'Unmatched DP Tags' );
             }
         ],
         [
             'command',
-            'Unmatche~d Brackets Check',
+            'Unmatche~d Brackets Check...',
             -command => sub {
                 ::errorcheckpop_up( $textwindow, $top, 'Unmatched Brackets' );
             }
         ],
         [
             'command',
-            'Unmatched ~Block Markup Check',
+            'Unmatched ~Block Markup Check...',
             -command => sub {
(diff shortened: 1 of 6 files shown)


windymilla:
Fixes #1186 

Intention is to add a similar feature for Sidenotes as described in #1195 

Also note that bell might ring unnecessarily while moving illos - appears to be a Perl/Tk bug as described in #1194

srjfoo (approved):
Small re-wording suggestion ...

> Once an illo is moved, Ctrl+shift+j jumps between original and new locations for illo.

to

> Once an illo is moved, Ctrl+shift+j jumps the cursor between original and new locations for illo. Use Ctrl+z to move it back to the original location.

(At least that's how it worked for me on macOS.)

src/lib/Guiguts/MenuStructure.pm
```diff
@@ -480,35 +480,42 @@ sub menu_tools {
...
-            'Unmatched ~Block Markup Check',
+            'Unmatched ~Block Markup Check...',
             -command => sub {
                 ::errorcheckpop_up( $textwindow, $top, 'Unmatched Block Markup' );
             }
         ],
         [
             'command',
-            'Load Checkf~ile...',
+            'Load Checkfile~...',
```
cpeel, line 504: Does this make `.` the shortcut key? That seems a little ... odd.

cpeel (approved):
Apologies for the long delay in getting to this. Life has been a bit nuts lately.

PR #1199, opened 2023-05-16
Improve pphtml report of unused/undefined classes

windymilla:
Code was more complicated than necessary, and still reported things like `h1` possibly not defined.

Use common code for undefined/unused class check

### DIFF
--- a/src/lib/ppvchecks/pphtml.pl
+++ b/src/lib/ppvchecks/pphtml.pl
@@ -330,9 +330,10 @@ sub runProgram {
         my $class = shift;
         my @list  = shift;
         return (
-            grep      { $_ eq $class } @list                                                       # class is used/defined
-              or grep { $_ eq $class } qw(a blockquote body h1 h2 h3 h4 h5 h6 hr img ins p table)  # ignore these element names
-              or $class =~ /^x-ebookmaker/                                                         # special ebookmaker class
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes: A suitable file to start from is https://github.com/DistributedProofreaders/guiguts/blob/master/tests/errorcheck.html
You could add/remove CSS and/or use of classes to that, then run HTML->pphtml

cpeel (approved):

src/lib/ppvchecks/pphtml.pl
```diff
@@ -317,52 +300,40 @@ sub runProgram {
...
+    }
 
+    # Return true if class/element is in given list (i.e. used/defined)
+    # or if it's one we want to ignore, or it's a special ebookmaker class
+    sub classisknown {
+        my $class = shift;
+        my @list  = shift;
+        return (
+            grep      { $_ eq $class } @list                                                       # class is used/defined
+              or grep { $_ eq $class } qw(a blockquote body h1 h2 h3 h4 h5 h6 hr img ins p table)  # ignore these element names
```
srjfoo, line 334: Should `div` be included in this list, or is that too borderline, and all divs are expected to have an associated class?

srjfoo (approved):
Looks good to me -- my question is more along the lines of "this was one of the things I tested, so I'm asking about it" 😁

PR #1203, opened 2023-05-27
Only report unused/undefined classes in pphtml

### LINKED ISSUE #1202: PPHTML flagging almost everything (incorrectly) in GG 1.6.0-test3; OK with test2
PPHTML flags almost everything (incorrectly) in GG 1.6.0-test3; OK with test2, test1, and 1.5.3. Happens with both of the files I tested; one is attached for testing.

[pphtmltest.zip](https://github.com/DistributedProofreaders/guiguts/files/11582817/pphtmltest.zip)

windymilla:
1. Embarrassing use of `shift` only popped first element of array of used/defined classes passed to `classisknown`, so it reported almost all classes as both undefined and unused.
2. Added to list of HTML elements to ignore
3. Cope with when two classes/element names are combined in a selector using space, comma or plus
4. Ignore id selectors, since only attempting to find classes

Fixes #1202

### DIFF
--- a/src/lib/ppvchecks/pphtml.pl
+++ b/src/lib/ppvchecks/pphtml.pl
@@ -286,11 +286,13 @@ sub runProgram {
         foreach my $idx ( 0 .. $#css ) {
             my $cssdef = $css[$idx];
             my $count  = $cssline[$idx];
-            $cssdef =~ s/\@media\s+[^\{]+\{//;    # Remove any @media query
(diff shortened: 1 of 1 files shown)


cpeel (approved):
Whoops! We've all done similar things though!

srjfoo (approved):

srjfoo:
> Whoops! We've all done similar things though!

Most definitely!

PR #1206, opened 2023-06-01
Report unused element selectors in pphtml

### LINKED ISSUE #1204: PPHTML missed a simple, unused selector
The attached files does not use `<h3>` but it's in the CSS at line 18, and PPHTML didn't mark it as "possibly unused".  This occurs with the updated PPHTML of 5/28/23.

[pphtmlsample2.zip](https://github.com/DistributedProofreaders/guiguts/files/11587792/pphtmlsample2.zip)

windymilla:
Although it was correct to ignore elements that are not used as selectors in CSS, it was wrong to ignore CSS selectors that use elements which don't occur in the file body.

Also, tidy up uses of `while` with regexps to avoid manipulation of the `$_` variable.

Fixes #1204

### DIFF
--- a/src/lib/ppvchecks/pphtml.pl
+++ b/src/lib/ppvchecks/pphtml.pl
@@ -295,7 +295,7 @@ sub runProgram {
             $cssdef =~ s/\@media\s+[^\{]+\{//;      # Remove any @media query
             $cssdef =~ s/^(.*?)\{.*$/$1/;           # Remove any declaration block
             $cssdef = trim($cssdef);
-            my @sp = split( /[,\+ ]/, $cssdef );    # Split selectors like "class1, class2", "p+p", ".myclass p"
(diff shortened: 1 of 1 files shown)


cpeel (approved):

srjfoo:
> Testing: it should now report `h6.center {text-align:center;}` if there are no `h6` elements in the body.

I used Charlie's test file from issue #1204 and added `h6.center {text-align:center;}`, and it was not reported. I added a space to change it to `h6 .center {text-align:center;}`, and it was reported. I changed it back to the original, and added bare `h4 and h5` to the bare element definitions, and `h6` was not reported, but `h4` and `h5` in addition to the `h3` were reported.

windymilla:
Thanks @srjfoo 
I'm glad I chose that example as one that would work! And that you tested that it did (not)!
New commit adds period in as a character to use when splitting compound CSS selectors, along with space, comma and plus.

srjfoo (approved):

Issue #1205, opened 2023-05-30
Explainable behaviour in 1.6-test3 but is it a bug?


G4OEU:
**Describe the bug**
A paragraph spans four 'pages'. There are three illos within the paragraph. Two are on the second page and the third illo is on the third page. The page markers are still present. 

Ctrl+Shift+click is used to move the first illo to the top of the para. It does so correctly by moving that illo to the first blank line it finds that is above the location of the illo. That is found at the start of the para on the page above.

Ctrl+Shift+click is used again to move the second illo to the top of the para. It also operates correctly and places that illo immediately below the first one moved. 

However when Ctrl+Shift+click is used to move the third illo (which is on the third 'page' of the para text) it of course looks for the first blank line above it and finds that blank line where the first two illos were moved from; that is on page two of the para text.

That wasn't what I was naively expecting to happen which was to have all three illos, immediately at the head of the para. However I can see why it has behaved that way.

Now when I Ctrl+Shift+click again on that third illo it finally moves to its correct position under the other two. But it has required an extra Ctrl+Shift+click to achieve the desired end.

**To Reproduce**
1. Open Guiguts with
[illustration-fixup-test.txt](https://github.com/DistributedProofreaders/guiguts/files/11602130/illustration-fixup-test.txt)
2. Open the Tools>Illustration Fixup menu.
3. Scroll down to line 697. The illo is labeled 'Fig. 5'. Ctrl+Shift+click to move it to the top of the para it is in.
4. Scroll down to line 699. The illo is labeled 'Fig. 6'. Ctrl+Shift+click to move it to the top of the para it is in.
5. Scroll down to line 719. The illo is labeled 'Fig. 7'. Ctrl+Shift+click to move it to the top of the para it is in.
6. See where it is moved to which is not the top of the para beneath Fig. 5 and Fig. 6.
7. Ctrl+Shift+click again to move it to the top of the para beneath the other two illos.

**Expected behavior**
If the assumption is always that the illos being moved are within a paragraph then when an illo is moved the blank line following the illo should deleted.

**Environment**
* Guiguts version: 1.6-test3
* OS and version: Windows 10
* Perl version: 
* Perl distribution: 

**Screenshots**

**Additional context**

windymilla:
[...]

PR #1208, opened 2023-06-03
Remove unwanted blank line after illo move

windymilla:
After using Illustration Fixup, if the illo is flagged as mid-paragraph, then when it is moved, the blank line after it (if present) needs to be deleted, or would leave behind a paragraph break.


windymilla:
This fixes #1205

cpeel (approved):

srjfoo (approved):

Issue #1207, opened 2023-06-02
Allow "Display Character Names" to be set as a default


okrick:
I am writing to request that you allow users to set the "Display Character Names" preference as a default preference for the Toolbar. I believe this could be accomplished by making the setting sticky so that the program remembers the last setting used even after quitting and reopening the program.

This feature would be valuable for users who frequently use the "Display Character Names" preference. It would save them time and effort by not having to re-enable the selection every time they start the program.

I appreciate your consideration.

PR #1209, opened 2023-06-03
Make `long ordinal label` setting persistent

windymilla:
User can click on hex/dec ordinal label in status bar, or use Prefs, Toolbar to Display Character Names in ordinal label. This commit makes that setting persistent instead of resetting to "off" when GG is restarted.

Minor tidy up of unnecessary double toggling - clicking the menu option toggles the variable between the default values of 0 and 1.


windymilla:
Fixes #1207

srjfoo (approved):

cpeel (approved):

PR #1211, opened 2023-06-08
Override "busy" behavior for main text window

windymilla:
All Tk widgets ring the bell if they receive keystrokes or mouse clicks while they are busy. However, if the focus changes to the main text window while the Shift or Control modifier keys are held down, a "fake" keypress of Shift/Control appears to be generated, causing the bell to ring.

Override the `_busy` routine to omit the bell if the key stroke is Shift or Control.

### DIFF
--- a/src/lib/Guiguts/TextUnicode.pm
+++ b/src/lib/Guiguts/TextUnicode.pm
@@ -871,4 +871,19 @@ sub redo {
     ::errorcheckilloupdateneeded();
 }
 
+#
+# Overridden from Widget class
+# When focus returns to text window while Shift/Control still held down
+# having been pressed when focus was in a different window
+# (as can happen after Shift/Control click in ErrorCheck dialog)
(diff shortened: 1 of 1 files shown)


windymilla:
Testing (only fails sometimes): Open 
[illustration-fixup-test.txt](https://github.com/DistributedProofreaders/guiguts/files/11687060/illustration-fixup-test.txt) and run Illustration Fixup. 
Then use Ctrl-Shift-click on figures 8, 9 and 10 a bit. The only time the bell should ring (and the status bar flash in the bottom left) is if you try to move illo 10 to a position before illo 9, for example. It should not ring when an illo moves successfully.

cpeel (approved):

srjfoo (approved):
I wasn't seeing the problem on macOS, and after further testing and discussion with @windymilla, I'll accept that this fixes the problem on Windows.

PR #1214, opened 2023-06-11
Implement Sidenote Fixup

windymilla:
Sidenote Fixup is essentially identical to Illustration Fixup, but it looks for `[Sidenote` instead of `[Illustration`

90% of this commit is just renaming variables and functions, e.g. `errorcheckilloupdateneeded` to `errorcheckillosnupdatedneeded` as well as the comments that previously only mentioned illos.

The old Sidenote Fixup and the new one have been moved into a submenu, and renamed to be Automatic (old) and Manual (new) versions. The function of the old, automatic fixup is unchanged, and is still potentially useful to PPers.

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -7,7 +7,7 @@ BEGIN {
     our ( @ISA, @EXPORT );
     @ISA    = qw(Exporter);
     @EXPORT = qw(&errorcheckpop_up &spellquerycleardict &spellqueryinitialize &spellquerywfwordok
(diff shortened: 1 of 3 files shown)


windymilla:
This [normal-sidenotes.txt](https://github.com/DistributedProofreaders/guiguts/files/11715156/normal-sidenotes.txt) file has some sidenotes that can be moved up and down.

Note that this PR does not currently contain the as-yet-unmerged fixes to the two bell-ringing bugs #1211 and #1212

srjfoo (approved):

cpeel (approved):

PR #1221, opened 2023-06-24
Handle page break markers when indenting blocks

windymilla:
Code to determine how much to indent/unindent blocks marked with `/*` did not take note of `x7f` used as a placeholder for page markers. Hence if a block had a page break marker at the start of a line,  it miscalculated the minimum number of leading spaces in the block, and also did not delete spaces correctly if unindenting was required.


windymilla:
Testing notes: In the attached file, there is a block beginning at line 10480, which, instead of being correctly indented to the number of spaces, always had spaces added. Experiment with different values for "Default indent for /* blocks" which can be set using Prefs, Processing, Set Rewrap Margins
[mirror-common.txt.zip](https://github.com/DistributedProofreaders/guiguts/files/11857512/mirror-common.txt.zip)

cpeel (approved):

srjfoo (approved):

PR #1223, opened 2023-06-26
Stop page markers moving in ML Auto-lists

### LINKED ISSUE #1219: HTML List conversion moves page markers
See https://www.pgdp.net/phpBB3/viewtopic.php?p=1312398#p1312398

windymilla:
If a multi-line list is converted to HTML using HTML Markup, Auto List, with the ML flag checked, it deleted all the text and reinserted a version with `<li>` markup. Unfortunately this had the effect of moving all the page markers in the list to the top.

Recoded to work line-by-line, like the non-ML case.

Fixes #1219

### DIFF
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -3297,28 +3297,27 @@ sub autolist {
 
     $textwindow->addGlobStart;
 
-    # Multiline means list entries are separated by double newlines
+    # Multiline means list entries are separated by blank lines
     if ( $::lglobal{list_multiline} ) {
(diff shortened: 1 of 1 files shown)


cpeel (approved):

windymilla:
The attached file has a full-size and a small version of a file that shows the problem.
Right-click the "Img" status bar field to show page marker positions.
Use HTML->HTML Markup->Auto List (ensuring the "ML" flag is set) to convert a selected section to an HTML list.
You can convert the whole "small" file, or lines 847 through 36251 in the full-size file.
[dicamauth_prelist.html.zip](https://github.com/DistributedProofreaders/guiguts/files/11995102/dicamauth_prelist.html.zip)

srjfoo (approved):

PR #1224, opened 2023-06-29
Avoid bug when page marker is on closing markup

windymilla:
If there is a blank line or other closing markup immediately before closing blockquote markup `#/`, the wrapping code has a special check to avoid accidentally indenting the closing markup itself.

However, in the previous work (#1111) to ensure the markup was detected, even if the line was polluted with temporary page markers, the above case was overlooked. Hence, if there was a page marker on the closing markup line, the markup was indented during rewrap.

Using the `notpmselection` variable instead fixes the issue as expected.

### DIFF
--- a/src/lib/Guiguts/SelectionMenu.pm
+++ b/src/lib/Guiguts/SelectionMenu.pm
@@ -439,8 +439,8 @@ sub selectrewrap {
                     $inblock      = 0;
                 }
             } else {
-                my $skipit = ( $selection =~ /^(\/#|#\/)$/ );    # Happens when /c or /r directly follows /# or blank line before #/
-                if ( $::blockwrap > 0 ) {                        # block wrap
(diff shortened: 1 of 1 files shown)


windymilla:
Testing note: [pagemarkers.zip](https://github.com/DistributedProofreaders/guiguts/files/11910234/pagemarkers.zip) has a longer and shorter file which can be rewrapped with Ctrl-a Ctrl-w. Both of these indent the final closing markup when wrapped in `master`. File, Project, Display/Adjust Page Markers (or right-click Img label in status bar) can be used to show where the page markers are prior to rewrap, and can also be adjusted up/down to either move them away from the close markup, or to move them to other positions within the markup.

cpeel (approved):

srjfoo (approved):

PR #1226, opened 2023-07-11
Generate valid code for italic superscripts

windymilla:
Since the "rule" is that single character superscripts do not require enclosing braces, that is what Guiguts did.

However, where a single character superscript is italicised, the rule is a little ambiguous and can therefore lead to markup such as `a^<i>x</i>` instead of
`a^{<i>x</i>}`
Handle this case by allowing a single character which is surrounded by "markup" to be superscripted without the enclosing braces. The chances of the reverse case of a book containing a single superscripted angle bracket immediately followed by apparently valid markup around a single character are vanishing small, so this seems a safe edit.

### DIFF
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -59,6 +59,11 @@ sub html_convert_superscripts {
         $textwindow->ntinsert( "$step.0", $selection );
     }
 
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: [def.txt](https://github.com/DistributedProofreaders/guiguts/files/12015102/def.txt) might be useful for testing. Use Autogenerate to convert it to HTML.

cpeel (approved):

srjfoo (approved):

PR #1229, opened 2023-07-23
Permit svg images in ppvimage check

windymilla:
This is mainly to avoid error messages when checking HTML files using the HTML+SVG method for LaTeX formula inclusion. The width/height check isn't that important since the tools don't output the `img` code with a width/height that ppvimage checks. However, for compatibility with other image types I've added rudimentary width/height extraction code.

### DIFF
--- a/tools/ppvimage/ppvimage.pl
+++ b/tools/ppvimage/ppvimage.pl
@@ -248,6 +248,23 @@ sub runProgram {
         return ( 0, 0 );
     }
 
+    # svgsize : gets the width and height (in pixels) of an svc file
+    sub svgsize {
+        my ($SVG) = @_;
+        my $buf = '';
+
+        if ( defined($SVG)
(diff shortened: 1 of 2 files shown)


tools/ppvimage/ppvimage.pl
```diff
@@ -248,6 +248,23 @@ sub runProgram {
         return ( 0, 0 );
     }
 
+    # svgsize : gets the width and height (in pixels) of an svc file
```
cpeel, line 251: Minor: should that comment end with `svg file` instead?

cpeel (approved):

tools/ppvimage/ppvimage.pl
windymilla: Oops - thanks

srjfoo (approved):

PR #1232, opened 2023-07-29
Fix Mac FileOpen dialog error when loading TIA OCR

windymilla:
Previous work to fix the error only applied to opening the main text and HTML files, and some importing, but hadn't been applied to TIA OCR loading.


cpeel (approved):

srjfoo (approved):

srjfoo (approved):
Changes look good!

PR #1236, opened 2023-08-10
Really final fix for Mac nonexistent dir bug

windymilla:
Searching for `initialdir` found all places that construct was used, which on Mac causes error if dir doesn't exist.
Replaced with default safe last path if dir doesn't exist.

Embarrassingly the third(?) time of fixing the last occurrence :)

### DIFF
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -2228,6 +2228,7 @@ sub htmlautoconvert {
 sub thumbnailbrowse {
     my $name = shift;
     unless ($name) {
+        $::globalimagepath = ::getsafelastpath() unless ( -d $::globalimagepath );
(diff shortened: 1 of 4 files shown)


windymilla:
Testing notes: 4 places fixed

1. Choose png folder by right-clicking status bar "Img" icon (when pngs subfolder of project folder doesn't exist)
2. Loading an image from the HTML Markup dialog (when previous folder used doesn't exist)
3. Using Tools, Stealth Scannos (when scannos folder doesn't exist)
4. Using Locate Scannos from Prefs, File Paths, Set File Paths (when parent of scannos folder ceases to exist)

All of these probably give an error on the Mac without the fix. With the fix, they should open either the GG top level folder (where `guiguts.pl` is) or the most recently used dir, quite likely the project folder.

cpeel (approved):
Third time's the charm!

srjfoo (approved):

PR #1237, opened 2023-08-18
Don't pop Page Marker Adjust when using next page

windymilla:
When using the next/previous page icons on the status bar (`<` and `>`) it used to pop the Page Marker Adjust dialog, and turn off Auto Img if that was enabled. It would then move to the previous/next page marker and show the image, although it left the cursor just before the page marker, so in the status bar said you were on the page before the one that was shown.

Now more consistent. The two buttons move to the previous or next page marker as before, but only automatically show the image if Auto Img is enabled; using the buttons does not turn Auto Img off; the cursor is positioned 1 character into the correct page, i.e. the earliest point that counts as being on that page; it also briefly shows the yellow page marker labels, but doesn't pop the Marker Adjustment dialog.

Unused argument also removed from `togglepagenums()`

### DIFF
--- a/src/lib/Guiguts/PageNumbers.pm
+++ b/src/lib/Guiguts/PageNumbers.pm
@@ -17,15 +17,15 @@ sub hidepagenums {
 }
 
 #
-# Show visible poage markers
+# Show visible page markers
 sub displaypagenums {
     ::togglepagenums() unless $::lglobal{seepagenums};
 }
 
-## Toggle visible page markers
+## Toggle visible page markers - optional argument to leave dialog visibility unchanged
 sub togglepagenums {
-    my $showallmarkers = shift;
-    my $textwindow     = $::textwindow;
+    my $nodialog   = shift;
+    my $textwindow = $::textwindow;
     if ( $::lglobal{seepagenums} ) {
         $::lglobal{seepagenums} = 0;
         my @marks = $textwindow->markNames;
@@ -33,13 +33,10 @@ sub togglepagenums {
             if ( $_ =~ m{Pg(\S+)} ) {
                 my $pagenum = " Pg$1 ";
                 $textwindow->ntdelete( $_, "$_ +@{[length $pagenum]}c" );
-            } elsif ($showallmarkers) {
-                my $pagenum = " @_";
-                $textwindow->ntdelete( $_, "$_ +@{[length $pagenum]}c" );
             }
         }
         $textwindow->tagRemove( 'pagenum', '1.0', 'end' );
-        ::killpopup('pagemarkerpop');
+        ::killpopup('pagemarkerpop') unless $nodialog;
     } else {
         $::lglobal{seepagenums} = 1;
         my @marks = $textwindow->markNames;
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes: There are two ways to move forward/back a page - either via the `<` and `>` status bar buttons, or via `Next/Previous Marker` buttons in Page Marker Dialog (accessible by right-clicking `Img: 001` icon in status bar.
These should behave similarly, showing the image if and only if Auto Img (also in status bar) is enabled.
After clicking one of these, the image number reported in the `Img: 001` icon should match the Page Marker you jumped to (unlike previously) and hence the same image as will be shown if you click See Img, or have Auto Img enabled.

srjfoo:
> Testing notes: There are two ways to move forward/back a page - either via the `<` and `>` status bar buttons, or via `Next/Previous Marker` buttons in Page Marker Dialog (accessible by right-clicking `Img: 001` icon in status bar. These should behave similarly, showing the image if and only if Auto Img (also in status bar) is enabled. After clicking one of these, the image number reported in the `Img: 001` icon should match the Page Marker you jumped to (unlike previously) and hence the same image as will be shown if you click See Img, or have Auto Img enabled.

It works as described, but I wanted to clarify something. Given the fact that the `<` and the `>` are on either side of the `See Img` button, I don't actually see a problem with the image being shown, also. In fact, given the location of those buttons, I think that people would expect to see the image (as well as having the cursor move to the next page in the text).

Given the current behavior, would it make more sense for the `<` and `>` to be on each side of the `Img: ###` button. The expected behavior of that button if just clicking is to bring up a dialog popup to navigate to a specific page in the text.

Caveat: Since I am not actively PP'ing, I don't know which behavior would be more useful:
* Have the `<` and `>` navigate just in the text file unless `Auto Img` is turned on, or
* Always show the image in question, as well as navigating to the correct location in the text file.

windymilla:
> Given the current behavior, would it make more sense for the `<` and `>` to be on each side of the `Img: ###` button. The expected behavior of that button if just clicking is to bring up a dialog popup to navigate to a specific page in the text.

Yes, that does make more sense - second commit added - thanks

> Caveat: Since I am not actively PP'ing, I don't know which behavior would be more useful:
> 
> * Have the `<` and `>` navigate just in the text file unless `Auto Img` is turned on, or
> * Always show the image in question, as well as navigating to the correct location in the text file.

I personally rarely use `Auto Img`, and prefer to explicitly `See Img` when needed, but others do like it. Also, on some platforms the focus gets shifted when the image is shown, so it may be annoying to them. I think the flexibility of the user choosing whether to have the image shown when they use Prev/Next Img (via `Auto Img`) is therefore best.

windymilla:
After further discussion with users, decided to revert the button move, and save any redesign of the status bar for when there is more to do, and additional feedback from a range of users has been gathered.

srjfoo (approved):

cpeel (approved):
LGTM!

Since you reverted a commit in the same PR, consider doing a "Squash and Merge" on the PR and editing the message to be just the message from your first commit. This will prevent the second and third commits from going in at all.

windymilla:
> Since you reverted a commit in the same PR, consider doing a "Squash and Merge" on the PR and editing the message to be just the message from your first commit. This will prevent the second and third commits from going in at all.

Thanks for the tip @cpeel

PR #1239, opened 2023-08-25
Clarify which count dialog is which

windymilla:
Show in the title of the two count dialogs which one is which. The one invoked from the main window on the selected text does a "whole word" search.
The other is from the "Quick Search" (QS) dialog.

### DIFF
--- a/src/lib/Guiguts/SearchReplaceMenu.pm
+++ b/src/lib/Guiguts/SearchReplaceMenu.pm
@@ -2219,7 +2219,7 @@ sub quickcount {
 
     # Inform user
     my $dlg = $::top->Dialog(
-        -text    => "Count: $count        ",    # Extra space needed to force dialog wide enough for title
+        -text    => "Word Count: $count",
         -bitmap  => "info",
         -title   => "Word Count",
         -buttons => ['Ok']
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: `Ctrl-Shift-b` when you have text highlighted in the main window should do a whole word search for the text. When in the Quick Search dialog (bring up with `Ctrl-Shift-f`) the same keystroke should to a QS Count, using whatever is set up in the QS dialog, e.g. regex, case, etc. The only change should be the title of the two count dialogs are now different to one another.

cpeel (approved):

src/lib/Guiguts/SearchReplaceMenu.pm
```diff
@@ -2219,9 +2219,9 @@ sub quickcount {
 
     # Inform user
     my $dlg = $::top->Dialog(
-        -text    => "Count: $count",
+        -text    => "Count: $count        ",    # Extra space needed to force dialog wide enough for title
```
srjfoo, line 2222: Just a comment, and not a deal-breaker at all, but that extra space did not force the window wider on macOS. I has to manually widen the window.

srjfoo (approved):

src/lib/Guiguts/SearchReplaceMenu.pm
windymilla: New commit adjusts wording so it's obvious it's a word count, even on macOS where the title is truncated - confirmed OK on Mac by @srjfoo

PR #1242, opened 2023-08-28
Distinguish between classes and elements in pphtml

windymilla:
When checking for unused/undefined CSS, classes and elements were stored in the same array, which was OK for most things. However, if a classname was also an element name (e.g. `br`) it would be incorrectly reported or ignored.

This commit prepends a period to stored class names to match the class name in the CSS. It also treats anything like `<string` as a potential element name. This is safer than trying to maintain a list of all valid element names, and if the user has an invalid element name, the validator will tell them, so pphtml doesn't need to do that.

### DIFF
--- a/src/lib/ppvchecks/pphtml.pl
+++ b/src/lib/ppvchecks/pphtml.pl
@@ -16,10 +16,10 @@ my @cssline      = ();
 my %classes_used = ();
 my %classes_line = ();
 
-my $help      = 0;                  # set true if help requested
-my $srctext   = "book.html";        # default source file
-my $outfile   = "xxx";
-my $validchar = "[-_[:alnum:]]";    # Valid characters for classnames, etc.
+my $help         = 0;                              # set true if help requested
+my $srctext      = "book.html";                    # default source file
+my $outfile      = "xxx";
+my $validpattern = "[[:alpha:]][-_[:alnum:]]*";    # To check valid classnames, etc.
 
 my $filename;
 my $frm_detail;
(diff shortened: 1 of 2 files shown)


srjfoo (approved):

src/lib/ppvchecks/pphtml.pl
```diff
@@ -16,22 +16,15 @@
 my %classes_used = ();
 my %classes_line = ();
 
-my $help    = 0;              # set true if help requested
-my $srctext = "book.html";    # default source file
-my $outfile = "xxx";
+my $help      = 0;                  # set true if help requested
+my $srctext   = "book.html";        # default source file
+my $outfile   = "xxx";
+my $validchar = "[-_[:alnum:]]";    # Valid characters for classnames, etc.
```
cpeel, line 22: Not sure how precise you care to be here, but valid elements must start with an alpha character, then 0 or more `[:alnum:]`, `-`, or `_`:
```
[:alpha:][-_[:alnum:]]*
``` 
rather than
```
[-_[:alnum:]]+
```

cpeel (approved):
Small suggestion on the regex, although it's less a `validchar` in that case and more a `validpattern` that would impact how it's used elsewhere.

src/lib/ppvchecks/pphtml.pl
windymilla: I wasn't too worried, as it's not really checking valid HTML - just the class use. However, it's probably better to do as you suggest, so it doesn't confuse PPers or future developers, so I'll rustle up another commit - thanks!

windymilla: @cpeel - thanks - I've done that now. At least in Perl regex, `[:alpha:]` can only be used inside a character class, so the regex needed to be `[[:alpha:]][-_[:alnum:]]*`

cpeel (approved):

PR #1248, opened 2023-09-09
Add history to Highlight Char/Regex entry field

windymilla:
1. Use generic routines to add history pulldown (thanks past windy :)
2. Save highlight history to `setting.rc`
3. Add `savesettings` call to generic routine when history is altered (boo! past windy :)
3. Set focus to entry field when dialog is popped to improve UX

### DIFF
--- a/src/guiguts.pl
+++ b/src/guiguts.pl
@@ -236,6 +236,7 @@ our @recentfile;
 our @replace_history;
 our @search_history;
 our @quicksearch_history;
+our @highlight_history;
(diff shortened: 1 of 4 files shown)


cpeel (approved):
> Use generic routines to add history pulldown (thanks past windy :)

> Add savesettings call to generic routine when history is altered (boo! past windy :)

Ah yes, the ever-present "past-me was brilliant!" followed closely by "what was past-me thinking!?" that we all know too well 🤣❤️

srjfoo (approved):

PR #1250, opened 2023-09-11
Ignore `@media` when open brace is on next line

windymilla:
Previously, the code only stripped `@media` and any media types when there was a brace on the same line as the `@media`. If brace was on next line, then `media`, `print`, `screen`, etc. would be reported as "unused" classes.
Now, make the brace optional - strip it if it's there. If it's on the next line, it will be dealt with (ignored) there.


windymilla:
Testing notes:

A file that has 
```
@media print
{
.abc {width:100%;}
}
```
should not report `media` or `print` as classes.

Previously required it to be
```
@media print {
.abc {width:100%;}
}
```

srjfoo (approved):

PR #1261, opened 2023-10-02
Add good_words to project dict more efficiently

### LINKED ISSUE #1255: Large good_words files are very slow to load
By large, we're talking 5000+ words.

windymilla:
Previously it saved the dictionary after reading each word from the `good_words.txt`. Now it reads them all, then does a single save at the end.

Fixes #1255

### DIFF
--- a/src/lib/Guiguts/SpellCheck.pm
+++ b/src/lib/Guiguts/SpellCheck.pm
@@ -226,18 +226,16 @@ sub spelladdword {
 }
 
 #
-# Add a word to the project dictionary
-# Optional second argument if it's a bad word
+# Add a single word to the project dictionary and save it - slow if used for bulk additions
 sub spellmyaddword {
     my $textwindow = $::textwindow;
     my $term       = shift;
-    my $bad        = shift;
     unless ($term) {
         ::soundbell();
         return;
     }
     return if $term =~ /^\s*$/;
-    ( $bad ? $::projectbadwords{$term} : $::projectdict{$term} ) = '';
+    $::projectdict{$term} = '';
     spellsaveprojdict();
 }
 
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: You can use almost any text file as a `good_words.txt` file (5000 lines long is about a 10 second wait on my computer in `master`). Once you have created your good words file, open any other file (e.g. `mytextfile.txt`) in GG, run Tools->Spell Query, and in the dialog, use `Add good/bad words`. You can also create a `bad_words.txt` and load those via the same method. Each line in a good/bad words file is (always has been) treated as a good/bad "word". After running `Add good/bad words` you can edit `mytextfile.dic` and you should see all the good words listed (as Perl code), then below all the bad words.

cpeel (approved):

srjfoo (approved):

src/lib/Guiguts/SpellCheck.pm
```diff
@@ -226,18 +226,16 @@ sub spelladdword {
 }
 
 #
-# Add a word to the project dictionary
-# Optional second argument if it's a bad word
+# Add a single word to the project dictionary and save it - slow if used for bulk additions
```
srjfoo, line 229: This branch is undoubtedly faster than `master`, but this comment seems to imply that it's still saving after each addition? Or am I misinterpreting something?

windymilla: That function `spelladsword` is still used for adding a single word to the project dictionary - I added the word "single" to the comment, and that it saves the dictionary, to emphasise that. However, it used to be called multiple times when a good word list was added - I now add all the words to the internal structures, then save the dictionary once for that case, so don't use this function at all for that. The comment that it is "slow if used for bulk additions" is intended to alert a future developer not to use it for adding lots of words.

PR #1262, opened 2023-10-10
Various LaTeX-->HTML/SVG support features

windymilla:
Tools support conversion of existing LaTeX file (one taken from PG for reprocessing & reposting as an HTML/SVG project) or a project that has been through the rounds with HTML/SVG as the target format.
1. Add `LaTeX Tools` submenu to the File menu
2. Convert LaTeX page separators (comment format used commonly by PG submitters) to Guiguts format to link pngs (if available) to the correct page, and also to allow setting of printed page numbers for HTML format.
3. Convert LaTeX-style curly quote markup to standard curly quotes
4. Preview selected piece of LaTeX from an HTML/SVG file (for use while checking/adjusting the chunks of LaTeX in the HTML file as it comes out of the rounds)
5. Preview whole file by converting HTML/SVG file using `m2svg` tool
6. HTML autogen is a bit overenthusiastic, e.g. changing ampersands to HTML entity codes everywhere, which would break LaTeX code that contains ampersands. Add option to undo such changes.

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -7,7 +7,7 @@ BEGIN {
     our ( @ISA, @EXPORT );
     @ISA    = qw(Exporter);
     @EXPORT = qw(&errorcheckpop_up &spellquerycleardict &spellqueryinitialize &spellquerywfwordok
-      errorcheckillosnupdateneeded);
+      &errorcheckillosnupdateneeded &savetoerrortmpfile);
 }
 
 my @errorchecklines;
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -310,6 +310,7 @@ sub html_convert_body {
     my $blkrightmargin = 0;
     my $blkrstart      = 0;               # Value of $step when a block right starts
     my @blkrlens       = ();
+    my $blklatex       = 0;
     my $closure        = voidclosure();
 
     ::hidelinenumbers();                  # To speed updating of text window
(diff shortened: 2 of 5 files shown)


windymilla:
DP user has been requesting/trialling these features in the work they have been doing. Regular PPers will not use them at all. I thought it would be good to get them into the system, rather than hanging around in a test version.
TBH, I'm not too worried about them being tested independently, and I think it would be hard for someone to test them who was not doing this type of work.
If problems do arise in the future relating to these tools, they can be fixed as needed.

cpeel (approved):
Looks like `m2svg` is a nodejs package and I'm assuming we want to leave (the optional) installation of that up to the user rather than try to package that with all its dependencies. Do we need to include anything about that in `INSTALL.md`?

windymilla:
Thanks, Casey - I don't think we need to mention it in INSTALL.md at this point (since it probably affects either 1 or 2 users, both of whom already have m2svg installed of course) but it's something to bear in mind if HTML/SVG gets a sudden increase in popularity amongst PMs & PPers.

srjfoo (approved):
> 6\. HTML autogen is a bit overenthusiastic, e.g. changing ampersands to HTML entity codes everywhere, which would break LaTeX code that contains ampersands. Add option to undo such changes.

I wondered about this one when I looked at the menu. Just confirming that it's non-selective; all such changes are reverted by using this button? I think it was the "some" that confused me, but I can't think of a better word.

Just a note that I looked over the code, but since you feel like it's been adequately tested, let's go with it. Let me know if you do need me to do testing on macOS, but if so, I'll probably need help in the testing.

PR #1263, opened 2023-10-12
Add new troubleshooting file w/ macOS details

cpeel:
While testing the python GG proto I needed to update some wildly out-of-date homebrew stuff and thought it worthwhile to add it to the GG docs. This is a lot of "kill it with fire and start over" approach when GG fails to start after a macOS upgrade.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -193,13 +193,13 @@ it before starting this process.
 
 ### XQuartz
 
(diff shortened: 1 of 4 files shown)


srjfoo (approved):

windymilla (approved):

windymilla:
Thanks

PR #1267, opened 2023-10-17
Don't assume good/badwords files are utf8-encoded

windymilla:
Apparently, they can be "Latin-1" - previous work assumed utf8.

### DIFF
--- a/src/lib/Guiguts/SpellCheck.pm
+++ b/src/lib/Guiguts/SpellCheck.pm
@@ -494,10 +494,11 @@ sub spelladdgoodwords {
 
     my $fh;
 
-    # Load good words first
-    if ( open( $fh, "<:encoding(utf8)", "good_words.txt" ) ) {
+    # Load good words first - these files may not be utf8-encoded, so don't assume they are
+    if ( open( $fh, "<", "good_words.txt" ) ) {
         ::busy();
         while ( my $line = <$fh> ) {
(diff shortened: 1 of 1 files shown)


cpeel (commented):
When we rolled out Unicode we converted all good & bad word files to UTF-8 and that's how Apache serves them up as well.

I'd really like to know where these non-UTF-8 files are coming from, because there's something wrong somewhere and it isn't with Guiguts.

windymilla:
Sharon just explained to me that the goodwords were recreated, but the zip that the PPer downloads were not. This zip dates from 2018. I checked the goodwords file itself directly downloaded from the project page (last modified May 2020, presumably when it was converted to utf8) and it is utf8. So, it's just the old zips that could contain non-utf8 goodwords files. GG used to cope with that before my recent assumptive "improvements".

cpeel (approved):
Got it - thank you for that clarification!

windymilla:
It's thanks to @srjfoo of course, very much on the ball! Thanks too, Casey, for checking up the explanation.

srjfoo:
Looking at the project history, the project was checked out at the time. It was just recently returned to the pool, whereupon Charlie checked it out. I don't remember for sure, but I don't think we regenerated files for projects that were checked out for PP, did we?

srjfoo (approved):

PR #1274, opened 2023-11-07
Position coincident page marks in the correct order

### LINKED ISSUE #1272: The HTML Generator is omitting some pagenums
The HTML Generator omitted several <span class="pagenum"....>'s.  All of the page numbers seem to be present in the .bin of the text file the Generator used, and I think all of them followed "no count" pages originally containing mid-paragraph illustrations.

To recreate the problem, open "h01-pregen.html" (actually a .txt file, of course) in GG and run the HTML Generator.

Before doing so, you may want to verify that the soon-to-be-missing page numbers are present in the status bar for "h01-pregen.html" and, of course, in its .bin.

Here is a list of the omitted page numbers.  The list was created by running "HTML Link Check" after running the HTML Generator: 

```
Beginning check: Link Check
+#Page_105: Internal link without anchor
+#Page_11: Internal link without anchor
+#Page_123: Internal link without anchor
+#Page_127: Internal link without anchor
+#Page_129: Internal link without anchor
+#Page_135: Internal link without anchor
+#Page_159: Internal link without anchor
+#Page_181: Internal link without anchor
+#Page_193: Internal link without anchor
+#Page_213: Internal link without anchor
+#Page_243: Internal link without anchor
+#Page_247: Internal link without anchor
+#Page_251: Internal link without anchor
+#Page_257: Internal link without anchor
+#Page_263: Internal link without anchor
+#Page_265: Internal link without anchor
+#Page_39: Internal link without anchor
+#Page_59: Internal link without anchor
+#Page_63: Internal link without anchor
+#Page_89: Internal link without anchor
Check is complete: Link Check
```

The project is "[The Macedonian Campaign](https://www.pgdp.net/c/project.php?id=projectID62504c8ca62f4)."

This happens in versions going back to 1.4.0; I didn't check further back than that. The input file to the regression tests was this same "h01-pregen.html," so issues in the .bin likely affected all versions of GG that used it.

Looking in the .bin, the missing page numbers have the same offsets as the two preceding "no count" pages. I believe that, in all of them, the illustrations on those "no count" pages were mid-paragraph, and moved (usually up, towards the beginning of the file) to paragraph breaks.

Looking at the sequence of early steps in preparing the .txt file, I moved the illustrations (using "Illustration Fixup" before running "Configure Page Labels." Could that be the culprit? Should "Configure Page Labels" be run before "Illustration Fixup"?

[autogen-pagenum-bug.zip](https://github.com/DistributedProofreaders/guiguts/files/13200475/autogen-pagenum-bug.zip)

windymilla:
If Pg123 and Pg124 were at the same location in the file, usually because Pg123 is blank, then when the page locations are loaded from the bin file and positioned in the text widget using "marks", the mark for Pg124 was inserted at the location where Pg123's mark was, and ended up before it in the text widget's internal list of marks.

Loading the marks in reverse order means that coincident marks appear in alphanumeric order,  which is preferable.

This was assumed to be the case by the HTML generation code when set to skip coincident pagenum spans. So this fixes #1272

### DIFF
--- a/src/lib/Guiguts/HTMLConvert.pm
+++ b/src/lib/Guiguts/HTMLConvert.pm
@@ -1245,20 +1245,19 @@ sub html_convert_sidenotes {
 sub html_convert_pageanchors {
     my $textwindow = $::textwindow;
     ::working("Inserting Page Number Markup");
-    $|++;
     my @pagerefs;    # keep track of first/last page markers at the same position
-    my $tempcounter;
-    my $mark = '1.0';
-    while ( $textwindow->markPrevious($mark) ) {
-        $mark = $textwindow->markPrevious($mark);
-    }
 
     my $closure = voidclosure();
 
-    # Work through all the text markers
-    while ( $mark = $textwindow->markNext($mark) ) {
-        next unless $mark =~ m{Pg(\S+)};              # Only look at page markers
-        my $markindex = $textwindow->index($mark);    # Get page marker's index
+    # Get a correctly sorted list of page markers - if two markers are at the
+    # same location, the order returned by markNext is not defined, but we need
+    # them ordered by png.
+    my @pagemarklist = sort grep ( /^Pg\S+$/, $textwindow->markNames );
+
+    for my $idx ( 0 .. $#pagemarklist ) {
+        my $mark      = $pagemarklist[$idx];
+        my $markindex = $textwindow->index($mark);                                # Get page marker's index
+        my $marknext  = $idx < $#pagemarklist ? $pagemarklist[ $idx + 1 ] : "";
 
         # This is the custom page label
         my $num = $::pagenumbers{$mark}{label};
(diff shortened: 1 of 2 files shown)


cpeel (approved):

windymilla:
Apologies - although the previous commit appeared to fix the problem, it only fixed some coincident page markers and broke others. The order of coincident marks does not appear to be well defined in Tk, regardless of the order they were created or their "gravity" setting. This commit just gets all the marks and sorts them by png name, so the list is guaranteed to be in page order.

src/lib/Guiguts/HTMLConvert.pm
```diff
@@ -1245,20 +1245,19 @@ sub html_convert_sidenotes {
 sub html_convert_pageanchors {
     my $textwindow = $::textwindow;
     ::working("Inserting Page Number Markup");
-    $|++;
```
cpeel, line 1248: What does this even do? (perl, you are weird)

cpeel (approved):

windymilla:
Testing notes for @srjfoo 
Probably [Charlie's test file in the linked issue](https://github.com/DistributedProofreaders/guiguts/files/13200475/autogen-pagenum-bug.zip) is as good as anything. Without this PR, HTML autogenerate would leave you with broken links (check with HTML->Link Checker)

srjfoo (approved):

PR #1275, opened 2023-11-11
Add Index Cross-referencing HTML links

### LINKED ISSUE #1264: Enhance index page linking
Some books have entries "See xyz" in the index, which it would be good to link to the "xyz" entry in the index.

Suggested solution: add id to every top-level index entry just using the first word of the entry as the id. Detect the "See" keyword and add link to the id.

Consider whether it would be better to add this as a subsequent operation to be done, rather than incorporating it into the HTML conversion mechanism.

Also, multi-volume indexes tend to end up linking pages that they shouldn't, e.g. page 100 in the current volume should be linked in the normal way, but page 459 in another volume should either not be linked or should be linked to page 459 in a PG text. This one may be possible/easier/more flexibly done via regexes than a feature.

windymilla:
Algorithm searches for "See..." references in the index and attempts to link them to the relevant index entry. Where it is unable to find a link due to differences in wording, etc., it displays a list of errors which can be clicked in order to be resolved manually.

The Error List has been implemented as a general feature that other operations could easily use.

Fixes #1264

### DIFF
--- a/src/lib/Guiguts/ErrorCheck.pm
+++ b/src/lib/Guiguts/ErrorCheck.pm
@@ -21,10 +21,12 @@ my $ENDMSG = "Check is complete:";
 # Handles Bookloupe, Jeebies, HTML & CSS Validate, Tidy, Link Check,
 # Unmatched Tag/Brackets/Double Quotes/Block Checks
 # pphtml, pptxt, ppvimage, Spell Query, EPUBCheck,
-# Illustration Fixup, Sidenote Fixup and Load External Checkfile.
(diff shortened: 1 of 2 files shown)


windymilla:
Testing notes: [mastersremovedids.zip](https://github.com/DistributedProofreaders/guiguts/files/13326561/mastersremovedids.zip) already has an HTML index with some "See" references, and one that should fail (Russian Peasantry). Highlight/select the index and use HTML->HTML Markup->Index Cross-references

cpeel (approved):

srjfoo (approved):

PR #1277, opened 2023-11-21
Handle coincident page marks better

windymilla:
The `ReplaceSelectionsWith` function handles keeping page marks in approximately the same location when a Search/Replace operation spans a page boundary (by maintaining the relative proportion of text each side of the page boundary).

This did not cope correctly when there were two or more page marks at the same location (i.e. blank pages). The page marks could get reversed, so that 015 came before 014, for example (although still both at the same location).

Avoiding changing the mark gravity (whether the mark goes left or right of text inserted at the mark) fixes this for marks within the S/R region. Temporarily setting "right" gravity for page marks that immediately follow the S/R region keeps those page marks correct.

### DIFF
--- a/src/lib/Guiguts/TextUnicode.pm
+++ b/src/lib/Guiguts/TextUnicode.pm
@@ -278,16 +278,18 @@ sub ReplaceSelectionsWith {
             push( @lengths, $fmlen );
         }
 
-        # Set gravity of page marker after string to "right" to keep it after string
-        # It doesn't matter if it's not immediately after - no harmful effect since restored later
-        my $end_mark = "";    # Assume there isn't one
+        # Find page markers immediately after string so we can set their gravity
+        # to "right" to keep them after string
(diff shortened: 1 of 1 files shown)


windymilla:
Testing notes: [Charlie's example](https://github.com/DistributedProofreaders/guiguts/files/13368791/pagenuml-sequence-bug.zip) in the linked issue shows the problem - you can right-click the Img label to show/hide where the page marks are. Around line 420 the page marks 014 and 015 are in the correct order, but without this PR, HTML->Generator, Autogenerate causes those to be swapped. They should now remain in the correct order.

The initial attempt at fixing this broke S/R across page boundaries. You can test this out by doing regex replaces, like replacing
`</p>\n<hr`
with something the same length, longer, and shorter. The page marks should stay in the correct position proportionally, and of course no text should end up duplicated unexpectedly.

srjfoo (approved):

PR #1280, opened 2023-11-24
Trap AutoIndex bug when no lines to process

windymilla:
Attempting to run HTML -> AutoIndex when there are no non-blank lines before the end of file, e.g. on empty file, caused an infinite loop


windymilla:
Testing notes: Run HTML -> AutoIndex on an empty file, or when cursor is at end of file.

cpeel (approved):

srjfoo (approved):

PR #1281, opened 2023-11-24
Don't skip bad words in Spell Query

windymilla:
The SQ threshold stops words that appear lots of times from being reported as spelling errors. That check shouldn't apply to bad words, which should always be reported.


windymilla:
Testing notes:
Create `good_words.txt` and `bad_words.txt`.
Put any word you like in the bad words file, e.g. `xyz`
Then open a text file in GG and add `xyz` in a few times (e.g. 3 times)
Also add 3 times another misspelling that is not in the bad words list, e.g. `speelcheck`
Run Tools->Spell Query
At the top of the dialog is a Threshold. When that is less than 3, and Spell Query is run, `speelcheck` should not appear, but `xyz` should (because it is bad). When the threshold is 3 or more, both `speelcheck` and `xyz` should appear.

cpeel (approved):

srjfoo (approved):

Issue #1282, opened 2023-11-26
Be more tolerant when adding good/bad words to file


windymilla:
If bad words doesn't exist, there is no error, but if bad words does exist and good words doesn't, there is.

Better to add good and/or bad words, and only give an error if neither are in the project folder.

srjfoo:
Why an error instead of a warning, if neither is there?

windymilla:
I meant a dialog that pops up to warn the user, rather than a printed error, so "warning" would be a better word.

PR #1283, opened 2023-11-28
Only warn user if neither gwl nor bsl is found

windymilla:
Previously used to just complain about gwl, and if there was no gwl, it didn't try to load the bwl.


windymilla:
Fixes #1282

cpeel (approved):

srjfoo (approved):

Issue #1291, opened 2024-01-29
Ubuntu 22, default Perl, library setup


owenBeresford:
As ever with OSS, thanks for your work.

**Describe the bug**
I think this is a DOCS bug.
I haven't done any Perl development on this machine, but it is a dev machine.   The Perl setup will be factory defaults.
`perl -v
This is perl 5, version 34, subversion 0 (v5.34.0) built for x86_64-linux-gnu-thread-multi
(with 60 registered patches, see perl -V for more detail)`

I was following "other install" notes in the github README.md.

Three small issues, but probably config derived
a) the first install script bails
`perl src/install_cpan_modules.pl  \n
Failed trying to install File::HomeDir`                                       << no technical detail

b) File::HomeDir installs without effort via a `cpan` shell.
So I manually try `cpan install Tk` which crashed out, as below
`
The current configuration of allow_installing_outdated_dists is 'ask/no', but for this option we would need 'CPAN::DistnameInfo' installed. Please install 'CPAN::DistnameInfo' as soon as possible. As long as we are not equipped with 'CPAN::DistnameInfo' this option does not take effect
Running make test for ASB/Tk-ToolBar-0.12.tar.gz
PERL_DL_NONLAZY=1 "/usr/bin/perl" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'blib/lib', 'blib/arch')" t/*.t
t/pod.t ... skipped: Test::Pod 1.14 required for testing POD
[...]

PR #1292, opened 2024-01-30
Add installation instructions for Ubuntu

cpeel:
Include instructions for installing on Ubuntu. Specifically to install the perl packages as a non-root user which is the accepted best practice.

This should address https://github.com/DistributedProofreaders/guiguts/issues/1291


srjfoo (approved):

srjfoo:
> Docs look good, and I checked that the edit to `install_cpan_modules.pl` didn't break it on Windows. Thanks.

Thanks for the reminder to check that out on macOS, @windymilla -- also doesn't break on a Mac, and GG came up with no problem.

PR #1296, opened 2024-05-27
Add Mac binding for beg/end of line

windymilla:
Also fix bug in "Select to end of file" shortcut - it did sel to beg.

### DIFF
--- a/src/lib/Guiguts/KeyBindings.pm
+++ b/src/lib/Guiguts/KeyBindings.pm
@@ -239,22 +239,26 @@ sub keybindings {
 
     # Extra bindings for Mac
     if ($::OS_MAC) {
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
`Cmd`+ left and right arrows now properly navigate to the beginning and end of a line, respectively. Yay!

Caveat on the "select from current position to end of file". There is a max buffer size for Macs (the problem is in XQuartz -- I haven't figured out what size it is, or whether it can be changed, and the ability to copy from GG and paste into a different instance of GG, or into a different application is fraught, anyway). "Select from current position to end of file" works now, but trying to copy and paste large amounts of text may have unexpected results. Make sure you save a copy of your file before trying it. I have had the GG window crash trying to copy as few as 300 lines.

PR #1298, opened 2024-06-04
Add page marker flags feature

windymilla:
This feature is in GG2, and allows editing in a different editor without losing the page break positions.

In particular, swapping between GG1 & GG2 should be easy, but any other editor is OK too.

Linked with GG2 PR: https://github.com/DistributedProofreaders/guiguts-py/pull/272

### DIFF
--- a/src/lib/Guiguts/FileMenu.pm
+++ b/src/lib/Guiguts/FileMenu.pm
@@ -11,7 +11,7 @@ BEGIN {
       &clearvars &savefile &_exit &file_mark_pages &file_guess_page_marks
       &oppopupdate &opspop_up &confirmempty &openfile &readsettings &savesettings &file_export_pagemarkup
       &file_import_markup &file_import_ocr &operationadd &isedited &setedited &charsuitespopup &charsuitecheck &charsuitefind &charsuiteenable
-      &cpcharactersubs &getsafelastpath);
+      &cpcharactersubs &getsafelastpath &add_page_marker_flags &remove_page_marker_flags);
 }
 
 #
@@ -477,7 +477,7 @@ sub file_mark_pages {
         my $pagemark = 'Pg' . $page;
 
         # Standardize page separator line format if necessary
-        unless ( $line =~ /^-----File: (\S+)\.(png|jpg)---/ ) {
+        unless ( $line =~ /^-----File: (\S+)\.(png|jpg)---/ or $line =~ /\[Pg\S+?\]/ ) {
             $textwindow->ntdelete( $linestart, $lineend );
             my $stdline = ( '-' x 5 ) . "File: $page.$ext";
             $stdline .= '-' x ( 75 - length($stdline) );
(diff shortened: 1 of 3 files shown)


windymilla:
Testing notes:
1. File-->Project-->Add Page Marker flags in GG1
2. Save file from GG1
3. Load in another editor
4. If that editor is GG2, then after loading the file, you can File-->Project-->Remove Page Marker flags in GG2 if you don't want them to get in the way.
5. Make the edits you want to in the other editor.
6. If that editor is GG2, then use File-->Project-->Add Page Marker flags in GG2 before a final save.
7. Load the file in GG1 again
8. File-->Project-->Remove Page Marker flags in GG1

These notes should also work if "GG1" & "GG2" are swapped above.

In summary, in both directions of transfer...
**_Immediately before saving, Add Page Markers, and immediately after loading, Remove Page Markers_**
(only if the editor is a version of Guiguts, of course, or it won't have those options).

srjfoo:
I found the `File->Set Page Markers` option in GG1 confusing, but remembered to look it up in the manual. Is this something that just won't be implemented in GG2, or is that something else we need to open an issue about?

srjfoo (approved):
I didn't go out of my way to try to break things on this one. I think that anyone who uses this should be very aware of when it should and shouldn't be used, because I'm not sure that there's much that can be done to protect the PPer from stupid mistakes (which we all make).

Observation: When testing I made a point of checking the contents of both bin files both before and after saving in the gg version I currently had it open in. My main test was just to add some lines at the top of the file and verify that when I saved, the start location for the second page changed to the new line number. It worked as expected for me, but it probably should be documented that until a save is done, the `bin`/`json` files won't be updated (unless, of course, that's pounded into PPers' heads much earlier in the process 😁).

windymilla:
> I found the `File->Set Page Markers` option in GG1 confusing, but remembered to look it up in the manual. Is this something that just won't be implemented in GG2, or is that something else we need to open an issue about?

I don't think we need it in GG2. Our new page marker flags remove any need that there may have been in the past. Historically, there is a Pref in GG1 that lets you *not* set the page markers when you first load a file, and then you would have had to use that menu option to set them later. I wonder if many versions ago, that was the only way of setting page markers, then someone added the feature to auto-set them when the file was loaded (also adding a Pref to keep the old-old-timers happy)

windymilla:
> It worked as expected for me, but it probably should be documented that until a save is done, the `bin`/`json` files won't be updated (unless, of course, that's pounded into PPers' heads much earlier in the process 😁).

However, if you don't do a save, then the changes you've made to the file itself won't be updated, so you've just loaded a file and done nothing to it (or discarded your changes). So, I don't think there would be any need to save the json file in that circumstance.

srjfoo:
> > It worked as expected for me, but it probably should be documented that until a save is done, the `bin`/`json` files won't be updated (unless, of course, that's pounded into PPers' heads much earlier in the process 😁).
> 
> However, if you don't do a save, then the changes you've made to the file itself won't be updated, so you've just loaded a file and done nothing to it (or discarded your changes). So, I don't think there would be any need to save the json file in that circumstance.

Good point.

PR #1299, opened 2024-06-06
Convert negative fractions properly

windymilla:
Previously converted `-1/2` to `½` due to error trying to handle `3-1/2`.

Bug noticed during GG2 development.

### DIFF
--- a/src/lib/Guiguts/CharacterTools.pm
+++ b/src/lib/Guiguts/CharacterTools.pm
@@ -1188,11 +1188,12 @@ sub fractionconvert {
             $outfraction = "$numerator$fracslash$denominator";
         }
 
-        # Add hyphen back if it's actually a minus sign for a negative fraction (not a hyphen in 1-3/4)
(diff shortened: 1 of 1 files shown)


srjfoo:
Granted, it's kind of a silly thing to do, but, if I try to convert a fraction like `-2/23` using "Unicode Only", the numbers and slash are completely thrown away, though the hyphen-minues remains, and I get this error:
> 18:27:10: Use of uninitialized value $outfraction in concatenation (.) or string at /Users/sharonjoiner/Desktop/guiguts-fractions/src/lib/Guiguts/CharacterTools.pm line 1193.

GG2 does nothing.

I'm wondering if that deserves a user-readable message, something like "Attempt to convert fraction to Unicode failed: no equivalent available" in both places? I'm also okay with just leaving it the way GG2 handles it, and ignoring it, with a warning in the manual.

srjfoo (approved):

